PR #1, opened 2016-02-01
Add first test

JohanLorenzo:
@fabricedesre As said offline, the first test doesn't look that fancy due to the lack of mocking. I'll see with Felix if there's a smart way to mock things out. 

In the meantime, we have something travis and coveralls.io can work with.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -33,7 +33,7 @@ impl Context {
     pub fn new(verbose: bool, hostname: Option<String>) -> Context {
         Context { services: HashMap::new(),
                   verbose: verbose,
-                  hostname:  hostname.unwrap_or(DEFAULT_HOSTNAME.to_owned()),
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
Code coverage requires the rust compiler to don't strip binaries. This is handled in rust-lang/rust#31368 . In the meantime, I removed code coverage from the PR.

r? @fabricedesre

src/context.rs
```diff
@@ -75,3 +75,37 @@ impl Context {
...
+            }
+        }
+        fn start(&self)  {}
+        fn stop(&self) {}
+        fn process_request(&self, req: &Request) -> IronResult<Response> {
+            return Ok(Response::new());
+        }
+    }
+
+    let hostname: Option<String> = Some("localhost".to_string());
```
fabricedesre: use to_owned() instead of to_string(), and inline it in Context::new(false, ...)

JohanLorenzo: Done.

PR #2, opened 2016-02-01
Change default listen address from localhost to 0.0.0.0

dhylands:
When you listen to an explicit IP address, then you only accept
connections over one particular interface (lo for localhost).
By using 0.0.0.0 connections will be accepted over any interface
(wifi, eth, lo, etc).


fabricedesre:
I changed to use :: instead.

PR #12, opened 2016-02-04
Turn on linter (clippy)

JohanLorenzo:
Until we get a better test framework, we can have more checks on the code itself.


src/main.rs
```diff
@@ -52,6 +52,11 @@ fn main() {
             return;
         }
     };
+
+    if { true } {
```
JohanLorenzo: This changeset will be removed once we know travis shows:
- a warning in [this job](https://travis-ci.org/fxbox/foxbox/builds/106964559)
- but not in [this one](https://travis-ci.org/fxbox/foxbox/builds/106964277)

JohanLorenzo: And that worked! 

```
src/main.rs:56:8: 56:16 warning: omit braces around single expression condition, #[warn(block_in_if_condition_expr)] on by default

src/main.rs:56     if { true } {

                      ^~~~~~~~

src/main.rs:56:8: 56:16 help: try

if true {

    println!("always true condition");

} ... 

for further information visit https://github.com/Manishearth/rust-clippy/wiki#block_in_if_condition_expr
```

JohanLorenzo:
@fabricedesre r?

PR #17, opened 2016-02-04
Bug #16 - Add selenium, r=johanlorenzo

### LINKED ISSUE #16: Add selenium
Developing here: https://github.com/michielbdejong/foxbox/tree/add-selenium

michielbdejong:
Work in progress.


michielbdejong:
r=? @JohanLorenzo 

I can do a follow-up to add mocha instead of calling selenium-webdriver straight from the main thread of the .js file.

.travis.yml
```diff
@@ -1,6 +1,19 @@
...
+    - sleep 5
+    - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.30.2/install.sh | bash
+    - nvm install 4.2
+    - nvm use 4.2
+    - npm install selenium-webdriver
+
 script:
     - cargo build
     - RUST_BACKTRACE=1 cargo test
+    - ((cargo run ; true) > /dev/null & node test/selenium/ftu_test.js && killall -HUP foxbox) || (killall -HUP foxbox ; exit 1)
```
michielbdejong: My bash is not that strong, I imagine this line can be improved to look nicer. I did test that it fails/passes on Travis if the test fails/passes, and that in both cases the cargo server is killed, and its exit code is ignored.

static/index.html
```diff
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>FoxLink</title>
+    <title>FoxBox</title>
```
michielbdejong, line 4: Since the repo is called foxbox, and neither of the two names have been decided on anyway, I changed this to match the repo name for consistency.

.travis.yml
```diff
@@ -1,6 +1,19 @@
 language: rust
 rust:
     - nightly
+
+before_script:
+    - "sh -e /etc/init.d/xvfb start"
+    - "export DISPLAY=:99.0"
+    - "wget http://selenium-release.storage.googleapis.com/2.50/selenium-server-standalone-2.50.1.jar"
```
JohanLorenzo, line 8: I found a [npm package](https://www.npmjs.com/package/selenium-standalone) which puts some sugar to installation/startup:

``` sh
npm install selenium-standalone@latest -g
selenium-standalone install
selenium-standalone start
```

.travis.yml
JohanLorenzo: I wonder if we could simplify that with:

``` yml
  - cargo run &> /dev/null & # make an instance run while we run integration tests
  - node test/selenium/ftu_test.js
```

Would this keep the Travis machine up if we don't kill the foxbox process?

michielbdejong: No, tried this, but it has to be in the same line.

.travis.yml
michielbdejong, line 8: Tried to get this working and it works on my machine, but not on Travis, not sure why. Will make switching to selenium-standalone a follow-up ticket.

michielbdejong:
I removed the step of installing nvm, because it turns out nvm is already installed. I also simplified the test line because I found out stopping the cargo server after the test finished is not necessary.

Splitting commands to separate script lines in .travis.yml doesn't work, the servers are killed in between.

Will make switching to selenium-standalone a follow-up.

.travis.yml
JohanLorenzo, line 8: Works for me

.travis.yml
```diff
@@ -1,6 +1,18 @@
...
+    - "java -jar selenium-server-standalone-2.50.1.jar > /dev/null &"
+    - sleep 5
+    - nvm install 4.2
+    - nvm use 4.2
+    - npm install selenium-webdriver
+
 script:
     - cargo build
     - RUST_BACKTRACE=1 cargo test
+    - cargo run > /dev/null & node test/selenium/ftu_test.js
```
JohanLorenzo: Nice!

Could you add a `;` just between `&` and `node`? It'll make the fact that 2 commands are on the same line, more obvious.
`cargo run > /dev/null & ; node test/selenium/ftu_test.js`

JohanLorenzo:
r+ @michielbdejong

.travis.yml
```diff
@@ -1,6 +1,18 @@
...
 rust:
     - nightly
+
+before_script:
+    - "sh -e /etc/init.d/xvfb start"
+    - "export DISPLAY=:99.0"
+    - "wget http://selenium-release.storage.googleapis.com/2.50/selenium-server-standalone-2.50.1.jar"
+    - "java -jar selenium-server-standalone-2.50.1.jar > /dev/null &"
+    - sleep 5
+    - nvm install 4.2
```
JohanLorenzo, line 11: Yay! nvm is already present :)

.travis.yml
michielbdejong: That syntax errors with `unexpected token ;`, but changed it to:
`(cargo run > /dev/null &) ; node test/selenium/ftu_test.js`

JohanLorenzo: My bad, it was working under my zsh. I should have checked on bash.

Issue #6, opened 2016-02-03
Consider using docopt for command line args


jedireza:
The docopt crate seems like a nice way to manage command line args:
https://github.com/docopt/docopt.rs

Related to: #3

hfiguiere:
We can close this now as the PR got merged!

PR #3, opened 2016-02-01
Add command line options to parse http and ws port numbers

dhylands:
Some services, like ntop already use port 3000, so this allows alternate ports to be used.

PR #18, opened 2016-02-05
Issue #6 - Parse command line with docopt.

hfiguiere:


hfiguiere:
@jedireza since you filed the issue, do you mind reviewing?

hfiguiere:
It should be noted that PR #3 will instantly rot with this one. And vice versa.

hfiguiere:
@fabricedesre ? Thanks !

src/main.rs
```diff
@@ -23,6 +22,8 @@ extern crate router;
 extern crate serde;
 extern crate staticfile;
 extern crate uuid;
+extern crate docopt;
+extern crate rustc_serialize;
```
fabricedesre: Why do we need that? It seems unused.

fabricedesre: also, add them in alphabetical order

Cargo.toml
```diff
@@ -17,3 +16,5 @@ serde_json = "0.6.0"
 serde_macros = "0.6.11"
 uuid = "0.1.18"
 clippy = "0.0.37"
+docopt = "0.6.78"
+rustc-serialize = "0.3"
```
fabricedesre: add the new dependencies in alphabetical order (bonus points if you also move clippy)

src/main.rs
```diff
@@ -33,39 +34,38 @@ mod controler;
...
+const USAGE: &'static str = "
+Usage: foxbox [-v] [-h] [-n <hostname>]
+
+Options:
+    -v, --verbose            Toggle verbose output.
+    -n, --name               Set local hostname.
+    -h, --help               Print this help menu.
+";
+
+#[derive(RustcDecodable)]
```
jedireza, line 48: Looks like this is what `rustc_serialize` was used for.

src/main.rs
hfiguiere: We need it for the Args. Will keep the crate sorted too.

src/main.rs
hfiguiere, line 48: Yup. That's the reason.

hfiguiere:
I have reordered the dependencies as requested. And squished commits.

hfiguiere:
@fabricedesre should be ready. Thanks !

PR #23, opened 2016-02-05
Test the command line options from bug #6

hfiguiere:
I forgot to have tests. This will perform the sanity testing that our command line arguments are consistent. `Docopt` does that when called and fail, so better ensure it is tested :-D


hfiguiere:
@fabricedesre I forgot tests.

PR #26, opened 2016-02-06
Better coverage for context.rs

fabricedesre:
I was trying an idea for mocks (that didn't work) and while doing that I expanded the current tests.

r? @JohanLorenzo

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -82,7 +82,7 @@ impl Context {
 
 #[test]
 #[allow(unused_variables)]
-fn test_should_add_a_service() {
+fn test_context_service() {
     use service::{ Service, ServiceProperties };
     use iron::{Request, Response, IronResult};
 
@@ -91,11 +91,11 @@ fn test_should_add_a_service() {
     impl Service for ServiceStub {
         fn get_properties(&self) -> ServiceProperties {
             ServiceProperties {
-                id: '1'.to_string(),
+                id: "1".to_owned(),
                 name: "dummy service".to_owned(),
                 description: "really nothing to see".to_owned(),
-                http_url: '2'.to_string(),
-                ws_url: '3'.to_string()
+                http_url: "2".to_owned(),
+                ws_url: "3".to_owned()
             }
         }
         fn start(&self)  {}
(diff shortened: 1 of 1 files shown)


src/context.rs
```diff
@@ -106,9 +106,37 @@ fn test_should_add_a_service() {
...
+
+    let id = "1".to_owned();
+    assert_eq!(foo.get_http_root_for_service(id.clone()),
+               "http://localhost:3000/services/1/");
+    assert_eq!(foo.get_ws_root_for_service(id.clone()),
+               "ws://localhost:4000/services/1/");
+
+    match foo.services_as_json() {
+        Ok(txt) => assert_eq!(txt, "{\"1\":{\"id\":\"1\",\"name\":\"dummy service\",\"description\":\"really nothing to see\",\"http_url\":\"2\",\"ws_url\":\"3\"}}"),
+        Err(err) => assert!(false, err)
```
JohanLorenzo, line 136: Is it possible to use `panic!(err)` in this context? If so, maybe the code will be less ambiguous than making sure the error is not false.

JohanLorenzo:
I would split the tests into many (one for each type of verification), so the test name would reflect what it actually does.

I wanted to verify if that was easily doable in rust. So I played with `mod` and their scope, and I turned out to take a long time to write. Then, it sounds more efficient if we land this patch first and I upload what I have in a follow up. What do you think @fabricedesre ? 

Edit: I left a proposal of split at #30 .

src/context.rs
```diff
@@ -106,9 +106,37 @@ fn test_should_add_a_service() {
...
+    assert_eq!(foo.services_count(), 1);
+
+    // Verify that the service with id "1" is there.
+    {
+        let service1 = foo.get_service("1");
+        match service1 {
+            Some(svc) => {
+                assert_eq!(svc.get_properties().id, "1");
+            }
+            None => assert!(false, "No service with id 1")
```
JohanLorenzo, line 124: Same comment about `panic!()`

Edit: Better to leave assert actually.

src/context.rs
JohanLorenzo, line 136: Actually, using `panic!()` doens't sound that great. If we use `#[should_panic]`, we can hide another error, which wasn't the one we expected. In this case, the test becomes a false positive, which can be harmful.

JohanLorenzo:
r+!

PR #27, opened 2016-02-06
Add command line options to set listening ports

hfiguiere:
This a rework of PR #3. (dhylands gave me the go ahead since his PR had bitrot)

Also fixed a serious error in the parsing and added much better tests for them.

r? @fabricedesre


fabricedesre:
thanks!

PR #30, opened 2016-02-08
Split tests into modules and smaller functions

JohanLorenzo:


src/context.rs
```diff
@@ -105,10 +106,83 @@ fn test_should_add_a_service() {
...
+        use super::super::*;
 
-    assert_eq!(foo.services.is_empty(), true);
-    foo.add_service(Box::new(service));
-    assert_eq!(foo.services.is_empty(), false);
+        #[test]
+        fn test_should_increase_number_of_services() {
+            let service = super::ServiceStub;
+            let context = Context::shared(false, Some("localhost".to_owned()), None, None);
+            let mut locked_context = context.lock().unwrap();
```
JohanLorenzo: In a perfect world, these variables should be initialized in a setup-like function. I tried to play with that. I have not managed to return the correct type, so the compiler wasn't very happy.

Also, if we declared the variables in the module itself, they should be `static mut`. Using a `static mut` variable needs to be done in an `unsafe` block. I don't think this is a good practice, even for testing.

fabricedesre: What was the setup function that you could not get to work? Can you try to turn these tree lines into a macro at least? I'm really not a big fan of all this duplication.

JohanLorenzo: I'm not a fan either. The macro approach sounds good. I'll give it a try.

I was trying something like:

``` rust
// in tests module
fn setup() -> (Context, ServiceStub) {
    let service = ServiceStub;
    let context = Context::shared(false, Some("localhost".to_owned()), None, None);
    let mut locked_context = context.lock().unwrap();
    (locked_context, service)
}

// in add_service
[test]
fn a_test() {
    let (context, service) = &mut super::setup();
    // and the test continues
}
```

The error I got was: 

```
src/context.rs:113:10: 113:24 error: mismatched types:
 expected `context::Context`,
    found `std::sync::mutex::MutexGuard<'_, context::Context>`
(expected struct `context::Context`,
    found struct `std::sync::mutex::MutexGuard`) [E0308]
src/context.rs:113         (locked_context, service)
```

JohanLorenzo: FYI, the macro approach gives another type of error:

```
src/context.rs:45:39: 45:40 error: unresolved name `shared_context` [E0425]
src/context.rs:45         setup!(service, shared_context, locked_context);
                                                        ^~
src/context.rs:45:9: 45:57 note: in this expansion of setup! (defined in src/context.rs)
src/context.rs:45:39: 45:40 help: run `rustc --explain E0425` to see a detailed explanation
src/context.rs:46:17: 46:40 error: the type of this value must be known in this context
src/context.rs:46         assert!(locked_context.services.is_empty());
                                  ^~~~~~~~~~~~~~~~~~~~~~~
src/context.rs:46:9: 46:53 note: in this expansion of assert! (defined in <std macros>)
```

 I isolated the [code that triggers the failure](https://github.com/JohanLorenzo/foxbox/blob/debug-setup-macro/src/context.rs#L45)

JohanLorenzo: It turns out one of the links given by @brson suits our current needs. I used [stainless](https://github.com/reem/stainless) which enables `before_each` macros and nesting.

Thanks @brson!

Cargo.toml
```diff
@@ -21,3 +21,4 @@ serde_macros = "0.6.11"
 staticfile = "0.1.0"
 uuid = "0.1.18"
 websocket = "0.15.0"
+stainless = { git = "https://github.com/reem/stainless.git", rev = "82392f2e" }
```
fabricedesre, line 27: Why can't we use https://crates.io/crates/stainless ?

src/main.rs
```diff
@@ -15,6 +15,10 @@
 #![plugin(clippy)]
 #![deny(clippy)]
 
+#![feature(const_fn)] // Dependency of stainless
+#![plugin(stainless)] // Test runner
+
+
```
fabricedesre: nit: extra blank line

Cargo.toml
JohanLorenzo, line 27: See https://github.com/reem/stainless/issues/43#issuecomment-174245604. I'll add a comment within Cargo.toml, to make that clear.

JohanLorenzo, line 27: Done.

src/main.rs
JohanLorenzo: Done.

PR #36, opened 2016-02-09
more explicit node dependency

jedireza:
It wasn't obvious that I need to have a node to run the selenium tests. This PR adds a `package.json` file so we can have a proper place for dev dependencies as well as defining the version of node that is supported (via the `engines` key).

Now locally we can simply setup via:

``` bash
$ npm install
```

And then run the tests by having the Rust server running on port 3000 and:

``` bash
$ npm test
```

:pager: @michielbdejong @JohanLorenzo 

If this is ok with you I'd like to add some instructions about this to the readme before merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ language: rust
 rust:
     - nightly
 
-addons:
-  hosts:
-    - foxbox.local
-
 before_script:
     - "sh -e /etc/init.d/xvfb start"
     - "export DISPLAY=:99.0"
@@ -14,10 +10,10 @@ before_script:
     - sleep 5
     - nvm install 4.2
     - nvm use 4.2
-    - npm install selenium-webdriver mocha jshint
+    - npm install
 
 script:
     - cargo build # Linter is also executed here
     - RUST_BACKTRACE=1 cargo test
     - jshint static/*.js test/selenium/*.js
-    - (cargo run &> /dev/null &) ; ./node_modules/.bin/mocha test/selenium/ftu_test.js
+    - (cargo run &> /dev/null &) ; npm test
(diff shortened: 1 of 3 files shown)


michielbdejong:
About adding package.json - if that's ok with @fabricedesre then I also fully support that change - I kept it out because nodejs is only used for running the GUI tests right now.

About setting the hostname back to localhost - right now, there is no build process that sets the foxbox.local hostname, but I would like this to exist in the future, that's why I kinda like it being there in the GUI tests and in [the user manual](https://github.com/fxbox/foxbox/blob/master/docs/user-manual.md).

fabricedesre:
I'm fine with adding package.json too.

About the foxbox.local name setting, it should be set by the daemon itself to reflect the --name command line argument. We'll need the avahi bindings that @azasypkin is working on and implement something similar to http://www.avahi.org/wiki/Examples/PythonPublishAlias (without dbus).

jedireza:
I updated the README in #37 since I changed more than just adding details for Node. We should merge this first and that second.

jedireza:
> About setting the hostname back to localhost - right now, there is no build process that sets the foxbox.local hostname, but I would like this to exist in the future, that's why I kinda like it being there in the GUI tests and in the user manual.

I'm not a fan of the added complexity of needing to edit my `/ect/hosts` just to run tests locally, especially when I have the server (`http://localhost:3000/`) running already.

JohanLorenzo:
> This PR adds a package.json 

:+1: 

> I'm not a fan of the added complexity of needing to edit my /ect/hosts

I agree. What do you all think of:
- using `localhost` as the default hostname for selenium tests
- and create a separate test called `FoxBox should be accessible by the domain name called "foxbox.local" (given by avahi daemon)` ? 

Then, if such a test failed on one's computer, the failure would document the intent.

jedireza:
Rebased to resolve conflicts. r? @fabricedesre

PR #37, opened 2016-02-10
rewrote and organized instructions

jedireza:
Depends on #36

PR #19, opened 2016-02-05
[WIP] Switch to selenium-standalone

michielbdejong:
Doesn't work yet on Travis for some reason (does work when I try it locally)

PR #39, opened 2016-02-10
[WiP] Installing selenium is not necessary

michielbdejong:
Trying out https://github.com/fxbox/foxbox/pull/37#discussion_r52428664

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,6 @@ addons:
     - foxbox.local
 
 before_script:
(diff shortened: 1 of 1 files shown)


michielbdejong:
The tests fail if you don't install selenium first. That's what I thought, but was intrigued by [@jedireza's remark](https://github.com/fxbox/foxbox/pull/37#discussion_r52428664). So it looks like some sort of installation of selenium is necessary, but maybe we can move it into the `npm install` step, see also #19.

PR #40, opened 2016-02-10
Mute foxbox's stderr when running selenium tests

JohanLorenzo:


JohanLorenzo:
r? @michielbdejong 

From mailing list:

> On another note, `cargo run` doesn't output stderr to /dev/null. This gives us those errors message, unrelated to what is being tested[1], and can lead someone to think the tests are not passing. One can be mislead by another discrepancy: errors are shown but the process exits with 0.
> Then, I propose to mute the HTTP server when tested with selenium: https://github.com/fxbox/foxbox/pull/40
> 
> Do you guys agree?

michielbdejong:
r+

PR #41, opened 2016-02-10
Use foxbox/users

ferjm:
r? @fabricedesre

### DIFF
--- a/src/http_server.rs
+++ b/src/http_server.rs
@@ -2,10 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-extern crate foxbox_users;
-
 use context::SharedContext;
-use self::foxbox_users::users_router::UsersRouter;
+use foxbox_users::users_router::UsersRouter;
 use iron::{Iron, Request, Response, IronResult};
 use iron::headers::ContentType;
 use iron::status::Status;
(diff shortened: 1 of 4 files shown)


jedireza:
I don't think this needs to be an external crate. It adds a level of indirection that I don't think is necessary.

Could this approach cause version drift across `foxbox` and `users` code? Does this add complexity to maintenance?

ferjm:
Cargo should handle the versioning. If something changes in fxbox/users, your fxbox/foxbox clone won't get the changes until you make "cargo update" locally. If that's too much trouble I can merge fxbox/users here.

src/http_server.rs
```diff
@@ -2,7 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+extern crate foxbox_users;
```
fabricedesre: Any reason to not move that in main.rs with the other external crates?

src/http_server.rs
```diff
@@ -56,9 +59,12 @@ impl HttpServer {
             }
         });
 
+        let users_router = UsersRouter::new();
+
         let mut mount = Mount::new();
         mount.mount("/", Static::new(Path::new("static")))
-             .mount("/services", router);
+             .mount("/services", router)
+             .mount("/users_admin", users_router);
```
fabricedesre: Can you inline UsersRouter::new() here?

Also, maybe /users as a url?

fabricedesre:
I'd also like to keep that as an external crate. This is one piece that could be useful to others, along some other like the mDNS & openzwave libraries.

src/http_server.rs
ferjm: There are several [endpoints ](https://github.com/fxbox/users/blob/master/doc/API.md#api-endpoints) with a path already including `users` so I figured that `users_admin` would be a better fit. So we would have urls like: https://foxbox.local/users_admin/setup https://foxbox.local/users_admin/users etc.

fabricedesre: ok!

PR #48, opened 2016-02-15
Add unit tests to service router

JohanLorenzo:


src/service_router.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        let mut response = Response::with(serialized);
+        response.status = Some(Status::Ok);
+        response.headers.set(ContentType::json());
+
+        Ok(response)
+    });
+
+    let context2 = context.clone();
+    router.get(":service", move |req: &mut Request| -> IronResult<Response> {
```
fabricedesre: Why did you remove the /:command part from the url? It's used by the current DummyService at https://github.com/fxbox/foxbox/blob/master/src/dummy_adapter.rs#L74 (yeah, if we had tests...)

src/service_router.rs
```diff
@@ -0,0 +1,90 @@
...
+            Some(service) => {
+                service.process_request(req)
+            }
+        }
+    });
+    router
+}
+
+
+extern crate iron_test;
```
fabricedesre, line 55: Can we scope that in a test-only scope?

src/service_router.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    it "should return an error if no service was found" {
+        let response = request::get("http://localhost:3000/unknown-id",
+                        Headers::new(),
+                        &service_router).unwrap();
+
+        let result = response::extract_body_to_string(response);
+        assert_eq!(result, "No Such Service: unknown-id");
+    }
```
fabricedesre, line 92: It's fine to do it in a follow up, but we should also test the case where the service exists. But we need websocket based notifications to prevent races.

src/service_router.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use context::{ ContextTrait, Shared };
+use iron::{Request, Response, IronResult};
+use iron::headers::ContentType;
+use iron::status::Status;
+use router::Router;
+use core::marker::Reflect;
+
+
+pub fn generate<Ctx: Send + Reflect + ContextTrait + 'static> (context: Shared<Ctx>) -> Router {
```
fabricedesre: nit: s/generate/create

src/service_router.rs
JohanLorenzo, line 55: The compiler is not super happy about it:

```
src/service_router.rs:55:35: 55:42 error: unresolved import `self::iron_test::request`. Could not find `iron_test` in `service_router::service_router` [E0432]
src/service_router.rs:55         pub use self::iron_test::{request, response};
```

I'll make the comment below clearer.

src/service_router.rs
JohanLorenzo, line 92: I'm not too sure to understand. [This test](https://github.com/fxbox/foxbox/pull/48/files#diff-2b028424809a6315630f8281a5f7d1e3R73) checks the case where the service exists. Maybe I should rename it.

src/service_router.rs
JohanLorenzo, line 55: Done.

src/service_router.rs
JohanLorenzo: Done.

src/service_router.rs
JohanLorenzo: Thanks for pointing this out. I misunderstood the function. The URL I wrote in the tests (like [this one](https://github.com/fxbox/foxbox/pull/48/files#diff-2b028424809a6315630f8281a5f7d1e3R83)) were failing. As I didn't see any `command` being used in this context, I thought it was a nit. 

I fixed the tests by adding a command in the URL.

However, that makes me wonder if we shouldn't make this function owner of `command` and pass it explicitly to the service. For instance:

``` rust
let router_plugin = req.extensions.get::<Router>().unwrap(); // TODO: Find a var name less close to `router`
let id = router_plugin.find("service").unwrap_or("");
let downstream_request = {
    command: router_plugin.find("command").unwrap_or("");
};

match ctx.get_service(id) {
    None => {// ...}
    Some(service) => {
        service.process_request(downstream_request)
    }
}
```

What do you think @fabricedesre ?

src/service_router.rs
fabricedesre, line 92: But it's not testing sending a command. Which is fine because we'll likely have a lot of churn in this area.

PR #50, opened 2016-02-16
Put test dependency out of regular build

JohanLorenzo:
- stainless in now under [dev-dependencies]
- the plugin is included only when cfg=test
- and so do the tests


ferjm:
LGTM!

PR #53, opened 2016-02-16
fix a spelling error in README.md

davidflanagan:


davidflanagan:
My first PR is completely pedantic. But the spelling mistake is in a section title so it is kind of prominent and seems worth fixing.

jedireza:
Thanks

PR #54, opened 2016-02-17
Unit tests to http server

JohanLorenzo:
WIP

### DIFF
--- a/src/service_router.rs
+++ b/src/service_router.rs
@@ -10,7 +10,7 @@ use router::Router;
 use core::marker::Reflect;
 
 
-pub fn create<T: Send + Reflect + ContextTrait + 'static> (context: Shared<T>) -> Router {
+pub fn create<Ctx: Send + Reflect + ContextTrait + 'static> (context: Shared<Ctx>) -> Router {
 
     let mut router = Router::new();
     let context1 = context.clone();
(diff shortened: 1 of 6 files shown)


src/service_router.rs
```diff
@@ -10,7 +10,7 @@ use router::Router;
 use core::marker::Reflect;
 
 
-pub fn create<T: Send + Reflect + ContextTrait + 'static> (context: Shared<T>) -> Router {
+pub fn create<Ctx: Send + Reflect + ContextTrait + 'static> (context: Shared<Ctx>) -> Router {
```
JohanLorenzo, line 13: Topic discussed in #56.

JohanLorenzo:
I'm under the impression this PR causes more problems than it actually solves.

Nonetheless, thanks to it, we noticed:
- 1 bug in stainless
- another in hyper
- 1 missing export in Iron.

This is mainly why we have so many TODOs. I think this PR has still some value, for the sake of documentation. This also paves the road for testing the main routes independently.

What do you think @fabricedesre ? r?

fabricedesre:
Too much noise here for the benefits. Let's postpone until there are less TODO and FIXME.

JohanLorenzo:
Fine by me. I close this PR, until there are less TODOs to deal with.

PR #56, opened 2016-02-19
following convention for generic parameter type

jedireza:
Minor syntax change.

>  Convention says that the first generic parameter should be T, for ‘type’

Source: https://doc.rust-lang.org/book/generics.html

### DIFF
--- a/src/http_server.rs
+++ b/src/http_server.rs
@@ -12,12 +12,12 @@ use std::path::Path;
 use std::thread;
 use core::marker::Reflect;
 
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
I'm not sure if we should follow this convention, here. Unlike what's described in the document, this part of the code can't deal with any kind of types, but only the ones which are implementing `ContextTrait`.

Hence, I'd keep something close to `Context`. What do you guys think @jedireza @fabricedesre ?

jedireza:
That's what our bounds are there for `Send + Reflect + ContextTrait + 'static`

PR #57, opened 2016-02-19
Add a generic process manager and some basic setup for the ngrok process

samgiles:
This requires `ngrok` in the `$PATH`. 

I just wanted to get some eyes on this - next I need to add `timebomb` to the tests and look at standing up some integration tests. 

If you want to try it, this should work with the free ngrok service :).

Couple of things I'm unsure about at this moment:  
1. Managing the process in the 'context', it seems to make sense at the moment. 
2. Also proxying the web socket connection.
3. Turning on/off the tunnel via the HTTP API.

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -44,6 +44,11 @@ impl Controller {
         // Start the dummy adapter.
         let dummy_adapter = DummyAdapter::new(self.sender.clone(), self.context.clone());
         dummy_adapter.start();
+
+        {
+            let mut context = self.context.lock().unwrap();
+            context.start_tunnel().unwrap();
+        }
     }
 }
 
--- a/src/managed_process.rs
+++ b/src/managed_process.rs
@@ -1,0 +1,252 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Assumes Unix
+use libc::{self,  c_int};
+
+use std::thread;
+use std::thread::JoinHandle;
+use std::sync::{ Arc, Mutex };
+use std::process::Child;
+use std::io::{ Error, ErrorKind, Result };
+use std::time::{ SystemTime, Duration, UNIX_EPOCH };
+
+/// Unix exit statuses
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+pub struct ExitStatus(c_int);
+
+const RESTART_TIME_THRESHOLD: f64 = 5.0; // seconds
+
+fn seconds_since_epoch() -> f64 {
+    let now = SystemTime::now();
+    now.duration_from_earlier(UNIX_EPOCH).unwrap().as_secs() as f64
+}
+
+pub struct ManagedProcess {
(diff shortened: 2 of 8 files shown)


samgiles:
Heh, and my tests are broken

src/controller.rs
```diff
@@ -40,6 +40,11 @@ impl Controller {
         // Start the dummy adapter.
         let dummy_adapter = DummyAdapter::new(self.sender.clone(), self.context.clone());
         dummy_adapter.start();
+
+        {
+            let mut context = self.context.lock().unwrap();
+            context.start_tunnel().unwrap();
+        }
```
jedireza, line 51: This is at the end of the function block and owned variables are about to go out of scope, what does the nested scope grant us here?

samgiles, line 51: It's to keep the lock contained.  It's going out of scope, but I liked to keep it obvious in case something is added after and the scope of the lock is neglected.

fabricedesre:
That looks nice overall. I haven't tried to run it yet, but I have a few comments/questions:
- please move the ManagedProcess in its own file.
- what happens if the process dies because of an external factor or crash? We need to monitor that and restart the failed process.
- obviously a followup, but that would be great to extend that to also setup an ipc channel between the two processes, eg. using a unix domain socket.

src/controller.rs
jedireza, line 51: I see. I had to do something just like this over here:
https://github.com/fxbox/foxbox/pull/59/files#diff-e1ea9eba222b6cc49837b3dd765989dfR21

In this case, since `context` is about to go out of scope, maybe a comment would suffice instead?

jedireza:
After this merges will it require everyone have `ngrok` or does it move on gracefully?

samgiles:
@jedireza No, if you specify a tunnel using the command line, it will keep retrying and say ngrok couldn't be found, if you don't specify a tunnel using the command line, it will not try to do anything.

@fabricedesre If the process dies because of a crash it will restart  (https://github.com/fxbox/foxbox/pull/57/files#diff-552eb6bc31362290e77bf8f3ea119452R138) - need to add some sort of intelligent back off.   Extending this should be super easy to add and use IPC channels with unix domain sockets - we don't get any value doing that with ngrok right now that I can see.

src/controller.rs
samgiles, line 51: I'd prefer to keep it explicitly there in all cases, comments can go out of date.

samgiles:
@fabricedesre I've added exponential back off.  I'd like to make that more generic and tested - might make a separate crate for it as a mixin type thing - but here's a working version for now.  Squashed and rebased too.

src/managed_process.rs
```diff
@@ -0,0 +1,248 @@
...
+                       backoff += 1;
+                       let backoff_seconds = (backoff * backoff) / 2;
+                       info!("Backing off creating a new process for {} seconds", backoff_seconds);
+                       thread::sleep(Duration::new(backoff_seconds, 0));
+                    } else {
+                        backoff = 0;
+                    }
+
+                    info!("Starting process. Restarted {} times", starts);
+                    start_time = time::precise_time_s();
```
fabricedesre: I would rather not depend on a crate just for that. I had a similar need and did it with https://github.com/fxbox/registration_server/blob/master/src/db.rs#L65

PR #58, opened 2016-02-20
Client side of the registration service.

fabricedesre:
This is the client side of https://github.com/fxbox/registration_server
By default it pings the registration server every minute and will expect the server to be on localhost.

There's a public endpoint running for test purposes: use the following command line to use it:
`RUST_LOG=info cargo run -- -r http://cloud.desre.org:4242/register`

r? @arcturus

### DIFF
--- a/src/registration.rs
+++ b/src/registration.rs
@@ -1,0 +1,111 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// This manages registration of the foxbox with the discovery endpoint.
+/// For now it simply register itselfs every N minutes with the endpoint,
+/// after trying more aggressively at first run.
+
+extern crate get_if_addrs;
+extern crate hyper;
+
+use self::hyper::Client;
+use self::hyper::header::Connection;
+use self::hyper::status::StatusCode;
+use self::get_if_addrs::IfAddr;
+use std::io::Read;
+use std::time::Duration;
+use std::thread;
+
+const DEFAULT_ENDPOINT: &'static str = "http://localhost:4242/register";
+const REGISTRATION_INTERVAL: u32 = 1; // in minutes.
+
+pub struct Registrar;
+
+impl Registrar {
+    pub fn new() -> Registrar {
+        Registrar
+    }
+
+    pub fn start(&self, endpoint: Option<String>) {
+        let url = match endpoint {
+            Some(u) => u,
+            _ => DEFAULT_ENDPOINT.to_owned()
+        };
+
+        info!("Starting registration with {}", url);
+        let ip_addr = self.get_ip_addr();
+        if ip_addr == None {
+            // TODO: retry later, in case we're racing with the network
+            // configuration.
+            return;
+        }
+
+        info!("Got ip address: {}", ip_addr.clone().unwrap());
+        let full_address = format!("{}?ip={}", url, ip_addr.unwrap());
+
+        // Spawn a thread to register every REGISTRATION_INTERVAL minutes.
+        thread::Builder::new().name("Registrar".to_owned())
+                              .spawn(move || {
+            loop {
+                let client = Client::new();
+                let res = client.get(&full_address)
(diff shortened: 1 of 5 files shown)


src/registration.rs
```diff
@@ -0,0 +1,110 @@
...
+use self::hyper::Client;
+use self::hyper::header::Connection;
+use self::hyper::status::StatusCode;
+use self::get_if_addrs::IfAddr;
+use std::io::Read;
+use std::time::Duration;
+use std::thread;
+
+const DEFAULT_ENDPOINT: &'static str = "http://localhost:4242/register";
+const REGISTRATION_INTERVAL: u64 = 1; // in minutes.
```
arcturus: nit: u32 will be enough :)

src/registration.rs
```diff
@@ -0,0 +1,110 @@
...
+pub struct Registrar;
+
+impl Registrar {
+    pub fn new() -> Registrar {
+        Registrar
+    }
+
+    pub fn start(&self, endpoint: Option<String>) {
+        let url = match endpoint {
+            Some(u) => u,
```
arcturus, line 32: Should we validate the url passed?
Not needed in this PR we could do a follow up.

src/registration.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        info!("Got ip address: {}", ip_addr.clone().unwrap());
+        let full_address = format!("{}?ip={}", url, ip_addr.unwrap());
+
+        // Spawn a thread to register every REGISTRATION_INTERVAL minutes.
+        thread::Builder::new().name("Registrar".to_owned())
+                              .spawn(move || {
+            loop {
+                let client = Client::new();
+                let res = client.get(&full_address)
```
arcturus, line 52: nit: not to be done in this pr but in the service, perhaps this should be a post.

src/registration.rs
fabricedesre: yep, just did that to avoid a cast in Duration::from_secs(). I can change it if you prefer.

src/registration.rs
```diff
@@ -0,0 +1,110 @@
...
+            if ifaces.is_empty() {
+                error!("No IP interfaces found!");
+                return None;
+            }
+
+            let mut ip_addr: Option<String> = None;
+
+            for iface in ifaces {
+                if ! (iface.name.starts_with("eth") ||
+                      iface.name.starts_with("wlan")) {
```
arcturus: In macosx interface names start with enX

src/registration.rs
```diff
@@ -0,0 +1,110 @@
...
+            ip_addr
+        } else {
+            error!("No IP interfaces found!");
+            None
+        }
+    }
+}
+
+#[test]
+fn check_ip_addr() {
```
arcturus, line 107: Maybe we can add this as a test module to avoid compilation when non testing: #[cfg(test)] ...

src/service_router.rs
```diff
@@ -4,7 +4,7 @@
 
 use context::{ ContextTrait, Shared };
 use iron::{Request, Response, IronResult};
-use iron::headers::ContentType;
+use iron::headers::{ ContentType, AccessControlAllowOrigin };
```
arcturus, line 7: Are the changes in `service_router.rs` intended for the origin of the PR? Or maybe I'm missing something.

src/registration.rs
arcturus: Confirm on my macosx, it will fail to fetch local ip unless you:

``` rust
.... ||
iface.name.starts_with("en"))
```

arcturus:
r+

mainly nits, the only problem we need to solve is the macosx use case for network interfaces names.

I patched locally and managed to get registered, this is the output in the registration server:

``` json
[{ "public_ip": "90.206.147.50",  "local_ip": "192.168.0.7", "timestamp": 1456255841 }]
```

arcturus:
I'm just thinking, what about those devices that have 2 ip addresses?
Should we add an extra parameter in the command line to specify, either the interface or the ip address foxbox should use?

Maybe we can do this in a follow up.

src/service_router.rs
fabricedesre, line 7: I added them to get my demo client working. I guess we'll change/revert once we have proper user authentication working.

PR #59, opened 2016-02-20
event loop notifications to websocket clients

jedireza:
@mcav and I worked on this.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -7,20 +7,21 @@ extern crate collections;
 
 use self::collections::vec::IntoIter;
 use service::Service;
-use std::collections::HashMap;
+use std::collections::{ HashMap, hash_map };
 use std::io::Error;
 use std::net::SocketAddr;
 use std::net::ToSocketAddrs;
 use std::sync::{ Arc, Mutex };
+use ws;
 
 // The `global` context available to all.
 pub struct Context {
     pub verbose: bool,
-
-    hostname: String,
-    http_port: u16,
-    ws_port: u16,
-    services: HashMap<String, Box<Service>>
+    pub hostname: String,
+    pub http_port: u16,
+    pub ws_port: u16,
+    services: HashMap<String, Box<Service>>,
+    websockets: HashMap<ws::util::Token, ws::Sender>,
 }
 
 const DEFAULT_HTTP_PORT: u16 = 3000;
@@ -35,9 +36,12 @@ pub trait ContextTrait {
               ws_port: Option<u16>) -> Shared<Self>;
     fn add_service(&mut self, service: Box<Service>);
     fn remove_service(&mut self, id: String);
+    fn add_websocket(&mut self, socket_out: ws::Sender);
+    fn remove_websocket(&mut self, socket_out: ws::Sender);
     fn services_count(&self) -> usize;
     fn get_service(&self, id: &str) -> Option<&Box<Service>>;
     fn services_as_json(&self) -> Result<String, serde_json::error::Error>;
+    fn websockets_iter(&self) -> hash_map::Values<ws::util::Token, ws::Sender>;
     fn get_http_root_for_service(&self, service_id: String) -> String;
     fn get_ws_root_for_service(&self, service_id: String) -> String;
     fn http_as_addrs(&self) -> Result<IntoIter<SocketAddr>, Error>;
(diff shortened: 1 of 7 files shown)


src/context.rs
```diff
@@ -35,6 +36,8 @@ pub trait ContextTrait {
               ws_port: Option<u16>) -> Shared<Self>;
     fn add_service(&mut self, service: Box<Service>);
     fn remove_service(&mut self, id: String);
+    fn add_socket(&mut self, socket_out: ws::Sender);
+    fn remove_socket(&mut self, id: ws::util::Token);
```
fabricedesre: I would rather have remove_socket() also take a ws::Sender. No need to show the implementation detail used by add_socket()

src/controller.rs
```diff
@@ -53,6 +57,11 @@ impl mio::Handler for Controller {
         println!("Receiving a notification! {}", data.description());
 
         let mut context = self.context.lock().unwrap();
+
+        for socket in context.sockets.values() {
+            let _ = socket.send(data.description());
```
fabricedesre: can we at least log when sending fail? And think about the right thing to do (eg. closing the socket?)

src/ws_server.rs
```diff
@@ -0,0 +1,67 @@
+use std::thread;
```
fabricedesre, line 5: Please add the MPL2 header

src/context.rs
jedireza: updated

src/controller.rs
jedireza: logging now

src/ws_server.rs
jedireza, line 5: added

src/ws_server.rs
```diff
@@ -0,0 +1,67 @@
...
+use context::{ ContextTrait, SharedContext };
+use ws::{Handler, Sender, Result, Message, Handshake, CloseCode, Error};
```
fabricedesre: Nit: { Handler... , Error };

src/ws_server.rs
```diff
@@ -0,0 +1,67 @@
...
+            let port;
+            {
+                let ctx = context.lock().unwrap();
+                hostname = ctx.hostname.clone();
+                port = ctx.ws_port;
+            }
+
+            let host_port = (&hostname as &str, port);
+
+            listen(host_port, |out| {
```
fabricedesre: nit: inline host_port here?

src/ws_server.rs
```diff
@@ -0,0 +1,67 @@
...
+        println!("Hello new ws connection");
+
+        self.context.lock().unwrap().add_socket(self.out.clone());
+
+        Ok(())
+    }
+
+    fn on_message(&mut self, msg: Message) -> Result<()> {
+        // echo the message back to the client
+        self.out.send(msg)
```
fabricedesre: Not sure if this is a good idea to send back the message to the client. Let's just log for now.

src/ws_server.rs
jedireza: fixed

src/ws_server.rs
jedireza: fixed

src/ws_server.rs
jedireza: logging now instead of echoing

src/context.rs
```diff
@@ -12,15 +12,16 @@ use std::io::Error;
...
-
-    hostname: String,
-    http_port: u16,
-    ws_port: u16,
-    services: HashMap<String, Box<Service>>
+    pub hostname: String,
+    pub http_port: u16,
+    pub ws_port: u16,
+    pub services: HashMap<String, Box<Service>>,
+    pub sockets: HashMap<ws::util::Token, ws::Sender>,
```
fabricedesre: it still looks like `services` and `sockets` can be private. I'd rather keep that.

src/context.rs
```diff
@@ -72,6 +76,14 @@ impl ContextTrait for Context {
         self.services.remove(&id);
     }
 
+    fn add_socket(&mut self, socket: ws::Sender) {
```
fabricedesre: Change to add{remove}_websocket(...). socket is a bit too generic.

src/context.rs
jedireza: I only made things public as needed.

src/context.rs
```diff
@@ -84,6 +97,10 @@ impl ContextTrait for Context {
         serde_json::to_string(&self.services)
     }
 
+    fn websockets_iterator(&self) -> hash_map::Values<ws::util::Token, ws::Sender> {
```
fabricedesre: sorry to be nitpicking again: the trend in Rust land seems to be to just use the short `iter` form (eg. https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.iter) so let's use websockets_iter(&self) here.

jedireza: fixed

fabricedesre:
r=me, please squash the commits before merging.

jedireza:
:raised_hands:

PR #64, opened 2016-02-22
Improve global state management and concurrency (Context, Controller).

mcav:
This commit improves and simplifies the way we manage global state, in collaboration with @jedireza.

_With this commit:_
- **`Controller` coordinates the interaction of services, servers, and events.** `ContextTrait` has been merged into `Controller`; `FoxBox` is the concrete implementation of `Controller`.
  
  (Why the name change? In practice, `Context` was acting more like a controller than a context; as was `Controller`. The term `Controller` is more precise.)
- **We no longer coarsely lock the global state.** Previously, we passed around `Arc<Mutex<Context>>` everywhere (aliased as `SharedContext`). When we needed to access the global context, we would coarsely `lock()` the mutex. This is inefficient and cumbersome. Instead, we only lock what we need to. 

This allows us to clone and share references to a generic `Controller` (i.e. `<T: Controller>`) safely and concurrently.

_NOTE:_ Git did not detect the rename of `context.rs` to `controller.rs`; the diff is slightly misleading in that regard. I'd recommend using the [split view](https://github.com/fxbox/foxbox/pull/64/files?diff=split) for easier reading.


src/service_router.rs
```diff
@@ -27,25 +24,13 @@ pub fn create<T: Send + Reflect + ContextTrait + 'static> (context: Shared<T>) -
...
     router.get(":service/:command", move |req: &mut Request| -> IronResult<Response> {
         // Call a function on a service.
-        let ctx = context2.lock().unwrap();
-
         let id = req.extensions.get::<Router>().unwrap().find("service").unwrap_or("");
-        match ctx.get_service(id) {
-            None => {
-                let mut response = Response::with(format!("No Such Service: {}", id));
-                response.status = Some(Status::BadRequest);
-                response.headers.set(AccessControlAllowOrigin::Any);
```
fabricedesre, line 40: I think you forgot to add that one back when you rebased.

fabricedesre:
Very nice!
One nit: Context was meant to carry all the "shared" things. Now we end up exposing things like start_tunnel() and stop_tunnel() to places that should not have access to that. We should break that up in a follow-up.

Please fix the cors header and we'll merge.

jedireza:
:raised_hands:

src/service_router.rs
mcav, line 40: Yeah, good catch.

Issue #65, opened 2016-02-22
Hook up service notifications to the websocket(s)


fabricedesre:
Right now the dummyservice is just printing to the console (https://github.com/fxbox/foxbox/blob/master/src/dummy_adapter.rs#L58) but it should actually send that to websocket listeners.
We need to add a new event kind in events.rs and go from there.

~~@jedireza, it's all yours.~~

PR #66, opened 2016-02-23
sending service bips to the event loop

jedireza:
Related to: #65

### DIFF
--- a/src/dummy_adapter.rs
+++ b/src/dummy_adapter.rs
@@ -20,11 +20,14 @@ struct DummyService {
 }
 
 impl DummyService {
+
     fn new(sender: EventSender, context: SharedContext, id: u32) -> DummyService {
         println!("Creating dummy service");
+
         let ctx_clone = context.clone();
         let ctx = ctx_clone.lock().unwrap();
         let service_id = Uuid::new_v4().to_simple_string();
+
         DummyService {
             properties: ServiceProperties {
                 id: service_id.clone(),
@@ -40,6 +43,7 @@ impl DummyService {
 }
 
 impl Service for DummyService {
+
     fn get_properties(&self) -> ServiceProperties {
         self.properties.clone()
     }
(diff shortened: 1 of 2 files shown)


jedireza:
r? @fabricedesre 

So this will send the service bips to the event loop, which automatically get sent to the websocket clients. Sorry about the whitespace noise, I wanted to clean up the readability while I was at it.

fabricedesre:
We need something that is not specific to the dummyservice (it should be remove at some point).

Instead, consider doing something that let any service send back either a string or a vec<u8> since this is what our ws library supports (http://housleyjk.github.io/ws-rs/ws/struct.Sender.html#method.send) I haven't checked if that maps to ArrayBuffers or ArrayBufferViews on the client side but that doesn't matter much. 

That all lead to really wonder if having a single websocket is a good option. For an app connected to a box an interacting with several services it's impossible to know who's sending what in general, and for binary payloads especially. Eg if 2 cameras sends pictures as ArrayBuffer, how do we distinguish them?

jedireza:
> We need something that is not specific to the dummyservice (it should be remove at some point).
> 
> Instead, consider doing something that let any service send back either a string or a vec since this is what our ws library supports (http://housleyjk.github.io/ws-rs/ws/struct.Sender.html#method.send) I haven't checked if that maps to ArrayBuffers or ArrayBufferViews on the client side but that doesn't matter much.

I added a new event type and sent it to the event loop and from there the controller dispatches messages to the websockets. We do need to get more specific about these events; what kinds we have, the data they contain and how they're consumed (ex: serialized to json for websocket client consumption).

> That all lead to really wonder if having a single websocket is a good option. For an app connected to a box an interacting with several services it's impossible to know who's sending what in general, and for binary payloads especially. Eg if 2 cameras sends pictures as ArrayBuffer, how do we distinguish them?

It should be trivial for the central websocket connection to handle most (90%+) of our event communication. Raw camera feeds are indeed a special case that should be handled outside of the central websocket connection.

fabricedesre:
Websocket payloads in ws-rs are either Strings or Vects of u8. We should also support any struct implementing Serialize and send it as a json representation. The current PR only does String.

I don't understand your argument about the central websocket connection. With a single websocket, we'll need to have some ID be part of every message, and that's really hard for any non-structured message. How do you plan to distinguish between 2 opaque vectors of bytes?

jedireza:
Our ideas about the implementation are drifting. Let's setup a time to sync-up. Closing this PR for now, there's not much to it.

Issue #61, opened 2016-02-22
Sign in screen


ferjm:
We need to implement the client side of https://github.com/fxbox/users/blob/master/doc/API.md#post-login

ferjm:
Closed by #67

PR #67, opened 2016-02-23
Sessions UI. r=fabrice

### LINKED ISSUE #60: Save session token after sign up
After https://github.com/fxbox/users/pull/32 we give the consumer a session token to use in following requests to authenticated endpoints. We need to save this token after the user creates the admin account.

ferjm:
Fixes #60 and #61


static/js/session.js
```diff
@@ -0,0 +1,82 @@
...
+  var SIGN_IN = '/users/login';
+
+  function sessionRequest(sessionInfo, endpoint) {
+    return new Promise(function(resolve, reject) {
+      var xhr = new XMLHttpRequest();
+      xhr.open('POST', endpoint, true);
+      xhr.onload = function() {
+        var response;
+        try {
+          response = JSON.parse(xhr.responseText);
```
fabricedesre, line 18: why not set xhr.responseType to `json` ?

fabricedesre:
If responseType works cross-browser (MDN could not tell me) let's change that. If not, let's merge.

ferjm:
IE and Opera Mini do not support it. http://caniuse.com/#feat=xhr2

PR #68, opened 2016-02-23
Fix build on rust-nightly: bump dependencies where appropriate.

hfiguiere:
Tested with rustc 1.8.0-dev (c8fc4817d 2016-02-22)

@fabricedesre r?

(I had to bump clippy too)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["fabrice <fabrice@desre.org>"]
 build = "build.rs"
 
 [dependencies]
-clippy = "0.0.41"
+clippy = "0.0.44"
 docopt = "0.6.78"
 docopt_macros = "0.6.80"
 env_logger = "0.3.2"
@@ -22,10 +22,10 @@ router = "0.1.0"
 rustc-serialize = "0.3"
 serde = "0.6.13"
 serde_json = "0.6.0"
-serde_macros = "0.6.13"
+serde_macros = "0.6.14"
 staticfile = "0.1.0"
 uuid = "0.1.18"
-ws = "0.4.5"
+ws = { git = "https://github.com/housleyjk/ws-rs.git", rev = "d154fc5" }
 
 [dev-dependencies]
 stainless = "0.1.4"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: rust
 rust:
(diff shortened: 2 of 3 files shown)


hfiguiere:
Seems that it fails with rustc from Feb 12... _sigh_

fabricedesre:
Thanks for looking at that.
When testing locally with a fresh rustc (0de3cace0 2016-02-23), bumping clippy to 0.0.44 let us build on x86, but cross compiling still fails because of errors in the nix crate (actually a single error u8 vs i8 that repeats in a bunch of places).
So we need the nix crate fixed first.

hfiguiere:
urgh. Ok, I can look at that.

hfiguiere:
(I don't have a cross compiler since I don't have anything to run the output on. Will set one up)

hfiguiere:
the problem is that ws pulls mio from crates that pull nix 0.4.2. while we pull mio from git that depends on nix-0.5.0 (git)

hfiguiere:
And nix 0.5.0 builds fine....

fabricedesre:
yes, this is why I used the git version of mio. I'm doing a fork of ws-rs to get us back on track.

hfiguiere:
I started looking at ws and with jthe mio API change.... it is not that trivial. I'll let you continue with that instead. And then we can see about landing this change.

hfiguiere:
I have ws working in my master branch https://github.com/hfiguiere/ws-rs

fabricedesre:
Ok cool. Can you do a PR to point foxbox/Cargo.toml to your fork?

hfiguiere:
@fabricedesre this one does (third commit). Or do you want something separate? I can squash things.

Cargo.toml
```diff
@@ -22,10 +22,10 @@ router = "0.1.0"
 rustc-serialize = "0.3"
 serde = "0.6.13"
 serde_json = "0.6.0"
-serde_macros = "0.6.13"
+serde_macros = "0.6.14"
 staticfile = "0.1.0"
 uuid = "0.1.18"
-ws = "0.4.5"
+ws = { git = "https://github.com/hfiguiere/ws-rs.git" }
```
fabricedesre: We can switch back to upstream https://github.com/housleyjk/ws-rs.git since they merged your fixes.

hfiguiere:
squahed and everything

PR #70, opened 2016-02-23
Try dependency caching on Travis

samgiles:
Wayy faster builds with this: the bottleneck really is compiling, but because we're using a fixed version of nightly it's ok to cache target. r? @fabricedesre

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: rust
 rust:
     - nightly-2016-02-13
+cache:
+    directories:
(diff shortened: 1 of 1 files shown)


samgiles:
This build: Elapsed time 1 min 48 sec
Cold build: Elapsed time 4 min 48 sec

jedireza:
With #71 we can probably get faster start times too :wink:

jedireza:
LGTM r+

PR #71, opened 2016-02-23
use containers on travis for faster builds

jedireza:
If we don't need `sudo` we should be able to use the container infrastructure on Travis.
https://docs.travis-ci.com/user/workers/container-based-infrastructure/


jedireza:
Related to #70

samgiles:
I thought we were already doing this, r+.

PR #72, opened 2016-02-24
Move the tunnel start outside of the shared controller

fabricedesre:


fabricedesre:
That let us not share the tunnel setup. Also, less code!

jedireza:
:+1:

PR #73, opened 2016-02-24
Make tests pass.

hfiguiere:
On my Fedora 23 laptop, wifi is wlp3s0
As per dmesg:
[    6.394537] iwlwifi 0000:03:00.0 wlp3s0: renamed from wlan0


hfiguiere:
@fabricedesre r? - not sure if we shouldn't try a more elegant solution in the long term.

fabricedesre:
Yeah, this is not great, but not a huge deal either.

PR #75, opened 2016-02-24
Add support for custom host name. r=fabrice

azasypkin:


.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
```
azasypkin, line 1: So I'm not really happy with these changes (especially the fact that we run foxbox with `sudo`), but don't have idea how to fix this easily :/ We need sudo for 2 things here:
- Install avahi libs + daemon (those are not in the Travis apt plugin white list);
- Run foxbox with root privileges to be able to update host name.

All this burden is caused by the way `multicast_dns` crate works. I had three choices how to implement it:
1. Write native Rust mDNS lib from scratch - neither sound real in the time-frame we have nor feasible for the project stage we are in :)
2. Write Rust wrapper around `avahi-common` and `avahi-core` libs and use it _instead of_ Avahi Daemon - so, no external dependencies as well as in option 1 - sounds like the best option, but as long term solution since we'll need to practically rewrite some parts of Avahi Daemon;
3. Write Rust wrapper around `avahi-common` and `avahi-client` (a simplifying C wrapper around the D-Bus API) libs and use it _in conjunction with_ running Avahi Daemon - the communication between daemon and `avahi-client` is built on top of DBus so +1 dependency - `dbus-1` - the easiest, but the heaviest option at the same time. Per Avahi docs this is the recommended way for desktop non-embedded environment. I still didn't dig enough, but what I can say for now is the Avahi Daemon requires privileged access for some operations - like changing host name. 

So as you know I chose option 3 and that requires `privileged access` :/

Any suggestions/objections/ideas/critique are welcome, @fabricedesre  :)

In the meantime I'll investigate if it's somehow possible to change host name without `sudo`

.travis.yml
```diff
@@ -22,4 +26,4 @@ script:
     - cargo build # Linter is also executed here
     - RUST_BACKTRACE=1 cargo test
     - jshint static/*.js test/selenium/*.js
-    - (cargo run &> /dev/null &) ; npm test
+    - (sudo ./target/debug/foxbox -n foxbox &> /dev/null &) ; npm test
```
azasypkin: I'm not sure where cargo and rust is installed by Travis, but I get `cargo: command is not found` if I run `cargo run` with sudo, do you have an idea how to fix this in a nicer way?

src/controller.rs
```diff
@@ -38,6 +38,7 @@ pub struct FoxBox {
 pub const DEFAULT_HTTP_PORT: u16 = 3000;
 const DEFAULT_WS_PORT: u16 = 4000;
 const DEFAULT_HOSTNAME: &'static str = "::"; // ipv6 default.
+const DEFAULT_DOMAIN: &'static str = ".local";
```
azasypkin, line 41: Not 100% sure that we need this constant at all, maybe it will be enough to just hardcode it in the `format!` call below

src/main.rs
```diff
@@ -80,6 +82,29 @@ Options:
         flag_register: Option<String>,
         flag_tunnel: Option<String>);
 
+/// Updates local host name with the provided host name string. If requested host name
```
azasypkin, line 94: Not sure if it the best place for this function, my main rationale was - `we need to update host name as early as possible in the bootstrap flow`

.travis.yml
fabricedesre, line 1: We really can't run the foxbox daemon as root... Instead of using avahi_client_set_host_name() did you try to add a cname record like in http://www.avahi.org/wiki/Examples/PythonPublishAlias ? This works for me without being root.
You don't even need to use dbus hopefully, just getting a group and adding the record should work.

azasypkin, line 1: > We really can't run the foxbox daemon as root... Instead of using avahi_client_set_host_name() did you try to add a cname record like in http://www.avahi.org/wiki/Examples/PythonPublishAlias ? This works for me without being root.

Nope haven't tried it, seems it's too naive and doesn't support collision resolution as we wanted. But I'll check if I can extend it to support collision resolution, thanks for the pointer!

> You don't even need to use dbus hopefully, just getting a group and adding the record should work.

This python script uses DBus/Avahi Daemon for managing entry groups, and if I'm not mistaken, not using DBus just means not using Avahi Daemon at all (and implement everything on our own instead) if we don't want troubles as it will basically mean having two "daemons" who is in charge of managing mDNS host.

So let me dig a bit, to understand if we can get rid of "root" requirements with option 3 or option 2 from message above is the only way.

Ah and just curious why we want to avoid using DBus by all means - is it heavy\buggy or ..?

fabricedesre, line 1: The CNAME record solution let you deal with collisions. You'll get that from the AvahiEntryGroupState that is passed back to callbacks like http://avahi.org/download/doxygen/publish_8h.html#afbd8ce431f0416537fc7b6101e0f9a1c

That being said I don't have anything against using DBus. I've only seen the argument about it being to heavy in the context of small embedded platforms, but that's not our case.

azasypkin, line 1: > The CNAME record solution let you deal with collisions. You'll get that from the AvahiEntryGroupState that is passed back to callbacks like http://avahi.org/download/doxygen/publish_8h.html#afbd8ce431f0416537fc7b6101e0f9a1c

Yeah I know, it just behaves weirdly for me - never complains about collision - will figure that out

azasypkin, line 1: While trying to figure out why entry group doesn't change state to `COLLISION` if CNAME conflicts with other .local domain on the network, I found out that avahi-daemon provides `with-avahi-priv-access-group=` flag that accepts name of the user group that will have ability to change host-name without sudo. By default it's `netdev` group.

So I'm curious if our `foxbox` user will be in a special group that we can feed to avahi-daemon's `with-avahi-priv-access-group=` while deployment, @fabricedesre? That would be really nice :)

fabricedesre, line 1: Yes, I don't see any problem with doing the appropriate user / group configuration.

azasypkin, line 1: Good, thanks - will see, how to configure avahi-daemon on Travis then.

.travis.yml
```diff
@@ -8,6 +8,15 @@ cache:
         - node_modules
         - $TRAVIS_BUILD_DIR/target
 
+# $BOX_HOST_NAME and $BOX_PORT are also used by Selenium tests.
+env:
+  - BOX_HOST_NAME=foxbox BOX_PORT=3000
```
azasypkin, line 13: Original reasoning behind that: custom host name is currently not supported for MacOS, so mac users will still be able to use `npm test` locally.

.travis.yml
```diff
@@ -8,6 +8,15 @@ cache:
...
+
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install -y avahi-daemon libavahi-client-dev libavahi-common-dev libavahi-common3 libdbus-1-dev
+  - sudo usermod -a -G netdev $USER # "netdev" group membership allows to set custom host name via avahi-daemon.
```
azasypkin, line 18: I added note to the Readme on how to configure any user/group instead of [netdev](https://wiki.ubuntu.com/Security/Privileges#Connect_to_wireless_and_ethernet_networks). Also looks like default `pi` user in RPi2 is a member of this group by default.

.travis.yml
azasypkin, line 1: Hey @fabricedesre, could you please do a first review of the entire PR?

I've ended up with editing `/etc/dbus-1/system.d/avahi-dbus.conf` to provide any user/user-group based avahi-daemon access approach we may need (explained below in README.md). Once we have some RPi deployment script/whatever instead we may install `avahi-daemon` from sources with `./configure --with-avahi-priv-access-group=any_group`. Or if someday we end up with Yocto there is already ready reciepe for it: http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-connectivity/avahi that configures access as well.

azasypkin, line 1: Ah and here is the PR for x-compilation https://github.com/fabricedesre/rustpi2/pull/2/files

src/main.rs
```diff
@@ -89,6 +91,29 @@ Options:
...
+/// is not available (occupied by anyone else on the same network) then collision
```
fabricedesre: nit: s/occupied/used

PR #76, opened 2016-02-24
Allow dead code until this is plumbed into the API

samgiles:
r? @cr


samgiles:
Rust follows this path through to the managed_process shutdown code.  If we used this function, all the warnings disappear.

cr:
Interesting. Works for me.

PR #77, opened 2016-02-24
[WIP] Local HTTPS Support

samgiles:
Configuring TLS:

Create a `certs/` directory containing your cert's private key and certificate:
The private_key and cert must be X509 files named `crt.pem` and `private_key.pem`

For example:

```
certs/
certs/test.knilxof.org/crt.pem
certs/test.knixof.org/private_key.pem
```

The directory that `certs` are found in can be configured in the configuration file under:  `foxbox.certficate_directory`, by default the software will look for the `certs` directory in your current working directory.

You can disable TLS entirely using the `--disable-tls` command line argument.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,4 +33,4 @@ script:
     - $TRAVIS_BUILD_DIR/tools/execute-unit-tests-with-coverage
     - jshint static/main/js/*.js static/setup/js/*.js test/selenium/*.js test/integration/lib/*.js test/integration/test/*.js
     - npm run test-integration # starts foxbox and kills it within the test script
-    - (cargo run -- -l $BOX_LOCAL_NAME -p $BOX_PORT &> /dev/null &) ; npm run test-selenium
+    - (cargo run -- -l $BOX_LOCAL_NAME -p $BOX_PORT --disable-tls &> /dev/null &) ; npm run test-selenium
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,13 +20,14 @@ foxbox_users = { git = "https://github.com/fxbox/users.git", rev = "3db32f0" }
 get_if_addrs = "0.3.1"
 hyper = "0.7.2"
 multicast_dns = { git = "https://github.com/fxbox/multicast-dns.git", rev = "a6e4bcc" }
-iron = "0.2.6"
 iron-cors = { git = "https://github.com/fxbox/iron-cors.git", rev = "137d42f" }
 libc = "0.2.7"
 log = "0.3"
 mio = { git = "https://github.com/carllerche/mio.git" }
 mount = "0.0.10"
 nix = { git = "https://github.com/nix-rust/nix.git", rev = "138080" } # Until 0.5.1 is released
+openssl = "0.7.6"
+openssl-sys = "0.7.6"
 router = "0.1.0"
 rust-crypto = "0.2.34"
 rustc-serialize = "0.3"
@@ -39,9 +40,14 @@ unicase = "1.3.0"
 time = "0.1"
 timer = "0.1.6"
 uuid = "0.1.18"
-url = "0.5.5"
+url = "0.5.7"
 ws = { git = "https://github.com/housleyjk/ws-rs.git", rev = "d154fc5" }
 xml-rs = "0.3.0"
+
+[dependencies.iron]
+version = "0.2.6"
+default-features = true
+features = ["ssl"]
 
 [dev-dependencies]
 stainless = "0.1.4"
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -24,12 +24,16 @@ use std::net::ToSocketAddrs;
 use std::sync::{ Arc, Mutex };
 use std::sync::atomic::{ AtomicBool, Ordering };
 use upnp::UpnpManager;
+use std::path::PathBuf;
+use tls::{ CertificateManager, TlsOption };
 use ws_server::WsServer;
 use ws;
 
 #[derive(Clone)]
 pub struct FoxBox {
     pub verbose: bool,
+    tls_option: TlsOption,
+    certificate_manager: CertificateManager,
     hostname: String,
     http_port: u16,
     ws_port: u16,
@@ -57,6 +61,10 @@ pub trait Controller : Send + Sync + Clone + Reflect + 'static {
     fn get_http_root_for_service(&self, service_id: String) -> String;
     fn get_ws_root_for_service(&self, service_id: String) -> String;
     fn http_as_addrs(&self) -> Result<IntoIter<SocketAddr>, io::Error>;
+
+    fn get_tls_enabled(&self) -> bool;
+    fn get_certificate_manager(&self) -> CertificateManager;
+
     fn add_websocket(&mut self, socket: ws::Sender);
     fn remove_websocket(&mut self, socket: ws::Sender);
     fn broadcast_to_websockets(&self, data: serde_json::value::Value);
@@ -72,10 +80,13 @@ impl FoxBox {
     pub fn new(verbose: bool,
                hostname: Option<String>,
                http_port: u16,
-               ws_port: u16) -> Self {
-
+               ws_port: u16,
+               tls_option: TlsOption) -> Self {
         let profile_service = ProfileService::new(None);
+
         FoxBox {
+            certificate_manager: CertificateManager::new(),
+            tls_option: tls_option,
             services: Arc::new(Mutex::new(HashMap::new())),
             websockets: Arc::new(Mutex::new(HashMap::new())),
             verbose: verbose,
@@ -101,8 +112,19 @@ impl Controller for FoxBox {
         {
             Arc::get_mut(&mut self.upnp).unwrap().start().unwrap();
         }
+
+        if self.get_tls_enabled() {
+            let certificate_directory = PathBuf::from(
+                self.config.get_or_set_default("foxbox", "certificate_directory", "certs/"));
+
+            // If this fails, it just means that no certificates will be configured, which
+            // shouldn't cause a crash.
+            self.certificate_manager.reload_from_directory(certificate_directory).unwrap_or(());
+        }
+
         HttpServer::new(self.clone()).start();
         WsServer::start(self.clone(), self.hostname.to_owned(), self.ws_port);
+
         let mut adapter_manager = AdapterManager::new(self.clone());
         adapter_manager.start();
         self.upnp.search(None).unwrap();
@@ -177,7 +199,8 @@ impl Controller for FoxBox {
     }
 
     fn get_http_root_for_service(&self, service_id: String) -> String {
-        format!("http://{}:{}/services/{}/", self.hostname, self.http_port, service_id)
+        let scheme = if self.get_tls_enabled() { "https" } else { "http" };
+        format!("{}://{}:{}/services/{}/", scheme , self.hostname, self.http_port, service_id)
     }
 
     fn get_ws_root_for_service(&self, service_id: String) -> String {
@@ -222,6 +245,14 @@ impl Controller for FoxBox {
     fn get_users_manager(&self) -> Arc<UsersManager> {
         self.users_manager.clone()
     }
+
+    fn get_certificate_manager(&self) -> CertificateManager {
+        self.certificate_manager.clone()
+    }
+
+    fn get_tls_enabled(&self) -> bool {
+        self.tls_option == TlsOption::Enabled
+    }
 }
 
 #[allow(dead_code)]
@@ -247,9 +278,10 @@ describe! controller {
 
     before_each {
         use stubs::service::ServiceStub;
+        use tls::TlsOption;
 
         let service = ServiceStub;
-        let controller = FoxBox::new(false, Some("foxbox".to_owned()), 1234, 5678);
+        let controller = FoxBox::new(false, Some("foxbox".to_owned()), 1234, 5678, TlsOption::Disabled);
     }
 
     describe! add_service {
@@ -269,10 +301,15 @@ describe! controller {
             }
         }
 
-        it "should create http root" {
+        it "should create https root if tls enabled and http root id disabled" {
             controller.add_service(Box::new(service));
             assert_eq!(controller.get_http_root_for_service("1".to_string()),
                        "http://foxbox.local:1234/services/1/");
+
+            let controller = FoxBox::new(false, Some("foxbox".to_owned()), 1234, 5678, TlsOption::Enabled);
+            controller.add_service(Box::new(service));
+            assert_eq!(controller.get_http_root_for_service("1".to_string()),
+                       "https://foxbox.local:1234/services/1/");
         }
 
         it "should create ws root" {
--- a/src/http_server.rs
+++ b/src/http_server.rs
@@ -3,8 +3,10 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use controller::Controller;
-use iron::{ AfterMiddleware, Chain, Handler, Iron, IronResult,
-            Request, Response };
+use hyper::net::{ NetworkListener };
+use iron::{ AfterMiddleware, Chain, Handler,
+            HttpServerFactory, Iron, IronResult, Request,
+            Response, ServerFactory };
 use iron_cors::CORS;
 use iron::error::{ IronError };
 use iron::method::Method;
@@ -13,7 +15,11 @@ use mount::Mount;
 use router::NoRoute;
 use service_router;
 use static_router;
+use std::net::SocketAddr;
 use std::thread;
+use tls::SniServerFactory;
+
+const THREAD_COUNT: usize = 8;
 
 struct Custom404;
 
@@ -79,11 +85,26 @@ impl<T: Controller> HttpServer<T> {
 
         let addrs: Vec<_> = self.controller.http_as_addrs().unwrap().collect();
 
-        thread::Builder::new().name("HttpServer".to_owned())
-                              .spawn(move || {
-            Iron::new(chain).http(addrs[0]).unwrap();
-        }).unwrap();
+        if self.controller.get_tls_enabled() {
+            let mut certificate_manager = self.controller.get_certificate_manager();
+            let server_factory = SniServerFactory::new(&mut certificate_manager);
+            start_server(addrs, chain, server_factory);
+        } else {
+            start_server(addrs, chain, HttpServerFactory {});
+        }
     }
+}
+
+fn start_server<TListener, T>(addrs: Vec<SocketAddr>, chain: Chain, factory: T)
+    where TListener: NetworkListener + Send + 'static,
+          T: ServerFactory<TListener> + Send + 'static {
+
+    thread::Builder::new().name("HttpServer".to_owned())
+                          .spawn(move || {
+        Iron::new(chain)
+             .listen_with(addrs[0], THREAD_COUNT, &factory, None)
+             .unwrap();
+    }).unwrap();
 }
 
 #[cfg(test)]
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,6 +29,7 @@ extern crate env_logger;
 extern crate foxbox_adapters;
 extern crate foxbox_taxonomy;
 extern crate foxbox_users;
+extern crate hyper;
 #[macro_use]
 extern crate iron;
 extern crate iron_cors;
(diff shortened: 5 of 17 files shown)


samgiles:
@fabricedesre can I get an initial review? Or suggest someone else.

I'm still trying to figure out how best to plumb in the hostname and certs from the controller as we'll need them before the server is started with this PR - we can change that, but maybe in a later PR.  Also, following this we'll want to use LetsEncrypt, and pre-generate, or generate the certs on demand.

fabricedesre:
I took avery quick look, but could you have a CertificateManager that you pass to the HttpServer instead of your current code?

@ferjm do you mind taking a look?

src/controller.rs
```diff
@@ -47,6 +52,10 @@ pub trait Controller : Send + Sync + Clone + Reflect + 'static {
     fn get_service_properties(&self, id: String) -> Option<ServiceProperties>;
     fn services_count(&self) -> usize;
     fn services_as_json(&self) -> Result<String, serde_json::error::Error>;
+    fn get_local_tls_key(&self)  -> PathBuf;
+    fn get_local_tls_crt(&self)  -> PathBuf;
+    fn get_remote_tls_key(&self) -> PathBuf;
+    fn get_remote_tls_crt(&self) -> PathBuf;
```
ferjm: As Fabrice mentioned, it'd be nice to move all this to a CertificateManager.

Cargo.toml
```diff
@@ -13,11 +13,13 @@ foxbox_users = { git = "https://github.com/fxbox/users.git", rev = "d00552c" }
 get_if_addrs = "0.3.1"
 hyper = "0.7.2"
 multicast_dns = { git = "https://github.com/fxbox/multicast-dns.git", rev = "0f211f0" }
-iron = "0.2.6"
+iron = { git = "https://github.com/samgiles/iron.git", rev="ssl-context" }
```
ferjm: Is this only for testing purpose or do we plan to use a fork of iron?

src/controller.rs
```diff
@@ -63,13 +72,25 @@ impl FoxBox {
...
 
+        let hostname = hostname.map_or(DEFAULT_HOSTNAME.to_owned(), |name| {
+            format!("{}{}", name, DEFAULT_DOMAIN)
+        });
+
+        let local_crt = PathBuf::from(format!("certs/server/{}-server.crt.pem", hostname));
+        let local_key = PathBuf::from(format!("certs/server/{}-server.key.pem", hostname));
+
+        let remote_crt = PathBuf::from(format!("certs/server/{}-server.crt.pem", "test.knilxof.org"));
+        let remote_key = PathBuf::from(format!("certs/server/{}-server.key.pem", "test.knilxof.org"));
```
ferjm: Could we get this string from the `tunnel` command line option like we do with the box hostname?

src/controller.rs
```diff
@@ -217,7 +254,7 @@ describe! controller {
         it "should create http root" {
```
ferjm: uber-nit: s/http/https :)

src/http_server.rs
```diff
@@ -5,22 +5,60 @@
 use controller::Controller;
 use service_router;
 use foxbox_users::users_router::UsersRouter;
-use iron::Iron;
+use hyper::error::Error as HyperError;
+use hyper::server::Server;
+use hyper::net::{ HttpsListener, Ssl, Openssl };
```
ferjm: nit: let's try to keep alphabetical order, please.

src/http_server.rs
```diff
@@ -5,22 +5,60 @@
...
+    }
+}
+
+impl ServerFactory<HttpsListener<Openssl>> for HttpsServerFactory<Openssl> {
+
+    fn protocol(&self) -> Protocol {
+        Protocol::Https
+    }
+
+    fn create_server(&self, sock_addr: SocketAddr) -> Result<Server<HttpsListener<Openssl>>, HyperError> {
```
ferjm: nit: in general, lines shorter than 80 chars, please.

src/http_server.rs
```diff
@@ -5,22 +5,60 @@
...
 impl<T: Controller> HttpServer<T> {
     pub fn new(controller: T) -> Self {
-        HttpServer { controller: controller }
+        HttpServer {
+            controller: controller,
+        }
     }
 
-    pub fn start(&mut self) {
+    pub fn start(&mut self, cert: PathBuf, key: PathBuf) {
```
ferjm: As suggested, you need to pass here the CertManager.

src/http_server.rs
```diff
@@ -30,9 +68,19 @@ impl<T: Controller> HttpServer<T> {
 
         let addrs: Vec<_> = self.controller.http_as_addrs().unwrap().collect();
 
-        thread::Builder::new().name("HttpServer".to_owned())
+        debug!("Starting server with cert: {:?} and key: {:?}", cert, key);
+
+        let mut config = TlsSniConfig::new();
+        config.add_ssl_cert("foxbox.local".to_owned(), cert, key).unwrap();
+        config.add_ssl_cert("test.knilxof.org".to_owned(), self.controller.get_remote_tls_crt(), self.controller.get_remote_tls_key()).unwrap();
```
ferjm: Do we need the `add_ssl_cert` function? Could we instead pass to `init_ssl_context` the list  of hostnames, certs and keys?

src/tls_config.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    pub fn new() -> Self {
+        TlsSniConfig {
+            ssl_hosts: HashMap::new()
+        }
+    }
+
+    pub fn context(&self) -> Result<SslContext, Error> {
+        for ctx in self.ssl_hosts.values() {
+            return Ok(ctx.clone());
```
ferjm: IIUC here we are always returning one context only. Is that what you wanted?

src/http_server.rs
ferjm: Actually, maybe we can do all the init process in `TlsSniConfig::new()` passing the list here.

ferjm:
Great work Sam!

I think @michielbdejong has a more solid understanding of what's needed here. Michiel, Could you take a look, please? Thanks!

src/tls_config.rs
samgiles: I've added a comment.  tl;dr; Yes I wanted that.

The comment explains why this is :)

src/http_server.rs
michielbdejong: Why use SNI at all on the Foxbox? we can just listen with two different https servers, right?

michielbdejong:
side note: I got https with self-signed certs, including QR, mDNS, and HSS, working with a nodejs proxy in https://github.com/fxbox/foxbox/pull/129 - may be useful as a reference.

src/http_server.rs
samgiles: Because this is a more efficient solution and we can reuse port 443? :)  + you're not restricted to only two names, not sure if that's beneficial.

Cargo.toml
samgiles: We're gonna need to use a fork, I've made a PR upstream.

src/http_server.rs
michielbdejong: On the local network you want to expose port 443, yes. But to the tunnel client, it would make more sense to me to expose either a high port or a unix socket (also depending on how we implement the tunnel client).

But yes, if we ever want more than one local and one bridged domain name, then you're right that SNI is useful. Could be the case if we ever want to activate https://github.com/fxbox/RFC/pull/4 and https://github.com/fxbox/RFC/pull/5 simultaneously.

Doesn't really matter though, I was just curious. :)

michielbdejong:
Latest scripts for DNS/TLS are in https://github.com/fxbox/foxbox/pull/163/. The way it works:
- [create a self-signed cert](https://github.com/fxbox/foxbox/pull/163/files#diff-ba25ea553e3ff05620224368284e4ae3R38)
- [set up DNS](https://github.com/fxbox/foxbox/pull/163/files#diff-dcaa0a22b76e30da2cfe0d10e94603aeR7)
- [register LetsEncrypt cert](https://github.com/fxbox/foxbox/pull/163/files#diff-dcaa0a22b76e30da2cfe0d10e94603aeR14) and generate QR code for the public-local URL.
- [Announce mDNS](https://github.com/fxbox/foxbox/pull/163/files#diff-ba25ea553e3ff05620224368284e4ae3R131) and generate QR code for the self-signed URL.
- Serve https with the self-signed and the LetsEncrypt-signed certs.
- (to do:) apart from local IP address, post all URLs to the registration server, something like:

``` js
{
  "self-signed": "https://ba25ea553e3ff05620224368284e4ae3.self-signed/a6450c08893/",
  "local": "https://dcaa0a22b76e30da2cfe0d10e94603ae.box.knilxof.org/",
  "mirrors": [
    "https://a.dcaa0a22b76e30da2cfe0d10e94603ae.box.knilxof.org/",
    "https://b.dcaa0a22b76e30da2cfe0d10e94603ae.box.knilxof.org/"
  ],
  "remote": "https://dcaa0a22b76e30da2cfe0d10e94603ae.box.knilxof.org/"
}
```

samgiles:
This requires the API compatible fork of Iron to allow custom SslContext support, see: https://github.com/iron/iron/pull/433

A few of our sub-dependencies also rely on Iron, it is not enough to simply replace iron with a fork because of Cargo's dependency resolution, see: https://github.com/rust-lang/cargo/pull/2385

fabricedesre:
@samgiles I guess the alternative is to fork all the sub-dependencies that depend on Iron?

samgiles:
I just thought: I'm doing this locally by setting a `.cargo/config` with a path override.  It overrides 'iron' with the implementation found at the configured path: 

My `.cargo/config` looks like this:

```
paths = ["/Users/sam/code/rust/iron"]
```

I could add the iron into the repository and commit a config.  @fabricedesre Does that sound reasonable as a work around?

fabricedesre:
Yes, but can we override with a http or git url?

samgiles:
No, it needs to be a local path.

fabricedesre:
So, something like:
paths = ["./iron-fork"]

.travis.yml
```diff
@@ -33,4 +33,4 @@ script:
     - $TRAVIS_BUILD_DIR/tools/execute-unit-tests-with-coverage
     - jshint static/main/js/*.js static/setup/js/*.js test/selenium/*.js test/integration/lib/*.js test/integration/test/*.js
     - npm run test-integration # starts foxbox and kills it within the test script
-    - (cargo run -- -l $BOX_LOCAL_NAME -p $BOX_PORT &> /dev/null &) ; npm run test-selenium
+    - (cargo run -- -l $BOX_LOCAL_NAME -p $BOX_PORT --disable-tls &> /dev/null &) ; npm run test-selenium
```
michielbdejong, line 36: Why test with `--disable-tls`?

samgiles, line 36: We don't have a solution for testing with self signed certs or real certs yet.  We don't lose anything by disabling TLS for the tests for the time being.

In the future we need to figure out how to test properly (and cross platform) with certs.  Maybe @JohanLorenzo or @npark-mozilla have some thoughts.

src/tls/certificate_manager.rs
```diff
@@ -0,0 +1,234 @@
...
+            ssl_hosts: Arc::new(RwLock::new(HashMap::new())),
+            context_provider: None,
+        }
+    }
+
+    pub fn set_context_provider(&mut self, context_provider: Arc<Mutex<Box<SslContextProvider>>>) {
+        if self.context_provider.is_none() {
+           self.context_provider = Some(context_provider);
+           self.notify_provider();
+        }
```
ferjm, line 73: Should we throw (or log an error) if the context provider is set more than once?

src/tls/https_server_factory.rs
```diff
@@ -0,0 +1,47 @@
...
+use tls::certificate_manager::CertificateManager;
+use tls::ssl_context::SniSslContextProvider;
+
+pub struct SniServerFactory<S: Ssl + Clone + Send> {
+    ssl: S
+}
+
+impl SniServerFactory<Openssl> {
+    pub fn new(ssl: &mut CertificateManager) -> Self {
+        use tls::SslContextProvider;
```
ferjm: Any reason to not move this with the rest of uses?

src/tls/certificate_manager.rs
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    pub fn reload_from_directory(&mut self, directory: PathBuf) -> Result<(), Error> {
+        let certificates =  try!(create_records_from_directory(&directory));
+        {
+            let mut current_hosts = self.ssl_hosts.write().unwrap();
```
ferjm: Shouldn't we handle the case where the RwLock is poisoned? (here and everywhere in this file)

src/tls/certificate_manager.rs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        self.notify_provider();
+    }
+
+    #[allow(dead_code)]
+    pub fn get_certificate(&self, hostname: String) -> Option<CertificateRecord> {
+        let ssl_hosts = self.ssl_hosts.read().unwrap();
+
+
```
ferjm: uber-nit: remove extra line

src/tls/certificate_record.rs
```diff
@@ -0,0 +1,16 @@
...
+
+/// Defines a certificate, including the hostname it is for,
+/// the private key file and the certificate file.
+#[derive(Clone, Hash, Eq, PartialEq)]
+pub struct CertificateRecord {
+    pub hostname: String,
+    pub private_key_file: PathBuf,
+    pub cert_file: PathBuf,
+    // TODO: We should probably keep a hash of the
+    // file contents as part of the CertRecord
```
ferjm: File a bug for that and add a reference to it here, please.

src/controller.rs
```diff
@@ -72,10 +80,12 @@ impl FoxBox {
     pub fn new(verbose: bool,
                hostname: Option<String>,
                http_port: u16,
-               ws_port: u16) -> Self {
-
+               ws_port: u16, tls_enabled: bool) -> Self {
```
ferjm: uber-nit: follow the pattern of previous params by moving `tls_enabled` to the next line

src/controller.rs
```diff
@@ -101,8 +111,19 @@ impl Controller for FoxBox {
...
             Arc::get_mut(&mut self.upnp).unwrap().start().unwrap();
         }
+
+        if self.get_tls_enabled() {
+            let certificate_directory = PathBuf::from(
+                self.config.get_or_set_default("foxbox", "certificate_directory", "certs/"));
+
+            // If this fails, it just means that no certificates will be configured, which
+            // shouldn't cause a crash.
+            self.certificate_manager.reload_from_directory(certificate_directory).unwrap_or(());
```
ferjm, line 122: Do we fallback to http if no certs are configured?

src/controller.rs
```diff
@@ -177,7 +198,8 @@ impl Controller for FoxBox {
     }
 
     fn get_http_root_for_service(&self, service_id: String) -> String {
-        format!("http://{}:{}/services/{}/", self.hostname, self.http_port, service_id)
+        let scheme = if self.get_tls_enabled() { "https" } else { "http" };
+        format!("{}://{}:{}/services/{}/", scheme , self.hostname, self.http_port, service_id)
```
ferjm, line 203: If line 121 errors do we still keep the `https` scheme?

src/controller.rs
```diff
@@ -249,7 +279,7 @@ describe! controller {
         use stubs::service::ServiceStub;
 
         let service = ServiceStub;
-        let controller = FoxBox::new(false, Some("foxbox".to_owned()), 1234, 5678);
+        let controller = FoxBox::new(false, Some("foxbox".to_owned()), 1234, 5678, false);
```
ferjm: uber-nit: `, false /* tls_enabled */);`

src/controller.rs
```diff
@@ -269,12 +299,19 @@ describe! controller {
...
 
-        it "should create http root" {
+        it "should create http root if tls disabled" {
             controller.add_service(Box::new(service));
             assert_eq!(controller.get_http_root_for_service("1".to_string()),
                        "http://foxbox.local:1234/services/1/");
         }
 
+        it "should create https root if tls enabled" {
+            let controller = FoxBox::new(false, Some("foxbox".to_owned()), 1234, 5678, true);
```
ferjm: Same thing: `, true /* tls_enabled */)`

src/main.rs
```diff
@@ -104,6 +108,7 @@ Options:
     -i, --iface <iface>      Specify the local IP interface.
     -t, --tunnel <tunnel>    Set the tunnel endpoint's hostname. If omitted, the tunnel is disabled.
     -s, --tunnel-secret <secret>    Set the tunnel shared secret. [default: secret]
+        --disable-tls               Run as a plain HTTP server, disabling encryption.
     -u, --remote-name <hostname>    Set remote hostname. This the URL to access the box through the bridge. If omitted, the tunnel is disabled
```
ferjm: I didn't know you could have options without the simplified version :) Could you take the chance to also remove the `-u` option, please? Thanks!

src/service_router.rs
```diff
@@ -155,7 +155,7 @@ describe! service_router {
         use iron_test::request;
         use mount::Mount;
 
-        let controller = FoxBox::new(false, Some("localhost".to_owned()), 1234, 5678);
+        let controller = FoxBox::new(false, Some("localhost".to_owned()), 1234, 5678, false);
```
ferjm: `false /* tls_enabled */`

src/tls/ssl_context.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+#[derive(Clone)]
+pub struct SniSslContextProvider {
+    main_context: Arc<RwLock<SslContext>>
+}
+
+impl SslContextProvider for SniSslContextProvider {
+    fn context(&self) -> Result<SslContext, Error> {
+        Ok(self.main_context.read().unwrap().clone())
```
ferjm: Same comment about handling poisoned RwLocks

src/tls/ssl_context.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        for record in configured_hosts.values() {
+            debug!("Creating SslContext for {}", record.hostname);
+            let ssl_context = create_ssl_context(&record.cert_file, &record.private_key_file);
+
+            if ssl_context.is_ok() {
+                let ssl_context = ssl_context.unwrap();
+                new_ssl_hosts.insert(record.hostname.clone(), ssl_context);
+            } else {
+                error!("Failed to configure SslContext for {} using key: {:?}, and cert: {:?}", &record.hostname, &record.cert_file, &record.private_key_file);
```
ferjm: This is a long line :)

Please, make sure that all lines are shorter than 80 chars, please.

ferjm:
This is an excellent job Sam!

I only added a few minor comments. Please, rebase and r=me with the comments addressed.

.travis.yml
michielbdejong, line 36: @samgiles Just run selenium with a Firefox profile that has the security exception for the certificate you're using in the tests, see e.g. https://blog.codecentric.de/en/2010/02/selenium-and-ssl-certificates/.

I don't agree we don't lose anything - currently, we have a http API _with_ a test. Once we merge this, we will have a https API _without_ a test. :(

If some upstream change in Iron causes our API server to break on https but not on http, we will not be able to see that in our tests. Right?

Can you create a follow-up ticket to repair the selenium test?

src/service_router.rs
fabricedesre: Even better, use an enum { TlsEnabled, TlsDisabled } instead of a bool

src/tls/https_server_factory.rs
samgiles: Not particularly, I think I did this so that the trait would only be available in this scope, but that doesn't add anything right now.

.travis.yml
nojunpark, line 36: I created this issue (https://github.com/fxbox/foxbox/issues/221) to create https test. Would like to find out how can I store / use certificates, will check the link you provided first on above comment.

Looks like for nodejs test, I need the .pem format of the cert/key/ca of  the generated certificate somehow

src/tls/certificate_manager.rs
samgiles: I've opted to ignore any poison, what do you think of the solution I've used?

src/controller.rs
samgiles, line 122: No, nothing is configured and the box won't respond to requests, this isn't too bad because once we've got LetsEncrypt support, we can update and add certs once it's already running.  Requiring stuff on disk to use TLS might be bad.

src/controller.rs
samgiles, line 203: Yes, you're either using TLS without any configured certs, in which case TLS won't work nor will http, or you're using plain HTTP.

PR #78, opened 2016-02-24
Adding Philips Hue adapter

cr:
Here's a prototype implementation of the Philips Hue adapter. See the included `huedemo.sh` script for a quick intro to the prototype API. Neither adapter nor API are "done" in any other way than that they cover our user stories. This PR is mainly for getting _some help with writing tests_.

**Before you pair with a Hue bridge using this adapter, be warned that the current code leaves your Hue bridge with a predictable, though non-trivial, API token which presents a considerable security risk.** If you know how to remove API tokens from the bridge, please tell me how.

Run with `RUST_LOG=info cargo run` or `RUST_LOG=debug cargo run` to see the adapter's console output. The Hue adapter will attempt to pair with all local Hue bridges announced via _nUPnP_ for 120 seconds each, then gives up. Pairing requests are published via _AdapterInfo_ events and are also announced on the console.

### DIFF
--- a/src/adapters/mod.rs
+++ b/src/adapters/mod.rs
@@ -1,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+mod philips_hue;
+
+use controller::Controller;
+use service::ServiceAdapter;
+
+pub struct AdapterManager<T> {
+    controller: T
+}
+
+impl<T: Controller> AdapterManager<T> {
+    pub fn new(controller: T) -> Self {
+        debug!("Creating Adapter Manager");
+        AdapterManager { controller: controller }
+    }
+
+    pub fn start(&self) {
+        // Start all the adapters.
+        philips_hue::PhilipsHueAdapter::new(self.controller.clone()).start();
+	}
+}
--- a/src/adapters/philips_hue/api/mod.rs
+++ b/src/adapters/philips_hue/api/mod.rs
@@ -1,0 +1,221 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#![allow(dead_code)]
+
+extern crate serde_json;
+
+pub mod structs;
+
+use adapters::philips_hue::http;
+use std;
+use std::collections::BTreeMap;
+use std::error::Error;
+use std::hash::{Hash, SipHasher, Hasher};
+
+
+#[derive(Debug, Hash)]
+struct HueHubApiToken {
+    iv: String,
+    id: String,
+}
+
+impl HueHubApiToken {
+    // This implementation offers 64 bit
+    fn as_hash(&self) -> String {
+        let mut hasher = SipHasher::new();
+        self.hash(&mut hasher);
+        format!("{:016x}", hasher.finish()).to_owned()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct HueHubApi {
+    pub id: String,
+    pub ip: String,
+    pub token: String,
+}
+
+impl std::fmt::Display for HueHubApi {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "Hue Bridge id:{} at {:?}", self.id, self.ip)
+    }
+}
+
+impl HueHubApi {
+    pub fn new(id: String, ip: String) -> HueHubApi {
+        // TODO: Generate a unique but reproducible access token.
+        // The token must not be guessable by an outsider.
+
+        // TODO: replace with a loooong stored random string
+        let stored_random = String::from("4"); // chosen by fair dice roll.
+        let token_gen = HueHubApiToken { iv: stored_random, id: id.clone() };
+        let token = format!("foxbox-{}", token_gen.as_hash());
+        HueHubApi { id: id, ip: ip, token: token }
+    }
+
+    pub fn get(&self, cmd: String) -> Result<String, Box<Error>> {
+        let url = format!("http://{}/api/{}/{}", self.ip, self.token, cmd);
+        debug!("GET request to Philips Hue bridge {}: {}", self.id, url);
+        let content = http::get(url);
+        debug!("Philips Hue API response: {:?}", content);
+        content
+    }
+
+    pub fn post(&self, cmd: String, data: String) -> Result<String, Box<Error>> {
+        let url = format!("http://{}/api/{}/{}", self.ip, self.token, cmd);
+        debug!("POST request to Philips Hue bridge {}: {} data: {}", self.id, &url, &data);
+        let content = http::post(url, data);
+        debug!("Philips Hue API response: {:?}", content);
+        content
+    }
+
+    pub fn post_unauth(&self, cmd: String, data: String) -> Result<String, Box<Error>> {
+        let url = format!("http://{}/{}", self.ip, cmd);
+        debug!("POST request to Philips Hue bridge {}: {} data: {}", self.id, &url, &data);
+        let content = http::post(url, data);
+        debug!("Philips Hue API response: {:?}", content);
+        content
+    }
+
+    pub fn put(&self, cmd: String, data: String) -> Result<String, Box<Error>> {
+        let url = format!("http://{}/api/{}/{}", self.ip, self.token, cmd);
+        debug!("PUT request to Philips Hue bridge {}: {} data: {}", self.id, &url, &data);
+        let content = http::put(url, data);
+        debug!("Philips Hue API response: {:?}", content);
+        content
+    }
+
+    pub fn is_available(&self) -> bool {
+        let url = format!("http://{}/", self.ip);
+        let content = http::get(url);
+        match content {
+            Ok(value) => {
+                value.contains("hue personal wireless lighting")
+            },
+            Err(_) => {
+                false
+            }
+        }
+    }
+
+    pub fn get_settings(&self) -> String {
+        // [{"error":{"type":1,"address":"/","description":"unauthorized user"}}]
+        self.get("".to_owned()).unwrap_or("".to_owned()) // TODO no unwrap
+    }
+
+    pub fn is_paired(&self) -> bool {
+        let settings = self.get_settings();
+        !settings.contains("unauthorized user")
+    }
+
+    pub fn try_pairing(&self) -> bool {
+        // [{"success":{"username":"foxboxb-001788fffe25681a"}}]
+        // [{"error":{"type":101,"address":"/","description":"link button not pressed"}}]
+        let url = "api".to_owned();
+        let req = format!("{{\"username\": \"{}\", \"devicetype\": \"foxbox_hub\"}}", self.token);
+        let response = self.post_unauth(url.clone(),
+            req.clone()).unwrap_or("".to_owned()); // TODO: no unwrap
+        response.contains("success")
+    }
+
+    pub fn get_lights(&self) -> Vec<HueLight> {
+        let mut lights: Vec<HueLight> = Vec::new();
+        let url = "lights".to_owned();
+        let res = self.get(url).unwrap(); // TODO: remove unwrap
+        let json: BTreeMap<String, structs::HueHubSettingsLightEntry> = structs::parse_json(&res).unwrap(); // TODO: no unwrap
+
+        for (key,_) in json {
+            let light = HueLight::new(self.id.clone(), self.ip.clone(), key);
+            lights.push(light);
+        }
+
+        lights
+    }
+
+    pub fn get_light_status(&self, id: &String) -> structs::HueHubSettingsLightEntry {
+        let url = format!("lights/{}", id);
+        let res = self.get(url).unwrap(); // TODO: remove unwrap
+        structs::parse_json(&res).unwrap() // TODO no unwrap
+    }
+
+    pub fn set_light_color(&self, light_id: &String, hue: u32, sat: u32, val: u32, on: bool) {
+        let url = format!("lights/{}/state", light_id);
+        let cmd = format!("{{\"hue\":{}, \"sat\":{}, \"bri\":{}, \"on\": {}}}",
+            hue, sat, val, on);
+        let _ = self.put(url, cmd);
+    }
+
+}
+
+#[derive(Debug)]
+pub struct LightState {
+    pub hue: f32,
+    pub sat: f32,
+    pub val: f32,
+    pub on: bool,
+}
+
+impl LightState {
+    pub fn new(hue: f32, sat: f32, val: f32, on: bool) -> LightState {
+        LightState { hue: hue, sat: sat, val: val, on: on }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct HueLight {
+    pub hub_id: String, // TODO: move to HueHubApi reference
+    pub hub_ip: String,
+    pub hue_id: String,
+}
+
+impl HueLight {
+    pub fn new(hub_id: String, hub_ip: String, light_id: String) -> HueLight {
+        HueLight {
+            hub_id: hub_id,
+            hub_ip: hub_ip,
+            hue_id: light_id,
+        }
+    }
+
+    pub fn get_settings(&self) -> structs::HueHubSettingsLightEntry {
+        let api = HueHubApi::new(self.hub_id.clone(), self.hub_ip.clone());
+        api.get_light_status(&self.hue_id)
+    }
+
+    pub fn get_state(&self) -> LightState {
+        // TODO: Work with api reference instead of instantiating one ad-hoc.
+        // Created issues in muti-treading.
+        let api = HueHubApi::new(self.hub_id.clone(), self.hub_ip.clone());
+        let ls = api.get_light_status(&self.hue_id);
+        let hue: f32 = ls.state.hue as f32 / 65535f32 * 360f32;
+        let sat: f32 = ls.state.sat as f32 / 254f32;
+        let val: f32 = ls.state.bri as f32 / 254f32;
+        let on = ls.state.on;
+
+        LightState::new(hue, sat, val, on)
+    }
+
+    pub fn set_state(&self, state: LightState) {
+        // Ensure valid ranges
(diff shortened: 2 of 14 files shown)


src/adapter/mod.rs
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
fabricedesre: Maybe rename the directory to `adapters` ?

src/adapter/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+    controller: T
+}
+
+impl<T: Controller> AdapterManager<T> {
+    pub fn new(controller: T) -> Self {
+        debug!("Creating Adapter Manager");
+        AdapterManager { controller: controller }
+    }
+
+	pub fn start(&self) {
```
fabricedesre: nit: this function's indentation is broken

src/adapter/philips_hue/api/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+
+pub mod structs;
+
+use adapter::philips_hue::http;
+use std;
+use std::collections::BTreeMap;
+use std::error::Error;
+use std::time::Duration;
+use std::thread;
+use std::hash::{Hash, SipHasher, Hasher};
```
fabricedesre: nit: we use the `std::hash::{ Hash, SipHasher, Hasher };` form elsewhere.

src/adapter/philips_hue/api/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub fn is_paired(&self) -> bool {
+        let s = self.get_settings();
+        !s.contains("unauthorized user")
+    }
+
+    pub fn pair(&self) -> String {
+        // [{"success":{"username":"foxboxb-001788fffe25681a"}}]
+        // [{"error":{"type":101,"address":"/","description":"link button not pressed"}}]
+        let url = format!("api");
+        let req = format!("{{\"username\": \"{}\", \"devicetype\": \"foxbox_hub\"}}", self.token);
```
fabricedesre: that's fragile in general because you don't escape potential special characters from the token.

src/adapter/philips_hue/api/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+                break;
+            }
+            thread::sleep(Duration::from_millis(1000));
+        }
+        content
+    }
+
+    pub fn get_lights(&self) -> Vec<HueLight> {
+        let mut lights: Vec<HueLight> = Vec::new();
+        let url = format!("lights");
```
fabricedesre: do you really need the format here?

src/adapter/philips_hue/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+
+        thread::spawn(move || {
+            controller.send_event(
+                EventData::AdapterStart {
+                    name: "Philips Hue Service Adapter".to_owned()
+                }).unwrap();
+
+            debug!("nUPnP reported Philips Hue bridges: {:?}", hubs);
+            'pairing_check: for hub in hubs {
+                hub.is_available();
```
fabricedesre: so we don't check if the hub is actually available?

src/adapter/philips_hue/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    controller: T,
+    hubs: Vec<api::HueApi>,
+}
+
+impl<T: Controller> PhilipsHueAdapter<T> {
+    pub fn new(controller: T) -> Self {
+        debug!("Creating Philips Hue adapter");
+        PhilipsHueAdapter { name: "PhilipsHueAdapter".to_owned(),
+                       controller: controller,
+                       hubs: nupnp::query(),
```
fabricedesre: This looks like a blocking call on the main thread. If so, we can't do that, and we should get the hubs in the thread spawned in start()

src/adapter/philips_hue/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+                hub.is_available();
+                info!("Philips Hue bridge {} at {}", hub.id, hub.ip);
+
+                // Querying the settings fails when not paired with the bridge
+                if !hub.is_paired() {
+                    controller.send_event(
+                        EventData::AdapterInfo {
+                            info: format!("Push the pairing button on your Philips Hue bridge with ID {}",
+                            hub.id)
+                        }).unwrap();
```
fabricedesre: I like the idea of sending adapter-level notifications! We'll need to add a bit more details to identify the adapter in the long run though.

src/adapter/philips_hue/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+                    let pairing_result = hub.pair();
+                    if !pairing_result.contains("success") {
+                        info!("Giving up on pairing with Philips Hue bridge {}", hub.id);
+                        break 'pairing_check;
```
fabricedesre: That `break` makes all this code hard to read. Please refactor this part.

src/adapter/philips_hue/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+                    controller.send_event(
+                        EventData::AdapterInfo {
+                             info: format!("Pairing with Philips Hue bridge {} successfull",
+                             hub.id)
+                         }).unwrap();
+                }
+
+                // Extract and log some info
+                let setting = hub.get_settings();
+                let hs = HueSettings::new(&setting).unwrap(); // TODO: no unwrap
```
fabricedesre: Renaming HueSettings to HubSettings may help (or Bridge setting if this is the right terminology for this product).

src/adapter/philips_hue/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+
+                let lights = hub.get_lights();
+                for light in lights {
+                    debug!("Spawning service for {:?}", light);
+                    id += 1;
+                    let service = PhilipsHueService::new(controller.clone(), id, light);
+                    let service_id = service.get_properties().id;
+                    service.start();
+                    controller.add_service(Box::new(service));
+                    controller.send_event(EventData::ServiceStart { id: service_id }).unwrap();
```
fabricedesre: Note to self: add_service() should send the ServiceStart event itself, not the individual Service implementation.

src/adapter/philips_hue/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+        });
+    }
+
+    fn stop(&self) {
+        info!("Stopping Philips Hue adapter");
+    }
+
+}
+
+struct PhilipsHueService<T> {
```
fabricedesre: Maybe rename to HueLightService in case they add other devices to their bridge.

src/events.rs
```diff
@@ -16,6 +17,7 @@ impl EventData {
     pub fn description(&self) -> String {
         let description = match *self {
             EventData::AdapterStart { ref name } => name,
+            EventData::AdapterInfo { ref info } => info,
```
fabricedesre: I would call that AdapterNotification

src/adapter/philips_hue/http.rs
```diff
@@ -0,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#![allow(dead_code)]
+
+extern crate hyper;
+
+use std::io::Read as stdRead; // required for hyper results' .read_to_string
```
fabricedesre: Nit: s/stdRead/StdRead

src/adapter/philips_hue/api/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+impl HueApi {
+    pub fn new(id: String, ip: String) -> HueApi {
+        // TODO: Generate a unique but reproducible access token.
+        // The token must not be guessable by an outsider.
+
+        // TODO //                                       // TODO //
+        // TODO // ___THIS IS A GAPING SECURITY HOLE!___ // TODO //
+        // TODO //                                       // TODO //
```
fabricedesre: no need to be overly dramatic ;) Just file a followup!

src/adapter/philips_hue/api/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+    // This implementation offers 64 bit
+    fn as_hash(&self) -> String {
+        let mut s = SipHasher::new();
+        self.hash(&mut s);
+        format!("{:016x}", s.finish()).to_owned()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct HueApi {
```
fabricedesre: This is the HueBridgeApi, right?

fabricedesre:
@cr I haven't looked at everything in detail. In general, can you check that things are named by what they are, and also refrain from single letter variables. The one blocker issue code-wise is the http request on the main thread.

And now, let's try that with my lights...

fabricedesre:
Ok, got it to work. Disco!

src/adapter/philips_hue/mod.rs
cr: Originally the query was inside the thread, but that resulted in and Arc weirdness and lifetime carnage in other structs used inside the hub loop. The fix was to make the ServiceAdapter static, resulting in this block. Luckily I don't need the hubs to be part of the ServiceAdapter's state any longer due to design changes towards static Hue objects in general. Trivial fix. Done. For now...

At some point though the ServiceAdapter and many other structs can't be static any longer. I have been down that road and there were lifetime roadblocks every meter. Can you suggest someone to help out with that?

src/adapter/philips_hue/mod.rs
cr: Fixed.

Was a redesign remnant. _.is_paired()_ checked implicitly.

src/adapter/philips_hue/mod.rs
cr: Refactored. Much better now.

src/adapter/mod.rs
cr: Done.

src/adapter/mod.rs
cr: Done.

Seems it was automatically fixed by my editor in one of my other commits.

src/adapter/philips_hue/api/mod.rs
cr: At this point we're fully controlling token generation and it is guaranteed to only include alphanums and a dash by the way it's constructed.

But you're raising a very good point about sanitizing data that comes from the outside. As dhylands mentioned, the token will be provided by the bridge in the future, although even then we know it will be only alphanum.

src/adapter/philips_hue/api/mod.rs
cr: Fixed. I just didn't want any of this to go unnoticed should I die in a fire before my work is done here. ;)

src/adapter/philips_hue/http.rs
cr: Fixed.

src/adapter/philips_hue/api/mod.rs
cr: Fixed.

src/events.rs
cr: Refactored.

src/adapter/philips_hue/mod.rs
cr: Done.

In fact they already have sensors. A "daylight sensor" is on every bridge. Doesn't seem to work for me. Might be connected to setting GPS coordinates and/or some external service. Also switches appear as sensors.

src/adapter/philips_hue/api/mod.rs
cr: Refactored to _HueHubApi_.

In the code I very much prefer _Hub_ over _Bridge_ for brevity.

src/adapters/philips_hue/api/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+        debug!("Philips Hue API response: {:?}", content);
+        content
+    }
+
+    pub fn is_available(&self) -> bool {
+        let url = format!("http://{}/", self.ip);
+        let content = http::get(url);
+        match content {
+            Ok(value) => {
+                value.contains("hue personal wireless lighting")
```
fabricedesre, line 95: I hope they won't change the html snippet they send to non-browser consumers! This returns (`<html><head><title>hue personal wireless lighting</title></head><body><b>Use a modern browser to view this resource.</b></body></html>`). But good enough for now

src/adapters/philips_hue/api/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    pub fn get_settings(&self) -> String {
+        // [{"error":{"type":1,"address":"/","description":"unauthorized user"}}]
+        self.get("".to_owned()).unwrap_or("".to_owned()) // TODO no unwrap
+    }
+
+    pub fn is_paired(&self) -> bool {
+        let settings = self.get_settings();
+        !settings.contains("unauthorized user")
+    }
```
fabricedesre, line 111: It seems like this should be a function of the Hub, not of the API. If you agree let's file a follow up.

cr, line 111: I only half agree. In the current scheme of things it makes sense to move it out, but over its lifetime, the _HueHubApi_ object has turned from something that manages its own state for buffering purposes into a thin static layer. It had to because cross-referencing objects turned ugly fast when multi-threading got involved, and given the deadline this week there was not enough time to make it proper.

In the not too long run I want the api and light objects to manage their state again. This will not only enable intelligent buffering again that provides considerable performance gains, it will also pave the way for elegant management of devices popping in and out of existence during runtime.

src/adapters/philips_hue/api/mod.rs
cr, line 95: It's probably the least invasive and most expressive way for such a ping. Poking around in the API leaves us just with rather generic responses to work with:

``` json
[{"error":{"type":4,"address":"/","description":"method, GET, not available for resource, /"}}]
```

or

``` json
[{"error":{"type":1,"address":"/","description":"unauthorized user"}}]
```

neither of which seem expressive enough. If I encode too much of these strings into the check, we in turn need to hope that the API doesn't change its ordering or something worse. Perhaps some time I'll come up with something that tries to be intelligent about this.

src/adapters/philips_hue/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+
+                    // Querying the settings fails when not paired with the bridge
+                    if hub.is_paired() {
+                        true
+                    } else {
+                        controller.send_event(
+                            EventData::AdapterNotification {
+                                info: format!(
+                                    "Push pairing button on Philips Hue bridge with ID {}",
+                                    hub.id)
```
fabricedesre: Let's turn that into a json snippet, as simple as: { "adapter": "philips_hue", "message": "NeedsPairing", "hub": "$hub_id" }
And the UI will manage the l10n part and copy-writing

src/adapters/philips_hue/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+                            }).unwrap();
+                        let pairing_result = hub.pair();
+                        if pairing_result.contains("success") {
+                            controller.send_event(
+                                EventData::AdapterNotification {
+                                    info: format!(
+                                         "Pairing with Philips Hue bridge {} successfull",
```
fabricedesre: Same here

src/adapters/philips_hue/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+                                         hub.id)
+                                 }).unwrap();
+                            true
+                        } else {
+                            controller.send_event(
+                                EventData::AdapterNotification {
+                                    info: format!(
+                                         "Giving up on pairing with Philips Hue bridge {}",
```
fabricedesre: And here

src/adapters/philips_hue/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+            controller.send_event(
+                EventData::ServiceStart { id: props.id.to_string() }).unwrap();
+            info!("Service {} started for Philips Hue light \"{}\" on bridge {}",
+                props.id, light.hue_id, light.hub_id);
+            let mut i = 0;
+            loop {
+                // TODO: Monitor and manage state changes.
+                i += 1;
+                thread::sleep(Duration::from_millis(60000));
+                debug!("Bip #{} from {}", i, props.id);
```
fabricedesre: No need to Bip.

src/adapters/philips_hue/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+            }
+            // controller.send_event(
+            //     EventData::ServiceStop { id: props.id.to_string() }).unwrap();
```
fabricedesre: Remove commented code.

src/adapters/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+impl<T: Controller> AdapterManager<T> {
+    pub fn new(controller: T) -> Self {
+        debug!("Creating Adapter Manager");
+        AdapterManager { controller: controller }
+    }
+
+    pub fn start(&self) {
+        // Start all the adapters.
+        philips_hue::PhilipsHueAdapter::new(self.controller.clone()).start();
+        dummy::DummyAdapter::new(self.controller.clone()).start();
```
fabricedesre: Let's not start this one anymore. It just cluttered the console output ;) But keep the adapter code itself for now as a simple example.

src/adapters/philips_hue/mod.rs
cr: Done. Also refactored for sending the event every second, else the app is likely to miss it.

src/adapters/philips_hue/mod.rs
cr: Done.

src/adapters/philips_hue/mod.rs
cr: Done.

src/adapters/philips_hue/mod.rs
cr: Removed.

src/adapters/philips_hue/mod.rs
cr: Removed.

src/adapters/mod.rs
cr: Bye, bye, Dummy!

Issue #69, opened 2016-02-23
Specify the network interface to use to extract the local ip from


arcturus:
In #58 we added the client side to connect to the registry service to allow discovery.

In the case of multiple interfaces connected to internet (weird but possible specially if we are in a rpi2), we should provide a way to hook the discovery of the ip to a selected interface specified by the user via command line parameters.

arcturus:
Closed by #83

PR #83, opened 2016-02-25
Issue #69 - Add -i/--iface to specify an interface

hfiguiere:
-Also return an error if no IP address is found
-Stop iterating if one IP address was found.
-Add em to the iface whitelist since:
[    2.340010] e1000e 0000:00:19.0 em1: renamed from eth0

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -63,7 +63,7 @@ use controller::{ Controller, FoxBox, DEFAULT_HTTP_PORT };
 use tunnel_controller:: { TunnelConfig, Tunnel };
 
 docopt!(Args derive Debug, "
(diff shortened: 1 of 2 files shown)


hfiguiere:
@arcturus @fabricedesre r?

The command-line override the white-list. Also fixed some bugs in get_ip_addr() (see commit message)

Haven't tested on rpi as I don't have hardware.

arcturus:
LGTM, tested and working fine even with interfaces that are not in the white list. r+

hfiguiere:
I wrote some more testing. Updated the PR because it is important to test.

hfiguiere:
@fabricedesre r?

PR #85, opened 2016-02-25
Remove useless thread spawning

fabricedesre:
@cr r?


cr:
I was planning on using that stub, but it'll be easy to bring it back with code that actually does something.

PR #87, opened 2016-02-26
Process the discovered hubs in parallel

fabricedesre:
The current code serializes the pairing of Hubs that are discovered by nupnp. In my case we have 2 hubs on the network, but I paired with only one of them, and don't want to pair with the second one.
In this situation I had to wait 120 seconds for the pairing to fail before moving to the actual service registration.

Instead, this patch processes each npunp-discovered hub in a short lived thread, so that unpaired ones don't block others.

@cr r?


cr:
Thanks! (Ticks this one off her TODO list.)

PR #88, opened 2016-02-26
Remove events.rs

fabricedesre:
I started by cleaning up the interaction between add_service() and the StartService event, and that lead to full removal of events.rs
We used to rely on events to bridge the Context and the Controller, but we don't need them anymore since the Controller is directly available everywhere.

feedback welcome, @jedireza & @mcav


jedireza:
At first glance, I like it! I'll spend more time reviewing soon.

fabricedesre:
Rebased an updated to carry json payload as adapter notifications.

mcav:
lgtm

Issue #86, opened 2016-02-25
registration::get_ip_addr() doesn't support IPv6


hfiguiere:
registration::get_ip_addr() doesn't support IPv6

Because it only get the IPv4 address.

Also get_if_addrs() doesn't support interfaces with two addresses....

hfiguiere:
Strike that last. get_if_addrs() does support interfaces with two addresses. It just list them separately.

hfiguiere:
I have a branch with this. And more testing.

hfiguiere:
Just ipv6 support now.

PR #89, opened 2016-02-26
Issue #86 - ipv6 support

hfiguiere:
It is pinned on my other branch. Has tests.

### DIFF
--- a/src/registration.rs
+++ b/src/registration.rs
@@ -98,6 +98,7 @@ impl Registrar {
                                want_iface: &Option<String>) -> Option<String> {
 
         let mut ip_addr: Option<String> = None;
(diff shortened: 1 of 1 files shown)


hfiguiere:
This is not ready. it is too unpredictable and given the poor support of IPv6 on overall might cause things to not work.
One easy work around is to be able to force IPv4 or IPv6 through the command line.

hfiguiere:
This time it does things better. We get IPv6 addresses and use that if no IPv4 is found. Of course specifying an interface will get the IPv4 from that address... or its IPv6.

r? @fabricedesre

fabricedesre:
thanks!

PR #90, opened 2016-02-26
Add CORS to service router. r=fabrice

ferjm:


src/service_router.rs
```diff
@@ -3,12 +3,79 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use controller::Controller;
-use iron::{Request, Response, IronResult};
-use iron::headers::{ ContentType, AccessControlAllowOrigin };
+use iron::{AfterMiddleware, headers, IronResult, Request, Response};
```
fabricedesre: nit: we use the { AfterMiddleware,... Response } style in foxbox.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde = "0.6.13"
 serde_json = "0.6.0"
 serde_macros = "0.6.14"
 staticfile = "0.1.0"
+unicase = "*"
```
fabricedesre: Any reason to not pin at unicase = "1.3.0" which is the current latest version in crates.io ?

fabricedesre:
r=me with nits addressed

PR #92, opened 2016-02-26
Switching service error response to JSON

cr:
Seems wrong to deliver plaintext errors there.

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -97,10 +97,10 @@ impl Controller for FoxBox {
         let services = self.services.lock().unwrap();
         match services.get(&id) {
             None => {
-                let mut response = Response::with(format!("No Such Service: {}", id));
+                let mut response = Response::with(json!({ error: "NoSuchService", id: id }));
                 response.status = Some(Status::BadRequest);
                 response.headers.set(AccessControlAllowOrigin::Any);
-                response.headers.set(ContentType::plaintext());
+                response.headers.set(ContentType::json());
                 Ok(response)
             }
             Some(service) => {
--- a/src/service_router.rs
+++ b/src/service_router.rs
@@ -216,7 +216,7 @@ describe! service_router {
                             &mount).unwrap();
 
             let result = response::extract_body_to_string(response);
-            assert_eq!(result, "No Such Service: unknown-id");
+            assert_eq!(result, r#"{"error":"NoSuchService","id":"unknown-id"}"#);
         }
     }
 


src/controller.rs
```diff
@@ -95,10 +95,11 @@ impl Controller for FoxBox {
         let services = self.services.lock().unwrap();
         match services.get(&id) {
             None => {
-                let mut response = Response::with(format!("No Such Service: {}", id));
+                let mut response = Response::with(format!(
+                    "{{\"result\": \"error\", \"details\": \"No Such Service: {}\"}}", id));
```
fabricedesre: Move to the new json macros, and do something like : { error: "NoSuchService", id: id }

ferjm:
FWIW we are using this response format for the user related endpoints https://github.com/fxbox/users/blob/master/doc/API.md#response-format

src/controller.rs
cr: Done.

cr:
@ferjm, that error spec is a great goal aim for. I'd prefer to be switching to your format in all the places at once in a separate PR, coordinating with gmarty who is already consuming API errors.

fabricedesre:
The fxbox/users repo uses numeric error codes. That's fine when you just mirror the http ones, but please don't add new custom ones. Instead, use strings like most DOM apis do, that's much more understandable.

cr:
So are we going to land this?

fabricedesre:
This needs to be rebased

cr:
Looks like a few _json!_ switches were already landed in other patches.

src/service_router.rs
```diff
@@ -147,7 +147,7 @@ describe! service_router {
                             &service_router).unwrap();
 
             let result = response::extract_body_to_string(response);
-            assert_eq!(result, "No Such Service: unknown-id");
+			assert_eq!(result, "{\"error\":\"NoSuchService\",\"id\":\"unknown-id\"}");
```
fabricedesre: There's a nicer string litteral syntax: r#{"error": "NoSuchService" ... }#

fabricedesre:
Also, please squash your commits

src/service_router.rs
cr: It doesn't work in this context:

```
error: found invalid character; only `#` is allowed in raw string delimitation: {
           assert_eq!(result, r#{"error":"NoSuchService","id":"unknown-id"}#);
                          ^~
```

fabricedesre: right, I got it wrong. See https://github.com/fxbox/foxbox/blob/master/src/utils.rs#L72

cr: Done. Also fixed the indent.

cr:
Ok, @fabricedesre ?

PR #96, opened 2016-02-26
Add a convenient macro to create json strings from ephemeral stucts

fabricedesre:
@cr I think you will appreciate the readability improvements! (it's also much more robust).

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+extern crate serde_json;
+
+// Macros to help with json serializing of undeclared structs.
+// json_value!({ }) retunns a serde_json::value::Value from an anonymous struct.
+// json!({ }) returns a string from an anonymous struct.
+// Combining both macros let you build arbitrary complex json objects easily.
+//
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -0,0 +1,51 @@
...
+//
+// Examples: json!({ type: "light", value: 1.2, on: true });
+//           json!({ type: "notification", value: json_value!({ r: 54, g: 12, b: 88 }) });
+
+#[macro_export]
+macro_rules! json {
+    ({ $($i:ident: $v:expr),* }) => {
+        {
+            use std::collections::BTreeMap;
+            let mut map: BTreeMap<String, serde_json::Value> = BTreeMap::new();
```
cr, line 20: This will work with "{...}", but not "[...]". Can you make this generic so it takes vectors, too?

cr:
Oh my, this is so much better. We should have Rust macros for JSON parsing, too.

src/utils.rs
```diff
@@ -0,0 +1,51 @@
...
+                map.insert(String::from(stringify!($i)), serde_json::to_value(&$v));
+            )*
+            serde_json::to_string(&map).unwrap_or("{}".to_owned())
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! json_value {
+    ({ $($i:ident: $v:expr),* }) => {
```
cr, line 42: Same in this macro. Looks like it may be sufficient to add another branch for _([...])_ and then use _Vec<...>_ inside.

PR #97, opened 2016-02-26
Turn the Hue off

fabricedesre:


fabricedesre:
@cr r?

cr:
@fabricedesre, please rebase such that the _json!_ commit is removed which already landed.

PR #99, opened 2016-02-29
Unit tests to tunnel_controller::Tunnel

JohanLorenzo:
Here are a few tests for `Tunnel`.

r? @samgiles

### DIFF
--- a/src/tunnel_controller.rs
+++ b/src/tunnel_controller.rs
@@ -10,9 +10,18 @@ use managed_process::ManagedProcess;
 
 pub type TunnelProcess = ManagedProcess;
 
-pub struct Tunnel {
-    config: TunnelConfig,
+pub struct Tunnel<T> where T: TunnelConfigTrait {
+    config: T,
     pub tunnel_process: Option<TunnelProcess>
+}
+
+pub trait TunnelConfigTrait {
+    fn new(tunnel_url: String,
+           tunnel_secret: String,
+           local_http_port: u16,
+           local_ws_port: u16,
+           remote_name: String) -> Self;
+    fn spawn(&self) -> Result<Child>;
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


src/tunnel_controller.rs
```diff
@@ -89,3 +96,52 @@ impl Tunnel {
...
+    before_each {
+        use std::thread::sleep;
+        use std::time::Duration;
+        use stubs::tunnel::TunnelConfigStub;
+
+        let config = TunnelConfigStub::new(1, "a".to_owned());
+        let mut tunnel = super::Tunnel::new(config.clone());
+
+        tunnel.start().unwrap();
+        // XXX Sleep needed otherwise the process is not started in time
```
JohanLorenzo, line 125: I'm not too sure why a wait is needed, when start waits on the process to be spawned and [so does ManagedProcess](https://github.com/fxbox/foxbox/blob/e78390d2b46447805a1c316a110075af27826970/src/managed_process.rs#L93). Do you have an idea of what I could change in the Stub, @samgiles ?

samgiles:
My apologies, I completely missed the mention, will review asap.

src/tunnel_controller.rs
samgiles, line 125: What happens if the wait is removed?

JohanLorenzo, line 125: `should_start_a_new_tunnel` and `should_stop_a_tunnel` fail. The count has a value decreased by one (compared to the expected value).

JohanLorenzo:
Unbitrot and test back to green. r? @samgiles

src/tunnel_controller.rs
samgiles, line 125: Ah, start doesn't wait for the process to be spawned in `ManagedProcess`.  `start` kicks off a thread, which will eventually start the process.

We could move these stats (https://github.com/fxbox/foxbox/blob/e78390d2b46447805a1c316a110075af27826970/src/managed_process.rs#L66) to an `Arc<RwLock<_>>` on the ManagedProcess and have an API to inspect current backoff and process starts from the controlling thread.

samgiles:
I think we should add the API to `ManagedProcess` I mentioned and use that in the tests, I foresee intermitents otherwise.

JohanLorenzo:
There hasn't been any activity on this PR for 2 years. Although, I see some fxbox repos still have some activity. I'm closing this PR for the sake of cleanup. Please reopen if you think it still has some use.

PR #100, opened 2016-02-29
Fix small unused import warning

JohanLorenzo:
Follow up #96 

r? @fabricedesre


JohanLorenzo:
Was actually done in #98

PR #101, opened 2016-02-29
Move iron-test out of regular build. r=jlorenzo

ferjm:


src/service_router.rs
```diff
@@ -104,71 +104,77 @@ pub fn create<T: Controller>(controller: T) -> Chain {
     chain
 }
 
-
-// FIXME: Declare this crate in the test module, and avoid the use of pub
-// When this comment was written, putting these 2 lines in before_each returns a compile error:
-// ``` error: unresolved import `self::iron_test::request`. Could not find `iron_test` in
-// service_router::service_router` [E0432] ```
-extern crate iron_test;
```
JohanLorenzo, line 112: Nice! Thanks for finding how to fix it. 

For reference, @ferjm gave me the trick on IRC: you have to do the import in main.rs

src/main.rs
```diff
@@ -31,6 +31,8 @@ extern crate env_logger;
 extern crate foxbox_users;
 #[macro_use]
 extern crate iron;
+#[cfg(test)]
```
JohanLorenzo, line 34: Nit: I'd move this import with the [other test dependencies](https://github.com/fxbox/foxbox/pull/101/files#diff-639fbc4ef05b315af92b4d836c31b023R22). Btw, they are probably [best placed here](https://github.com/fxbox/foxbox/pull/101/files#diff-639fbc4ef05b315af92b4d836c31b023R47).

src/service_router.rs
```diff
@@ -104,71 +104,77 @@ pub fn create<T: Controller>(controller: T) -> Chain {
...
     }
 
-    it "should get the appropriate CORS headers" {
-        use iron::headers;
-        use super::CORS;
+    describe! cors {
+        before_each {
+            use iron::Headers;
+            use controller::FoxBox;
+            use iron_test::request;
```
JohanLorenzo: Nit: Could we factorize these imports in a `before_each` in the `service_router` scope? 

Edit: Same comments with the variable initialization.

src/service_router.rs
```diff
@@ -104,71 +104,77 @@ pub fn create<T: Controller>(controller: T) -> Chain {
...
-                },
-                _ => {
-                    assert!(false)
+            let controller = FoxBox::new(false, Some("localhost".to_owned()), None, None);
+            let service_router = create(controller.clone());
+        }
+
+        it "should get the appropriate CORS headers" {
+            use iron::headers;
+            use super::super::CORS;
```
JohanLorenzo, line 157: Nit: Could we put these use in `service_router::cors::before_each`?

src/service_router.rs
```diff
@@ -104,71 +104,77 @@ pub fn create<T: Controller>(controller: T) -> Chain {
...
+                           &(path.join("/").replace("*", "foo"));
+                match request::options(&path, Headers::new(), &service_router) {
+                    Ok(res) => {
+                        let headers = &res.headers;
+                        assert!(headers.has::<headers::AccessControlAllowOrigin>());
+                        assert!(headers.has::<headers::AccessControlAllowHeaders>());
+                        assert!(headers.has::<headers::AccessControlAllowMethods>());
+                    },
+                    _ => {
+                        assert!(false)
```
JohanLorenzo: I think we can make this part smaller:

``` rust
                let response = request::options(&path, Headers::new(), &service_router).unwrap();
                let headers = &response.headers;
                assert!(headers.has::<headers::AccessControlAllowOrigin>());
                assert!(headers.has::<headers::AccessControlAllowHeaders>());
                assert!(headers.has::<headers::AccessControlAllowMethods>());
```

If the unwrap() fails, the thread will panic and the test will fail. More over, the error message will be more detailed than `panicked at 'assertion failed: false'`

JohanLorenzo:
r+

Looks good to me overall, modulo a few nits and a small improvement we could do.

PR #103, opened 2016-02-29
registration: Split test into many small ones

JohanLorenzo:
Not a lot of changes done, except the tests now use `stainless` and are broken down into 4. r? @hfiguiere

### DIFF
--- a/src/registration.rs
+++ b/src/registration.rs
@@ -17,7 +17,7 @@ use std::io::Read;
 use std::time::Duration;
 use std::thread;
 
-const REGISTRATION_INTERVAL: u32 = 1; // in minutes.
+const REGISTRATION_INTERVAL_IN_MINUTES: u32 = 1;
 
 pub struct Registrar;
 
@@ -38,7 +38,7 @@ impl Registrar {
         info!("Got ip address: {}", ip_addr.clone().unwrap());
         let full_address = format!("{}?ip={}", endpoint_url, ip_addr.unwrap());
 
-        // Spawn a thread to register every REGISTRATION_INTERVAL minutes.
+        // Spawn a thread to register every REGISTRATION_INTERVAL_IN_MINUTES.
         thread::Builder::new().name("Registrar".to_owned())
                               .spawn(move || {
             loop {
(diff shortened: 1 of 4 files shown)


JohanLorenzo:
Rebased on top of #89. IPv4 and IPv6 tests are not under 2 distinct submodules. I also used regex to validate `registrar.get_ip_addr()`. There, I didn't handle the case where we only have ipv6 addresses. I'm not sure how urgent it is to implement this particular check

r? @hfiguiere

src/registration.rs
```diff
@@ -138,115 +138,130 @@ impl Registrar {
...
+                .unwrap();
+            assert_eq!(ip_addr, "::ffff:192.168.0.4");
+        }
+
+        it "should return IPv4 if both are specified" {
+            let ip_addr = registrar.get_ip_addr_from_ifaces(&interfaces, &Some("eth0".to_owned()))
+                .unwrap();
+            assert_eq!(ip_addr, "192.168.0.4");
+        }
+    }
```
hfiguiere, line 265: what about testing `registrar.get_ip_addr_from_ifaces(&interfaces, &None)` ? This was tested before.

JohanLorenzo, line 265: I'm not too sure. Isn't that already covered by the combination of these 2 cases:
- ["should default to eth0"](https://github.com/fxbox/foxbox/pull/103/files#diff-81c55a5b1db198d473ae06791f9953ffR198)
- and ["should return IPv4 if both are specified"](https://github.com/fxbox/foxbox/pull/103/files#diff-81c55a5b1db198d473ae06791f9953ffR261)

hfiguiere, line 265: in the first case there is no IPv6 listed in the interfaces...

hfiguiere, line 265: since you removed it from the sample data.

hfiguiere:
Feel free to land this if you want. I'll address the test coverage issues afterwards.

src/registration.rs
JohanLorenzo, line 265: Done.

JohanLorenzo:
> Feel free to land this if you want. I'll address the test coverage issues afterwards.

[Test added](https://github.com/fxbox/foxbox/pull/103/files#diff-81c55a5b1db198d473ae06791f9953ffR261).

PR #105, opened 2016-02-29
Add IP Camera Adapter

dhylands:
I'm guessing you'll want o review this?

@fabricedesre r?


src/adapters/ip_camera_adapter/ipcam.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python3
```
fabricedesre: So we already have $src_dir/huedemo.sh and now this script. I don't care too much about python/bash/whatever, but they would be easier to find if we put them all in the same place. Maybe in $src_dir/tools/scripts? Or feel free to propose something else.

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+pub struct IpCameraAdapter<T> {
+    name: String,
+    controller: T
+}
+
+impl<T: Controller> IpCameraAdapter<T> {
+    pub fn new(controller: T) -> Self {
+        debug!("Creating IpCameraAdapter");
+        IpCameraAdapter { name: "IpCameraAdapter".to_owned(),
+                       controller: controller }
```
fabricedesre: nit: align the indentation of `controller` with `name`

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    fn start(&self) {
+        //let mut id = 0;
+        let controller = self.controller.clone();
+        thread::spawn(move || {
+            controller.adapter_started("IpCamera Service Adapter".to_owned());
+
+            // TODO: We should do discovery and instantiate a new IpCameraService
+            //       for each camera that we discover.
+
```
fabricedesre, line 369: what's the plan for discovery?

src/adapters/ip_camera_adapter/ipcam.py
dhylands: Yeah - I wasn't sure where the best place to put it would be. In ipcam.py's case, it's an example of a client. Putting them in tools/scripts seems reasonable to me.

I'll also put a comment blobk at the top to identify what its for.

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+    controller: T,
+    properties: ServiceProperties,
+    ip: String,
+    snapshot_dir: String,
+}
+
+impl<T: Controller> IpCameraService<T> {
+    fn new(controller: T, ip: &str) -> Self {
+        debug!("Creating IpCameraService");
+        let service_id = Uuid::new_v4().to_simple_string();
```
fabricedesre: We need stable ID to associate a piece of HW with persisted meta-data. For IP devices, the MAC address could do.

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+    fn cmd_snapshot(&self) -> IronResult<Response> {
+        let image_url = "image/jpeg.cgi";
+        let url = format!("http://{}/{}", self.ip, image_url);
+
+        let image = match get_bytes(url) {
+            Ok(image) => image,
+            Err(err) => {
+                let mut response = Response::with(format!("Error '{:?}' retrieving image from camera ip {}", err, self.ip));
+                response.status = Some(Status::NotFound);
+                response.headers.set(ContentType::plaintext());
```
fabricedesre: We should return some json formatted error instead.

src/adapters/ip_camera_adapter/mod.rs
dhylands: Yeah  - That's what I was thinking too. I just need to figure out how to get it. I was just following what the dummy adapter was doing to start with.

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+
+            break;
+        }
+        let mut writer = BufWriter::new(&image_file);
+        match writer.write_all(&image) {
+            Ok(_) => {}
+            Err(err) => {
+                let mut response = Response::with(format!("Error '{:?}' writing snapshot.jpg for camera ip {}", err, self.ip));
+                response.status = Some(Status::InsufficientStorage);
+                response.headers.set(ContentType::plaintext());
```
fabricedesre: here too

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+                return Ok(response);
+            }
+        }
+        let mut response = Response::with(format!("Took a snapshot: {}", full_filename));
+        response.status = Some(Status::Ok);
+        response.headers.set(ContentType::plaintext());
```
fabricedesre: and here too

src/adapters/ip_camera_adapter/mod.rs
dhylands, line 369: @aosmond has been working in upnp, which the DLink cameras supposedly support. So I'm going to look into that. I should be able to get a bunch of information from each camera (name, IP address, model, serial number).

fabricedesre:
- We really need discovery before we land that.
- Being able to set the user/passwd would be good too. I guess these cameras don't expose https endpoints?

src/adapters/ip_camera_adapter/mod.rs
fabricedesre, line 369: Cool. If you get the serial number, use that as an id instead of the mac address

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+use service::{ Service, ServiceAdapter, ServiceProperties };
+use std::fs;
+use std::io::{ BufWriter, Error, ErrorKind };
+use std::io::prelude::*;
+use std::path::Path;
+use std::sync::Arc;
+use upnp::{ UpnpListener, UpnpService };
+
+use self::hyper::header::{ Authorization, Basic, Connection };
+use self::url::{Host, Url};
```
fabricedesre: nit: { Host, Url }

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+    controller: T,
+    properties: ServiceProperties,
+    ip: String,
+    snapshot_dir: String,
+    name : String,
+    model: String,
+}
+
+impl<T: Controller> IpCameraService<T> {
+    fn new(controller: T, id: String, ip: String, name: String, model: String) -> Self {
```
fabricedesre: can all these strings be &str instead?

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+                ws_url: controller.get_ws_root_for_service(id.clone())
+            },
+            ip: ip,
+            snapshot_dir: format!("{}/{}", SNAPSHOT_DIR, id),
+            name: name,
+            model: model,
+        }
+    }
+
+    fn cmd_get(&self, filename: String) -> IronResult<Response> {
```
fabricedesre: same for filename

fabricedesre:
r=me with minor nits.

I guess we need a bunch of follow ups:
- the FIXME noted in this patch.
- ipv6 support?
- we support discovery, but not cameras "disappearing".
- storage quota.

dhylands:
@fabricedesre  Once we persist the cameras in the database I think that they would only get removed if the user removes them through the UI. Otherwise the camera could be offline temporarily due to being unplugged, or some other reason. My cameras periodically don't show up in the Dlink app just due to the vagaries of wifi.

The specifications for the DCS-5025, for example, list IPV4, but make no mention of IPV6.

To test ipv6, I guess I'd need some IPv6 equipment to setup an ipv6 network and see what I can get working. I can't imagine any normal home user knowing what IPv6 is, let alone using it.

I'll file issues for the things you mentioned.

fabricedesre:
\o/

PR #107, opened 2016-03-01
Adding persistent namespace/key/value storage

cr:


cr:
No clue why Travis tests are failing, but I can't see how it's connected to my code or tests.

Did someone land busted tests?

cr:
Worked around selenium timeouts by removing the Drop trait from the config object. Apparently the timeouts were caused by a >24 seconds foxbox init delay when run by Travis. I don't understand the connection, yet, because theoretically the config object – as part of the _FoxBox_ controller – should never go out of scope.

I am wondering if there is controller construction/destruction mayhem going on in other code paths that shouldn't. There is for example an unused controller instantiation for _FoxBoxEventLoop_. Could our selenium tests be triggering controller instantiation/destructions that I can't see during regular init? @ferjm @JohanLorenzo @fabricedesre

cr:
Looks like a few switches to *json!* were already landed in previous patches.

src/adapters/philips_hue/api/mod.rs
```diff
@@ -182,6 +182,12 @@ impl HueLight {
         api.get_light_status(&self.hue_id)
     }
 
+    pub fn get_unique_id(&self) -> String {
```
fabricedesre: This should be in its own PR, as this is not storage related.

src/adapters/philips_hue/api/mod.rs
```diff
@@ -182,6 +182,12 @@ impl HueLight {
...
+        let api = HueHubApi::new(self.hub_id.clone(), self.hub_ip.clone());
+        let ls = api.get_light_status(&self.hue_id);
+        ls.uniqueid
```
fabricedesre: Maybe just api.get_light_status(&self.hue_id).uniqueid ?

src/adapters/philips_hue/mod.rs
```diff
@@ -123,9 +126,20 @@ struct HueLightService<T> {
 }
 
 impl<T: Controller> HueLightService<T> {
-    fn new(controller: T, id: u32, light: HueLight) -> Self {
+    fn new(mut controller: T, id: u32, light: HueLight) -> Self {
         debug!("Creating HueLightService {} for HueLight {:?}", id, light);
-        let service_id = Uuid::new_v4().to_simple_string();
+
+        // Get stored service ID if available, default to random UUID
```
fabricedesre: why do you need to store the service ID since you get it from light.get_unique_id() ?

src/config_store.rs
```diff
@@ -0,0 +1,197 @@
...
+use std::collections::BTreeMap;
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::Path;
+
+
+#[derive(Debug)]
+pub struct ConfigStore {
+    file_name: String,
+    config: ConfigNameSpaces
```
fabricedesre: why not just config: BTreeMap<String, ConfigNameSpace> ?

fabricedesre:
There are a few issues with that:
- The store is locked globally. That's not acceptable since that mean that any namespace user can lock all others. Locking needs to happen at the namespace level.
- That needs to support more than gettting & setting strings. a serde value should just work.
- Storage is not robust enough. What happens if we fail to write? We should get at least something similar to writeAtomic in https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/OSFile.jsm/OS.File_for_the_main_thread. Or we need to switch to a sqlite db.

src/adapters/philips_hue/api/mod.rs
cr: It is storage-related because it is required to generate the config key.

fabricedesre: It doesn't _depend_ on storage. From what I understoo, @npark-mozilla would benefit to have a stable id for his tests. No storage needed for that right?

src/adapters/philips_hue/mod.rs
cr: _.get_unique_id()_ does not produce a service ID, so I only use it for generating a config key.

src/config_store.rs
cr: I have experimented with that before, but it messed with de/serialization and made return types awkward. Will try again.

src/adapters/philips_hue/mod.rs
fabricedesre: But why do you need to do that? Why not just hash config_key to get a service id? That seems overkill to need storage for that.

cr:
- Are _Arc_ and _Mutex_ handled transparently by _serde_json_? If not, locking at namespace level will break JSON serialization. This would have to be captured by an in-memory transform on _.save()_ which seems cost-prohibitive as long as we save after every config write. Is global locking even such a risk at this point? The mutex is handled by the config setter and getter.
- All current config needs are covered by simple strings. If we switch to _json_value_, all the strings must be converted to _json_value_s before write. If we remain with _String_ for now, potential future uses of complex data types can still use _String_ by serialization to begin with. Apart from that, most complex use requirements can easily be mapped to formatting the key.
- To make writes more reliable, I'd propose to backup the original file, then write the new file, if that fails, delete the backup, else restore.

I can easily spend the whole week engineering this _data_store_, but where to stop for the prototype?

src/adapters/philips_hue/mod.rs
fabricedesre: The uuid crate supports v5 uuid (see https://doc.rust-lang.org/uuid/uuid/struct.Uuid.html#method.new) so that looks like a good way of creating unique id here no?

src/config_store.rs
cr: Oh, it works! _serde_json::from__\* is black magic.

src/adapters/philips_hue/api/mod.rs
cr: I removed any changes to the adapter from this pull request.

src/adapters/philips_hue/api/mod.rs
cr: Removed that function.

src/adapters/philips_hue/mod.rs
cr: Removed this. Service ID is random again for now.

cr:
@fabricedesre, on to the next iteration:
- I have switched from _Mutex_ to _RwLock_, so only writes (rare) are locking now.
- The inner data type is still _String_. I'll switch to _json_value_ in an instant if someone comes up with a strong use case. So far all planned uses are covered by _String_.
- Saving is now protected by a _Mutex_, and it keeps the old file around until the new file has been written successfully.

src/config_store.rs
```diff
@@ -0,0 +1,169 @@
...
+impl ConfigStore {
+    pub fn new(file_name: String) -> Self {
+        ConfigStore {
+            file_name: file_name.clone(),
+            save_lock: Mutex::new(()),
+            config: ConfigStore::load(file_name)
+        }
+    }
+
+    pub fn set(&mut self, namespace: String, property: String, value: String) {
```
fabricedesre: Could all these parameters be &str instead? that may help with not cloning at call sites

src/config_store.rs
```diff
@@ -0,0 +1,169 @@
...
+        debug!("Setting config for {}::{} to {}", namespace, property, value);
+        if !self.config.contains_key(&namespace) {
+            self.config.insert(namespace.clone(), ConfigNameSpace::new());
+        }
+        self.config.get_mut(&namespace).unwrap().insert(property, value);
+        // TODO: Should be more intelligent than save on every write
+        self.save();
+    }
+
+    pub fn get(&self, namespace: String, property: String) -> Option<&String> {
```
fabricedesre: here too

src/config_store.rs
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            debug!("No config result for {}::{}", namespace, property);
+            None
+        }
+    }
+
+    fn load(file_name: String) -> ConfigTree {
+        let empty_config = BTreeMap::new();
+        let path = Path::new(&file_name);
+        let file = match File::open(&path) {
```
fabricedesre: Inline Path:new(...)

src/config_store.rs
```diff
@@ -0,0 +1,169 @@
...
+                    file_name, error.to_string());
+                return empty_config;
+            }
+        };
+        let parsed_config: ConfigTree = match serde_json::from_reader(&file) {
+            Ok(value) => value,
+            Err(error) => {
+                error!("Unable to generate JSON from config file {}: {}",
+                    file_name, error.to_string());
+                return empty_config;
```
fabricedesre: No need to return here

src/config_store.rs
```diff
@@ -0,0 +1,169 @@
...
+            }
+        };
+
+        debug!("Parsed config file: {:?}", parsed_config);
+        parsed_config
+    }
+
+    #[allow(unused_variables)]
```
fabricedesre: Is that for lock = ... ? If so, replace by let _ = self.save_lock.lock().unwrap()

src/controller.rs
```diff
@@ -49,6 +51,9 @@ pub trait Controller : Send + Sync + Clone + Reflect + 'static {
     fn services_as_json(&self) -> Result<String, serde_json::error::Error>;
     fn get_http_root_for_service(&self, service_id: String) -> String;
     fn get_ws_root_for_service(&self, service_id: String) -> String;
+    fn get_config(&self, namespace: String, property: String, default: String)
+        -> String;
```
fabricedesre: nit: feel free to keep that at the end of the previous line. Looks more readable to me.

src/controller.rs
```diff
@@ -153,6 +159,18 @@ impl Controller for FoxBox {
         format!("ws://{}:{}/services/{}/", self.hostname, self.ws_port, service_id)
     }
 
+    fn get_config(&self, namespace: String, property: String, default: String)
+        -> String {
```
fabricedesre: I'd prefer keeping the return type be Option<String>. That's more Rust-y and let callers fallback as they like.

src/controller.rs
```diff
@@ -153,6 +159,18 @@ impl Controller for FoxBox {
...
+        match self.config.read().unwrap().get(namespace, property) {
```
fabricedesre: what if read() failed?

src/controller.rs
```diff
@@ -153,6 +159,18 @@ impl Controller for FoxBox {
...
+            Some(v) => v.to_owned(),
+            None => default
+        }
+    }
+
+    fn set_config(&mut self, namespace: String, property: String, value: String) {
+        self.config.write().unwrap().set(namespace, property, value);
```
fabricedesre, line 173: here too, what if write() fails?

fabricedesre:
One more thing: could you move the Rwlock on the `config: ConfigTree` field? This way any user would be safe.

src/config_store.rs
cr: Yes! Much better.

src/config_store.rs
cr: Changed.

src/config_store.rs
cr: Done.

src/config_store.rs
cr: Replaced. Didn't remember that one can use that idiom.

src/config_store.rs
cr: Changed. It leads to a somewhat redundant _debug!_ message, but perhaps it even makes sense.

src/controller.rs
cr: Done. Especially after the switch to _&str_.

src/controller.rs
cr: I disagree. This feels inconsistent with the idea of the _default_ parameter and it leads to very repetitive and ugly code in the callees. Been there.

The consistent alternative would be to rename this to _.get_config_with_default()_, and have _.get_config()_ behave as you suggest.

src/controller.rs
cr: It only fails on poisoned locks which only occur when a writer panics. If that happens, there is not much sense in trying to recover here, is it?

src/controller.rs
cr, line 173: Same argument as above. What should we do if it fails?

src/controller.rs
fabricedesre: So let's go with .get_config_with_default() and .get_config().  But get_config_with_default() is basically .get_config().unwrap_or(default) which is not so ugly in my view.

cr:
Having the _RWLock_ on the _ConfigTree_ makes the black _serde_json_ magic fail:

```
src/config_store.rs:88:28: 88:56 error: the trait `serde::ser::Serialize` is not implemented for the type `std::sync::rwlock::RwLockReadGuard<'_, controller::collections::btree::map::BTreeMap<controller::collections::string::String, controller::collections::btree::map::BTreeMap<controller::collections::string::String, controller::collections::string::String>>>` [E0277]
src/config_store.rs:88         let conf_as_json = serde_json::to_string_pretty(&self.config.read().unwrap()).unwrap();
                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

cr:
I reverted from trying to move the lock inside config_store because it upsets serde_json, complaining about _RwLock_ lacking _Serialize_. Trying to derefence the lock leads down a slippery lifetime error road. Chasing this much longer may not be worth it.

src/controller.rs
cr: I just realized that the committed version got dumbed down from the original intention to also **set** the default value if it's not present. That allows efficiently working with random init values.

I went for option 2.

PR #109, opened 2016-03-01
two simple tests added

isabelrios:


.travis.yml
```diff
@@ -32,4 +32,4 @@ script:
     - cargo build # Linter is also executed here
     - RUST_BACKTRACE=1 cargo test
     - jshint static/js/*.js test/selenium/*.js
-    - (cargo run -- -n $BOX_HOST_NAME -p $BOX_PORT &> /dev/null &) ; npm test
+    - (cargo run -- -n $BOX_HOST_NAME -p $BOX_PORT
```
JohanLorenzo: My apologies, the line should be: `(cargo run -- -n $BOX_HOST_NAME -p $BOX_PORT &) ; npm test`.
- `&> /dev/null` means redirect the standard and the error output to nothing.
- `&` before `)` means we start a new process
- `;` means we continue to execute the next command, whatever the result of the first command was
- `npm test` runs the Selenium tests

isabelrios:
@JohanLorenzo, removing the r? since I realized I have not deleted many functions and variables that are not used in this test. Will fix that and ask for review then.

isabelrios:
I think now it is a bit better, thanks. @JohanLorenzo r?

test/selenium/sessions_ui_test.js
```diff
@@ -129,6 +132,21 @@ describe('sessions ui', function() {
         });
       });
 
+      it('should fail if password is not set', function() {
```
JohanLorenzo, line 134: Nit: Wrong test description (bad copy and paste [from here](https://github.com/fxbox/foxbox/pull/109/files#diff-13d61899e671b95e82cb9b44c6b8ee08R268))

test/selenium/signInPage.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
```
JohanLorenzo: Nits:
- This file should under `test/selenium/lib`. 
- The file name should lower case only (otherwise platforms like Windows or (some) Mac OS will have trouble if another is called `signinpage.js`)

test/selenium/sessions_ui_test.js
```diff
@@ -255,6 +288,34 @@ describe('sessions ui', function() {
...
+          signInPage = new SignInPage(driver);
+
+        });
+        beforeEach(function(){
+          return driver.wait(webdriver.until.titleIs('FoxBox'), 5000).then(
+            function() {
+          return signInPage.openPage();
+          }).then(function(signedInPageView) {
+             signedInPage = signedInPageView;
+             return signedInPage;
```
JohanLorenzo: Nit: Return not needed (there is no `.then()` that may use this value)

test/selenium/sessions_ui_test.js
```diff
@@ -255,6 +288,34 @@ describe('sessions ui', function() {
...
 
+    describe('tests changing views', function(){
+
+        before(function() {
+          driver.navigate().refresh();
+          signInPage = new SignInPage(driver);
+
+        });
+        beforeEach(function(){
+          return driver.wait(webdriver.until.titleIs('FoxBox'), 5000).then(
```
JohanLorenzo, line 299: _No need to change this line in the current PR_

This wait is done at some other places, we should think of extracting it in a page class. This is better done is a follow up PR

test/selenium/sessions_ui_test.js
```diff
@@ -255,6 +288,34 @@ describe('sessions ui', function() {
...
+            function() {
+          return signInPage.openPage();
+          }).then(function(signedInPageView) {
+             signedInPage = signedInPageView;
+             return signedInPage;
+          });
+        });
+
+        it('should go to sign out page' , function() {
+          signedInPage.tapSignOutButton().then(function(signedOutPageView){
```
JohanLorenzo: Nits:
- `return` is missing at the beginning of the line
- the first then can be combined with the second one:

``` js
signedInPage.tapSignOutButton().then(function(signedOutPageView){
   signedOutPage = signedOutPageView;
   return signedOutPage.confirmPage();
})
```

test/selenium/sessions_ui_test.js
```diff
@@ -255,6 +288,34 @@ describe('sessions ui', function() {
...
+              signedOutPage = signedOutPageView;
+              return signedOutPage;
+          }).then(function() {
+              return signedOutPage.confirmPage();
+          }).then(function(present){
+              return assert.equal(present, true);
```
JohanLorenzo: Nit: `assert.ok(present)` (without return)

test/selenium/lib/view/signed_in/view.js
```diff
@@ -0,0 +1,22 @@
...
+
+function SignedInPageView(driver) {
+    this.driver = driver;
+    this.accessors = new SignedInPageAccessor(this.driver);
+    signedOutPageView = new SignedOutPageView(this.driver);
+};
+
+SignedInPageView.prototype = {
+    tapSignOutButton: function() {
+        return this.accessors.getSignOutButton.click().then(function() {
```
JohanLorenzo: Nice! :+1:

test/selenium/lib/view/signed_in/view.js
```diff
@@ -0,0 +1,22 @@
...
+var signedOutPageView;
+
+function SignedInPageView(driver) {
+    this.driver = driver;
+    this.accessors = new SignedInPageAccessor(this.driver);
+    signedOutPageView = new SignedOutPageView(this.driver);
+};
+
+SignedInPageView.prototype = {
+    tapSignOutButton: function() {
```
JohanLorenzo: Nit: Could be renamed by something that translates the user's will. Something like `signOut`

test/selenium/sessions_ui_test.js
isabelrios: I had the return in the previous line (304) but travis was complaining: Did you mean to return a conditional instead of an assignment?

test/selenium/lib/view/signed_in/view.js
```diff
@@ -0,0 +1,22 @@
...
+
+var webdriver = require('selenium-webdriver');
+var SignedInPageAccessor = require('./accessors.js');
+var SignedOutPageView = require('../signed_out/view.js');
+var signedOutPageView;
+
+function SignedInPageView(driver) {
+    this.driver = driver;
+    this.accessors = new SignedInPageAccessor(this.driver);
+    signedOutPageView = new SignedOutPageView(this.driver);
```
JohanLorenzo, line 13: I'd put the the initialization of the view directly in functions like `signOut`. However, if we put it here. there might be an advantage that I didn't see. Is there an advantage of putting it here?

test/selenium/lib/view/signed_out/accessors.js
```diff
@@ -0,0 +1,20 @@
...
+    page: By.id('signin')
+});
+
+function SignedOutAccessor(driver) {
+  this.driver = driver;
+};
+
+SignedOutAccessor.prototype = {
+   get isSignOutPage(){
+   	  return this.driver.isElementPresent(SELECTORS.page);
```
JohanLorenzo: :warning: We should make sure the element is also displayed. If I remember the DOM correctly, the whole page is present, and we just toggle the `hidden` attribute. Maybe we could just do something like:

``` js
get root() {
  return this.driver.wait(until.elementIsVisible(SELECTORS.page));
}
```

If we call `SignedOutAccessor().root`, we'll know the page is present and displayed

test/selenium/sessions_ui_test.js
JohanLorenzo: Hmmm, what happens if we remove line 305 and just leave line 304 (without returning anything)?

isabelrios: Not sure...will try...

test/selenium/lib/view/signed_out/view.js
```diff
@@ -0,0 +1,17 @@
...
+var webdriver = require('selenium-webdriver');
+var SignedOutPageAccessor = require('./accessors.js');
+
+function SignedOutPageView(driver) {
+    this.driver = driver;
+    this.accessors = new SignedOutPageAccessor(this.driver);
+};
+
+SignedOutPageView.prototype = {
+    confirmPage: function() {
```
JohanLorenzo: I'm not too sure how to describe this function as a user's action. What did you have in mind?

test/selenium/lib/view/signed_in/view.js
isabelrios, line 13: In this case maybe not. I would save one line of code if more functions had to return a signedOutPageView, but is not the case in this test...

test/selenium/signInPage.js
```diff
@@ -0,0 +1,17 @@
...
+
+var SignedInPageView = require('./lib/view/signed_in/view.js');
+var signedInPageView;
+
+function SignInPage(driver) {
```
JohanLorenzo: Nit: The naming is a bit redundant with `SignedInPageView`. How about something like `SetupWebapp`?

test/selenium/sessions_ui_test.js
isabelrios, line 134: Well, both tests are the same, tapping on de submit/signin without introducing the password, but can change the name to: should fail if password is not typed, for the second test

test/selenium/signInPage.js
```diff
@@ -0,0 +1,17 @@
...
+    this.driver = driver;
+    signedInPageView = new SignedInPageView(this.driver);
+}
+
+SignInPage.prototype = {
+    openPage : function() {
```
JohanLorenzo: Nit: There's no real action done by the user, so `open` is a bit misleading. What do you think of `getSignedInPage` ?

_No need to make the following change in this PR:_ On another note, I think this class has to be in charge of initializing the right view. For example:

``` js
landOnSignUp: function() {
  // clear cookies and return the SignUpView
},

landOnSignedIn: function() {
  // POST a request to get a token and return the SignInView
}

// And so on
```

What do you think?

test/selenium/lib/view/signed_in/view.js
JohanLorenzo, line 13: Ok, we can leave it there, then :)

test/selenium/sessions_ui_test.js
JohanLorenzo, line 134: Sounds good to me. Thanks for clarifying!

JohanLorenzo:
The tests look good overall. There are a few small things to modify here and there. Once they are fixed, that should be good :) 

~~r? @JohanLorenzo~~

test/selenium/lib/view/signed_out/view.js
isabelrios: I wanted to be sure I was on the correct page. If I understood your previous comment, and changing the code, removing this function and doing and assert.equal(present, true) would be enough...will try that..

test/selenium/signInPage.js
isabelrios: Initially I defined a SetUpPage for the screen where you have to insert psw and confirm, but changed to this SignIn because is the initial screen we are on in this test following the flow of the previous ones...the name would not match with the view shown if I leave it as SetUpPage...that is why I changed the name to SingedIn...

test/selenium/signInPage.js
isabelrios: Ok, that makes sense

test/selenium/signInPage.js
isabelrios: After reading your next comment, that would make sense, as it was as the beginning, so that it will be in charge of initializing each view.
Will change the name

test/selenium/lib/view/signed_out/accessors.js
isabelrios: I got an error I do not understand...If I do:
return this.driver.wait(webdriver.until.elementIsVisible(SELECTORS.page));
   }
the error is: TypeError: element.isDisplayed is not a function
BUT, if I do this: 
get root() {
    var page = this.driver.findElement(webdriver.By.id('signin'));
    return this.driver.wait(webdriver.until.elementIsVisible(page));
   }
It works and the test passes...

Should I define the selector inside the root method or look for another solution?

isabelrios: A little bit better: get root() {
    var signinPage = this.driver.findElement(SELECTORS.page);
    return this.driver.wait(webdriver.until.elementIsVisible(signinPage));
   }
Should this be valid?

isabelrios:
@JohanLorenzo, Hi Johan, I made the changes and travis is green, what do you think now? r?
btw, should I squash all the commits (there are a lot...)?

test/selenium/lib/view/signed_out/accessors.js
JohanLorenzo: That looks valid to me! :smiley:

test/selenium/sessions_ui_test.js
```diff
@@ -255,6 +287,28 @@ describe('sessions ui', function() {
...
+            function() {
+          return setUpWebapp.getSignInPage();
+          }).then(function(signedInPageView) {
+             signedInPage = signedInPageView;
+          });
+        });
+
+        it('should go to sign out page' , function() {
+          return signedInPage.signOut().then(function(isSignedOut) {
+            assert.equal(isSignedOut, true);
```
JohanLorenzo: Nit: `isSignedOut` is actually a WebElement. Asserting on it to be true, is not totally correct. You can see it if you change this line by:

``` js
assert.strictEqual(isSignedOut, true)
```

I would recommend is smaller (and also more implicit) like:

``` js
return signedInPage.signOut(); // fails if we don't land on the signOut page
```

.travis.yml
```diff
@@ -32,4 +32,4 @@ script:
     - cargo build # Linter is also executed here
     - RUST_BACKTRACE=1 cargo test
     - jshint static/js/*.js test/selenium/*.js
-    - (cargo run -- -n $BOX_HOST_NAME -p $BOX_PORT &> /dev/null &) ; npm test
+    - (cargo run -- -n $BOX_HOST_NAME -p $BOX_PORT &) ; npm test
```
JohanLorenzo: TODO: Before squashing your commits, you can roll back this change.

test/selenium/lib/setup_webapp.js
```diff
@@ -0,0 +1,17 @@
...
+var signedInPageView;
+
+function SetUpWebapp(driver) {
+    this.driver = driver;
+}
+
+SetUpWebapp.prototype = {
+    getSignInPage : function() {
+        signedInPageView = new SignedInPageView(this.driver);
+        return signedInPageView;
```
JohanLorenzo: Nit: Can be simplified by: `return new SignedInPageView(this.driver);`

test/selenium/lib/view/signed_in/view.js
```diff
@@ -0,0 +1,26 @@
...
+};
+
+SignedInPageView.prototype = {
+    signOut: function() {
+    	return this.accessors.getSignOutButton.click().then(function() {
+            signedOut = signedOutPageView;
+            return signedOut.hasSignedOut();
```
JohanLorenzo: Nit: Can be simplified by `return signedOutPageView.hasSignedOut()`

JohanLorenzo:
**r+**

Looks good to me, modulo some nits.

Yes, please squash your commits before landing them.

isabelrios:
PR closed since the code was rebased and work finished in other branch

PR #112, opened 2016-03-02
Add initial UPnP support

aosmond:


src/adapters/dummy/mod.rs
```diff
@@ -84,6 +85,24 @@ pub struct DummyAdapter<T> {
     controller: T
 }
 
+struct DummyListener;
+
+impl DummyListener {
+    pub fn new() -> Arc<Mutex<Self>> {
+        Arc::new(Mutex::new(DummyListener))
```
aosmond: Maybe it doesn't need to be put in a mutex, similar to how controller and upnp work? (Any data would be protected inside the callback methods of the trait by their own mutex?)

PR #115, opened 2016-03-02
Bubble up defaults

JohanLorenzo:


JohanLorenzo:
With this changeset, we have the server variables centralized in main. A new unit test check their values. I also updated the existing test to check `-i`, `-r`, `-t`. 

r? @fabricedesre

fabricedesre:
Indeed [default: xxx] is nice!

PR #116, opened 2016-03-02
Adding support for stable UUIDs to core and Hue adapter

cr:


fabricedesre:
What about waiting for https://github.com/rust-lang-nursery/uuid/pull/54 ?

cr:
Good to know.

Cargo.lock
```diff
@@ -21,6 +21,7 @@ dependencies = [
  "serde 0.6.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_macros 0.6.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "sha1 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
```
fabricedesre: I just learned that sha1 implementation is not great (it does extra allocations). On the other hand https://github.com/DaGenix/rust-crypto/ looks better.

src/adapters/philips_hue/api/mod.rs
```diff
@@ -182,6 +182,12 @@ impl HueLight {
         api.get_light_status(&self.hue_id)
     }
 
+    pub fn get_unique_id(&self) -> String {
+        let api = HueHubApi::new(self.hub_id.clone(), self.hub_ip.clone());
+        let ls = api.get_light_status(&self.hue_id);
+        ls.uniqueid
```
fabricedesre: nit: you can fold that all in one expression.

src/adapters/philips_hue/mod.rs
```diff
@@ -125,7 +124,8 @@ struct HueLightService<T> {
 impl<T: Controller> HueLightService<T> {
     fn new(controller: T, id: u32, light: HueLight) -> Self {
         debug!("Creating HueLightService {} for HueLight {:?}", id, light);
-        let service_id = Uuid::new_v4().to_simple_string();
+        let seed = light.get_unique_id();
+        let service_id = StableUuid::new(seed);
```
fabricedesre: nit: inline `seed`

fabricedesre:
I think we should go with that (using rust-crypto instead of sha1) and not wait on uuid v5

Cargo.lock
cr: Switched to rust-crypto. FYI: Interface is cumbersome, docs non-existent.

src/adapters/philips_hue/mod.rs
cr: Inlined.

src/adapters/philips_hue/api/mod.rs
cr: Folded.

PR #119, opened 2016-03-02
Moving Hue demo script

cr:
The script was broken by the authentication requirement. Removing it until someone finds time to add authentication support.

**EDIT**

As per @ferjm's comment I'm just moving the script out of the way.


ferjm:
We can [disable authentication at build time](https://github.com/fxbox/foxbox#disable-authentication), so I see no need to remove this script.

cr:
Ok, I amended the commit.

PR #122, opened 2016-03-02
Switching to services/list, obsoleting services/list.json

cr:
This it to get the ball rolling for moving from _services/list.json_ to _services/list_. With this commit, both routes are supported. The former doesn't match our current naming scheme and can be removed once @gmarty is using the new route in his app.

### DIFF
--- a/docs/scripts/huedemo.sh
+++ b/docs/scripts/huedemo.sh
@@ -19,7 +19,7 @@ cmd=$1
 case $cmd in
 	status)
 		echo "Talking to FoxBox at http://$HOST/services"
-		curl -s -X GET http://localhost:3000/services/list.json \
+		curl -s -X GET http://localhost:3000/services/list \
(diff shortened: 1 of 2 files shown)


fabricedesre:
what's the rationale for keeping both urls?

cr:
The rationale is to allow @gmarty a smooth transition. Once he confirms, _list.json_ is gone

fabricedesre:
Just switch, and also send a PR to gmarty if you really fear to break him.
And this is not the last time this endpoint will change, it will be different once we plug in the taxonomy.

gmarty:
Feel free to merge it and I'll update accordingly (the redirect is a nice to have, but not required).
@ferjm will also have to update his simple authentication app.

gmarty:
Is there anything blocking this PR from landing?

fabricedesre:
Doesn't look like, no.

PR #128, opened 2016-03-03
[RFC] Early prototype of a clock adapter.

Yoric:
First RFC of a clock adapter. At this stage, the adapter is not connected to the REST API. Indeed, the main idea is to test traits `Adapt` and `Adapter`.

I hope to then be able to implement trait `Adapt` into something that will expose the clock as a REST (and later WebSocket) API.

### DIFF
--- a/src/adapt.rs
+++ b/src/adapt.rs
@@ -1,0 +1,724 @@
+//! An API for plugging in adapters.
+
+#![allow(dead_code)] // Implementation in progress, code isn't called yet.
+
+use transact::InsertInMap;
+
+use foxbox_taxonomy::selector::*;
+use foxbox_taxonomy::services::*;
+use foxbox_taxonomy::api::{ API, WatchOptions, Error as APIError, WatchEvent };
+use foxbox_taxonomy::util::*;
+use foxbox_taxonomy::values::*;
+
+use std::boxed::FnBox;
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+use std::rc::Rc;
+use std::sync::mpsc::{ channel, Sender };
+use std::thread;
+
+/// An error that took place while communicating with either an adapter or the mechanism that
+/// handles registeration of adapters.
+#[derive(Debug)]
+pub enum Error {
+    DuplicateGetter(Id<Getter>),
+    NoSuchGetter(Id<Getter>),
+    GetterDoesNotSupportPolling(Id<Getter>),
+    GetterDoesNotSupportWatching(Id<Getter>),
+    GetterRequiresThresholdForWatching(Id<Getter>),
+
+    DuplicateSetter(Id<Setter>),
+    NoSuchSetter(Id<Setter>),
+
+    DuplicateService(Id<ServiceId>),
+    NoSuchService(Id<ServiceId>),
+    TypeError(TypeError),
+
+    DuplicateAdapter(Id<AdapterId>),
+    NoSuchAdapter(Id<AdapterId>),
+    InvalidValue
+}
+
+/// A witness that we are currently watching for a value.
+/// Watching stops when the guard is dropped.
+pub trait WatchGuard {
+}
+
+/// API that adapters must implement
+pub trait Adapter: Send {
+    /// An id unique to this adapter. This id must persist between
+    /// reboots/reconnections.
+    fn id(&self) -> Id<AdapterId>;
+
+    /// The name of the adapter.
+    fn name(&self) -> &str;
+    fn vendor(&self) -> &str;
+    fn version(&self) -> &[u32;4];
+    // ... more metadata
+
+    /// Request a value from a channel. The FoxBox (not the adapter)
+    /// is in charge of keeping track of the age of values.
+    fn get_values(&self, set: Vec<Id<Getter>>) -> Vec<Result<Option<Value>, Error>>;
+
+    /// Request that a value be sent to a channel.
+    fn set_values(&self, values: Vec<(Id<Setter>, Value)>) -> Result<(), Error>;
+
+    fn register_watch(&self, id: Id<Getter>, threshold: Option<Value>, cb: Box<Fn(Value) + Send>) -> Result<Box<WatchGuard>, Error>;
+}
+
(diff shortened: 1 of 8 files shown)


Yoric:
Advancing towards the integration of Taxonomy in the FoxBox.

Yoric:
Reorganized the patch, added plenty of doc and a few tests.

src/adapt.rs
```diff
@@ -0,0 +1,481 @@
+//! An API for plugging in adapters.
```
fabricedesre, line 1: nit: rename this file to adapters.rs ?

src/adapt.rs
```diff
@@ -0,0 +1,481 @@
...
+
+#![allow(dead_code)] // Implementation in progress, code isn't called yet.
+
+use transact::{InsertInMap, InsertInSet};
```
fabricedesre: We use the transact:: { InsertInMap, InsertInSet } style everywhere else in foxbox. I don't know if the "no space" version is the commonly used style in the rust community, but I find it less readable. In any case, we need to choose one style and stick to it.

src/adapt.rs
```diff
@@ -0,0 +1,481 @@
...
+    fn vendor(&self) -> &str;
+    fn version(&self) -> &[u32;4];
+    // ... more metadata
+
+    /// Request a value from a channel. The FoxBox (not the adapter)
+    /// is in charge of keeping track of the age of values.
+    fn get_values(&self, set: Vec<Id<Getter>>) -> Vec<Result<Option<Value>, Error>>;
+
+    /// Request that a value be sent to a channel.
+    fn put_values(&self, values: Vec<(Id<Setter>, Value)>) -> Result<(), Error>;
```
fabricedesre: s/put_values/set_values to be consistent with Setter ?

src/adapt.rs
```diff
@@ -0,0 +1,481 @@
...
+    fn register_watch(&self, id: Id<Getter>, threshold: Option<Value>, cb: Box<Fn(Value) + Send>) -> Result<Box<WatchGuard>, Error>;
+}
+
+struct AdapterData {
+    /// The implementation of the adapter.
+    adapter: Box<Adapter>,
+
+    /// The services for this adapter.
+    services: HashSet<Id<ServiceId>>,
+}
```
fabricedesre, line 77: nit: here and in a few other places: add blank line

src/adapt.rs
```diff
@@ -0,0 +1,481 @@
...
+                adapters: HashMap::new(),
+                services: HashMap::new(),
+                getters: HashMap::new(),
+                setters: HashMap::new(),
+            }))
+        }
+    }
+
+    /// Auxiliary function to remove an adapter, once the mutex has been acquired.
+    /// Clients should rather use Adapt::remove_adapter.
```
fabricedesre: nit: here and in other comments: s/Adapt::_/AdapterControl::_

src/adapt.rs
```diff
@@ -0,0 +1,481 @@
...
+    fn aux_remove_adapter(id: Id<AdapterId>, state: &mut AdapterControlState) -> Result<(), Error> {
+        let adapters = &mut state.adapters;
+        let services = &mut state.services;
+        let getters = &mut state.getters;
+        let setters = &mut state.setters;
```
fabricedesre: It's a bit annoying that we have a flat structure with no direct composition of adapter->1..N service->(1..N getter, 1..N setter). Would that not simplify a lot these functions?

src/adapters/clock/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+use foxbox_taxonomy::services::*;
+use foxbox_taxonomy::util::Id;
+
+use std::collections::{HashMap, HashSet};
+
+use chrono;
+use chrono::*;
+use timer;
+
+static ADAPTER_ID: &'static str = "clock@link.mozilla.org";
```
fabricedesre: xpcom, the come back of the contract id! :)

src/adapters/clock/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+static ADAPTER_NAME: &'static str = "Clock adapter (built-in)";
+static ADAPTER_VENDOR: &'static str = "team@link.mozilla.org";
+static ADAPTER_VERSION: [u32;4] = [0, 0, 0, 0];
+
+const SERVICE_CLOCK_ID: &'static str = "service:clock@link.mozilla.org";
+
+const GETTER_TIMESTAMP_ID: &'static str = "getter:timestamp.clock@link.mozilla.org";
+const GETTER_TIME_OF_DAY_ID: &'static str = "getter:timeofday.clock@link.mozilla.org";
```
fabricedesre: Do these have to be global constants? or only unique for their service?

src/main.rs
```diff
@@ -53,7 +56,8 @@ extern crate regex;
 // Need to be declared first so to let the macros be visible from other modules.
 #[macro_use]
 mod utils;
-
+mod transact;
```
fabricedesre, line 61: please move it in alphabetical order

fabricedesre:
The `Clock` example is not ideal since it merges the adapter with a singleton service. I would have preferred to see something closer to the 1 adpater - N services model that we'll have to implement for the Hue and the IP camera cases.
But I don't want to wait for ever to land that either, so it's fine to for me to land and to iterate.

src/adapt.rs
Yoric, line 1: Well, there is already a module `adapters`. I could move this in `adapters/mod.rs`, but that would make that file too large for my tastes. I could also move this to, say, `adapters/api.rs`, if you prefer.

src/adapt.rs
Yoric: Well, `services` already contains getters and setters. Here, these maps are designed to go the other way round, from getter/setter to service, from service to adapter. I'll see if I can improve that.

Yoric:
Applied feedback and updated to the latest version of the Taxonomy API. Once tests are green, I'll merge.

src/adapt.rs
fabricedesre, line 1: I prefer adapters/api.rs

PR #132, opened 2016-03-03
Add iron logger and allow enabling it via 'CARGO_LOG=http' env var. r=fabrice

ferjm:


ferjm:
It seems that we can already see http requests with `RUST_LOG=debug cargo run`

PR #134, opened 2016-03-04
Turn on code coverage on Travis

JohanLorenzo:


JohanLorenzo:
We're still missing the ability to pass cargo some flags (including the one which leaves dead code). I workaround it with [this ugly eval](https://github.com/fxbox/foxbox/pull/134/files#diff-8f914ed363b13f08138ec5d97bbf0d8cR31). I believe it won't stay too long. 

I made a bash script to test the whole process locally. I can understand bash adds another extra language in the repo. I don't mind changing the script to nodejs for instance. 

Also, the report states we have about 40% of coverage. It's seems globally correct, but not 100% perfect. At least, it gives us an idea of which areas need more coverage.

Finally, I believe we need to activate coveralls for fxbox/foxbox. The ownership transfer didn't seem to propagate to coveralls.io. See https://travis-ci.org/fxbox/foxbox/builds/113623784#L1527

r? @fabricedesre

tools/execute-unit-tests-with-coverage
```diff
@@ -0,0 +1,52 @@
...
+
+CURRENT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_HOME="$CURRENT_PATH/.."
+KCOV_TEMP="$PROJECT_HOME/target/kcov"
+KCOV_HOME="$KCOV_TEMP/kcov-$KCOV_VERSION"
+KCOV_BINARY="$KCOV_HOME/src/kcov"
+
+
+get_and_compile_kcov_locally() {
+  mkdir -p "$KCOV_TEMP"
```
fabricedesre, line 17: Since we always build the same kcov version, maybe we should host a pre-built version to cut testing times? It's fine to do that in a followup.

fabricedesre:
No problem with bash since this is linux only. I added fxbox/foxbox to coveralls.io and retriggered - let's see if that works!

fabricedesre:
Ok, we get a report of 42.95% coverage at https://coveralls.io/github/fxbox/foxbox

PR #136, opened 2016-03-04
new folders structure and new test

isabelrios:
@JohanLorenzo, I guess it is necessary to change travis to tell the new folder structure for the tests. Also, do not know if something else is necessary I got this error when trying to run selenium tests locally: 
/Users/irios/git/foxbox/node_modules/mocha/lib/utils.js:628
        throw new Error("cannot resolve path (or pattern) '" + path + "'");
Error: cannot resolve path (or pattern) 'test/selenium/*_test.js'

Although I have removed node_modules folders and npm install...


JohanLorenzo:
Crap, the nasty line is located there https://github.com/fxbox/foxbox/blob/master/package.json#L9 :smiley:

PR #138, opened 2016-03-04
Implementing config override, support for nUPnP server

cr:
The nUPnp server for Philips Hue can be overridden by running:

$ cargo run -- -c "philips_hue;nupnp_url;http://localhost:2342/"

You can specify multiple configuration overrides on the command line.


fabricedesre:
Correct me if I'm wrong but it seems that this patch:
- uses the "overrides" namespace to store overrides - but nothing prevents someone else to also use it.
- builds the overrides key as $ns/$prop, but nothing prevents $ns and $prop to also contain `/` so you may end up with collisions.

If that's the case, this needs to be implemented with a separate ConfigNamespace that will be used explicitly to store the overrides.

cr:
@fabricedesre, overrides now use a separate _ConfigTree_, resolving the collision issue.

PR #140, opened 2016-03-04
Move all the config* functions from the controller to a service

fabricedesre:


fabricedesre:
This abstracts the access to the config store out of the controller as a service. For now the gain is small, but with the addition of overrides this will make even more sense to not bloat the Controller trait with lots of config methods. 

That's also easier to test (see increase in coverage at https://coveralls.io/builds/5311774).

@cr r?

cr:
Good move!

PR #141, opened 2016-03-05
Add timestamps and tid to logging output.

dhylands:
This also sets the default logging level to INFO.

You can still override this by running foxbox using:

```
RUST_LOG=foxbox=debug cargo run
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,6 +16,7 @@ dependencies = [
  "mio 0.5.0 (git+https://github.com/carllerche/mio.git)",
  "mount 0.0.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "multicast_dns 0.1.0 (git+https://github.com/fxbox/multicast-dns.git?rev=a6e4bcc)",
+ "nix 0.5.1-pre (git+https://github.com/nix-rust/nix.git?rev=138080)",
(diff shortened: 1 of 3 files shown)


fabricedesre:
Hm, it's sad that nix doesn't provide gettid() since we already depend on it. Maybe you can send them a PR (https://github.com/nix-rust/nix) instead of adding this custom sub-crate?

dhylands:
> Hm, it's sad that nix doesn't provide gettid() since we already depend on it. Maybe you can send them a PR (https://github.com/nix-rust/nix) instead of adding this custom sub-crate?

I created https://github.com/nix-rust/nix/pull/293 let's see if I can get that landed.

dhylands:
gettid is now in nix (currently just on github, it hasn't made it to a crate yet, but will show up in nix 0.5.1)

So I rebased to the latest master and dropped the sub-crate. Do we need foxbox to build on non-linux platforms? If yes, I can add support so that the gettid field isn't shown on non-linux platforms.

fabricedesre:
We have people building on mac, so hiding the thread id on non-linux is a must, yes.

dhylands:
Updated so that it only tries to call gettid if building for linux or android (the same platforms which nix::unistd::gettid is available).

PR #149, opened 2016-03-08
added note about build dependencies

jedireza:
r? @fabricedesre


fabricedesre:
If we add build requirements, they need to be complete.
This is missing (on debian based distros):
libssl-dev
libavahi-client-dev
libsqlite3-dev

I don't know how these packages are called in other distributions.
Ideally we could provide a shell script that checks & install dependencies.

jedireza:
Thanks for letting me know. I created a table to reduce the verbosity. Should be easy for us to fill-in as needed.

fabricedesre:
Travis is red but that's unrelated :(

PR #161, opened 2016-03-08
config_store: Add after_each{}

JohanLorenzo:
I saw [this comment](https://github.com/fxbox/foxbox/blob/master/src/config_store.rs#L147). I thought I could help out here.

r? @cr

### DIFF
--- a/src/config_store.rs
+++ b/src/config_store.rs
@@ -126,7 +126,7 @@ impl ConfigService {
 }
 
 #[cfg(test)]
-describe! config_store {
+describe! config {
 
     before_each {
         use uuid::Uuid;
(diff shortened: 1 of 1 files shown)


cr:
Did you check that after_each is actually called? Here it kept littering temp config files. If so: what was my mistake?

JohanLorenzo:
That's a bug on [stainless' side](https://github.com/reem/stainless/issues/47) :smiley: When there's no statement with a semi-colon on the `it` context, stainless actually puts `after_each` right before `it`. 

One way to workaround it is to add semi-colons at the end of the brackets, like [here for instance](https://github.com/fxbox/foxbox/pull/161/files#diff-483598a04d9d076ce6f839ca3796d92eR219).

PR #162, opened 2016-03-08
Allow redirection flow for setup and login. r=fabrice

ferjm:


build.rs
```diff
@@ -10,6 +10,41 @@ fn update_local_git_hook() {
...
+    for dest in vec!["./static/setup/shared", "./static/main/shared"] {
+        let dest = Path::new(&current).join(dest);
+        let dest_str = dest.to_str().unwrap();
+        if let Err(_) = fs::metadata(&dest_str) {
+            fs::create_dir(dest_str).unwrap();
+        }
+        cp_r(&shared, &dest);
+    }
+}
+
```
fabricedesre, line 46: I would almost prefer spawning out a shell call to `cp -R` than reimplement all that...

PR #165, opened 2016-03-08
Make Philips nUPnP URL configurable

### LINKED ISSUE #106: nupnp address should be stored in a config file for test customization
CR mentioned that  nupnp server address is currently hardcoded and there is no simple mechanism to pass down command arguments to adapters. However, the plan is to have the option to store it in a global config parameter which could relatively easily be overridden.

JohanLorenzo:
Fixes #106 

r? @cr

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -58,6 +58,7 @@ pub trait Controller : Send + Sync + Clone + Reflect + 'static {
     fn broadcast_to_websockets(&self, data: serde_json::value::Value);
 
     fn get_upnp_manager(&self) -> Arc<UpnpManager>;
+    fn get_config(&self) -> Arc<ConfigService>;
 }
 
 impl FoxBox {
@@ -200,6 +201,10 @@ impl Controller for FoxBox {
 
     fn get_upnp_manager(&self) -> Arc<UpnpManager> {
         self.upnp.clone()
+    }
+
+    fn get_config(&self) -> Arc<ConfigService> {
+        self.config.clone()
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/controller.rs
```diff
@@ -201,6 +202,10 @@ impl Controller for FoxBox {
     fn get_upnp_manager(&self) -> Arc<UpnpManager> {
         self.upnp.clone()
     }
+
+    fn get_config(&self) -> Arc<ConfigService> {
```
JohanLorenzo, line 206: This part is not that great, I concede. However, getting the upnp address from a config file is the last thing blocking the integration tests. So I'd prefer to change `get_config()` in a follow up patch. What do you think @cr ?

cr:
I'm working on a command line override. Let's please wait for that to land.

nojunpark:
@cr I thought the preference was on the config file approach in case there were multiple special parameter we need to enter- I suppose via cmdline it can be managed just as well?

nojunpark:
Aha, just saw https://github.com/fxbox/foxbox/pull/138, so it can handle multiple config values, we should have a way to link a pr to the bug, didn't know about your patch.

JohanLorenzo:
Sorry, I missed #138 was solving the issue too. I only relied on #106 to see if a patch was on its way or not :S 

What would be the ETA of #138? 

We targeted integration tests running for this week. Once configuration files are dealt with, No-Jun will have to handle the merge from https://github.com/npark-mozilla/foxlinker to https://github.com/fxbox/foxbox.

cr:
ETA is ASAP. I'm still fighting rust, but I'd say any minute now.

cr:
Closing. Will be fully covered by #138 which is now RTL.

JohanLorenzo:
Thank you very much for this fast landing @cr :+1: :smiley:

nojunpark:
This is awesome, it works well with the mocha too, thanks @cr !

PR #167, opened 2016-03-09
authorizing websocket connections

jedireza:
~~Depends on https://github.com/fxbox/users/pull/60~~

~~The `Cargo.*` file changes can be ignored. They are are artifacts of working with a local copy of https://github.com/fxbox/users/pull/60 checked out.~~

Accessing the headers of the `ws-rs` request wasn't great. We may want to move over to `rust-websocket`, which IIRC, uses the same `hyper::header` module that Iron does. That would reduce all the header shenanigans in this PR and be more elegant like what we see in the `fxbox/users` Iron middleware.

To test this you just add `?auth=<jwt>` to the websocket url.

### DIFF
--- a/src/ws_server.rs
+++ b/src/ws_server.rs
@@ -1,11 +1,15 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+extern crate url;
 
+use controller::Controller;
+use foxbox_users::auth_middleware::AuthMiddleware;
+use self::url::Url;
 use std::thread;
+use ws;
 use ws::{ Handler, Sender, Result, Message, Handshake, CloseCode, Error };
 use ws::listen;
-use controller::Controller;
 
 pub struct WsServer;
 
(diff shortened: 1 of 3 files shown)


jedireza:
:pager: r? @ferjm @arcturus

src/ws_server.rs
```diff
@@ -30,8 +34,40 @@ impl WsServer {
 
 impl<T: Controller> Handler for WsHandler<T> {
 
-    fn on_open(&mut self, _: Handshake) -> Result<()> {
+    fn on_open(&mut self, handshake: Handshake) -> Result<()> {
         info!("Hello new ws connection");
+
+        let resource = &handshake.request.resource()[..];
+        let url = Url::parse(&format!("http://fox.box{}", resource)).unwrap();
```
ferjm: Don't use `unwrap` here, please. If parsing fails close the connection and bail out.

src/ws_server.rs
```diff
@@ -30,8 +34,40 @@ impl WsServer {
...
+                        self.out.close_with_reason(
+                            ws::CloseCode::Error, "Missing authorization");
+                        return Ok(());
+                    }
+                }
+            },
+            _ => {
+                self.out.close_with_reason(
+                    ws::CloseCode::Error, "Missing authorization");
+                return Ok(());
```
ferjm: nit: this is repeated several times, maybe move it to an aux function.

ferjm:
Thank you Reza! LGTM. It could use some tests though. Feel free to file a follow up issue for the tests.

src/ws_server.rs
ferjm: Also:

> [20:26:28]  <fabrice> ferjm: Is that line hardcoding the host name? https://github.com/fxbox/foxbox/pull/167/files#diff-e1ea9eba222b6cc49837b3dd765989dfR41 
> [20:26:57]  <ferjm> fabrice no, it's just a random url
> [20:28:00]  <ferjm> fabrice he just need to build a random URL to get the query string cause handshake.request.resource only gives the path 
> [20:28:13]  <ferjm> at least that's how I understood the patch
> [20:28:14]  <fabrice>   that's... well
> [20:28:41]  <fabrice>   could use a comment because gross hack

ferjm: Also:

> [20:26:28]  <fabrice>	ferjm: Is that line hardcoding the host name? https://github.com/fxbox/foxbox/pull/167/files#diff-e1ea9eba222b6cc49837b3dd765989dfR41 
[20:26:57]  <ferjm>	fabrice no, it's just a random url
[20:28:00]  <ferjm>	fabrice he just need to build a random URL to get the query string cause handshake.request.resource only gives the path 
[20:28:13]  <ferjm>	at least that's how I understood the patch
[20:28:14]  <fabrice>	that's... well
[20:28:41]  <fabrice>	could use a comment because gross hack

src/ws_server.rs
jedireza: I'd agree if serveral was more than twice. If you mean making a function for calling the `close_with_reason` method, I don't see the benefit of an extra layer of abstraction.

src/ws_server.rs
jedireza: I added a comment line (good call out).

Unwrapping here shouldn't be an issue since the handshake was already successful and the connection is open for sending and receiving messages. Further, if we panic here only the the thread for that connection would be affected.

ferjm: I'd prefer to close the connection with a proper message to the client than just panicking.

src/ws_server.rs
ferjm: I see it repeated three times and if you address https://github.com/fxbox/foxbox/pull/167#discussion_r55713780 we will have four "close with error message + return". But not a big deal.

jedireza:
@ferjm thanks for the feedback. :pray: I ended up re-working this. Feel free to merge and I'll be sure to open a follow up issue for adding tests.

ferjm:
Thanks!

PR #170, opened 2016-03-09
refactoring tests to respect PO pattern

isabelrios:


isabelrios:
@JohanLorenzo, hi Johan, could you please have a look at how the tests are written according to the PO pattern? Thanks!

isabelrios:
Travis is red, although they pass locally, will check that

isabelrios:
@JohanLorenzo now that travis is green, asking for review, r?
Thanks!

test/selenium/sessions_ui_test.js
```diff
@@ -47,115 +54,69 @@ describe('sessions ui', function() {
...
-          pwd2: driver.findElement(webdriver.By.id('signup-pwd2')),
-          set: driver.findElement(webdriver.By.id('signup-button'))
-        };
-      });
+        beforeEach(function() {
+            setUpWebapp = new SetUpWebapp(driver);
+            setUpPage = setUpWebapp.getSetUpView();
+        });
 
       it('should show the signup screen', function() {
```
JohanLorenzo: Nit: I'd add something like "by default". 

``` js
it('should show the signup screen by default', function() {
```

test/selenium/lib/view/sign_up/view.js
```diff
@@ -0,0 +1,68 @@
...
+};
+
+SetUpView.prototype = {
+
+    isSetUpView: function() {
+        return this.accessors.root;
+    },
+
+    confirmAllFieldsArePresent: function() {
+        this.passwordField();
```
JohanLorenzo: This line and the followings have no effect, and will always pass. The distinction between views and acessors is not clear. Acessors usually return WebElements, and view manipulate them.

I'd propose something like (the following comments are not needed in the final code):

``` js
get areAllFieldsPresent() {
  var expectedElements = [this.accessors.passwordField, this.accessors.confirmPassword, this.accessors.submitButton] // We get WebElements so we can interact with.
  var promises = expectedElements.map((element) => { element.isDisplayed() }); // isDisplayed returns a promise (which is a bit a of a pain). We store every promise together, so we can wait on all of them
  return Promise.all((areDisplayed) => { // Here we wait on all of them to finish. Their results are store in the array "areDisplayed". Promise.all also returns a Promise
    return areDisplayed.reduce((previous, current) => { previous && current }) // Here we concatenate the array into one boolean 
  })
}
```

Side note: array functions `(foo) => { foo.bar }` are the equivalent of `function(foo) { return foo.bar }`. They only work on not too old browser (the one which support ES6) and NodeJS (> 4.x)

Edit: We also need to make sure the password fields hide password. I'd do it in another function, which should be similar to `areAllFieldsPresent`

test/selenium/sessions_ui_test.js
```diff
@@ -47,115 +54,69 @@ describe('sessions ui', function() {
...
-        }).then(function() {
-          return driver.switchTo().alert();
-        }).then(function(alert) {
-          return alert.getText().then(function(text) {
-            assert.equal(text, 'Passwords don\'t match! Please try again.');
-          }).then(function() {
-            alert.dismiss();
+      it('should reject non-matching passwords', function() {
+          
+          return setUpPage.typePassword('12345678').then(function(){
```
JohanLorenzo, line 89: _Proposed improvement for a follow up_

How about factorizing the failure of a login in 1 function in the View ? That would help in having function that reflect the user's intent in Views. For example:

``` js
it('should reject non-matching passwords', function() {
  return setUpPage.signUpFailure(12345678, 123456).then(alertMessage) {
    assert.equal(alertMessage, errorPasswordDoNotMatch);
  });
})
```

``` js
SetUpView.prototype.signUpFailure = function(password, confirmPassword) {
  this.accessors.passwordField.sendKeys(password).then(function() {
    this.accessors.confirmPasswordField.sendKeys(confirmPassword);
  }).then(function () {
 // and so on
}
```

test/selenium/sessions_ui_test.js
```diff
@@ -47,115 +54,69 @@ describe('sessions ui', function() {
...
-        }).then(function(elt) {
-          return driver.wait(webdriver.until.elementIsVisible(elt), 5000)
-            .then(function() {
-              return elt.getAttribute('innerHTML');
-            }).then(function(value) {
-              assert.equal(value, 'Thank you!');
-            });
-        });
+      it('should accept matching, long-enough passwords', function() {
+          return setUpPage.typePassword('12345678').then(function() {
```
JohanLorenzo, line 127: _Proposed improvement for a follow up_

Like in [this comment](https://github.com/fxbox/foxbox/pull/170/files#r55529272), we could factorize the sign up procedure in a function like `signUpSuccess(password)`

test/selenium/sessions_ui_test.js
```diff
@@ -47,115 +54,69 @@ describe('sessions ui', function() {
...
-      });
 
-      it('should have the right fields', function () {
-        var types = {
-          pwd1: 'password',
-          pwd2: 'password',
-          set: 'submit'
-        };
-        var promises = Object.keys(elements).map(function(key) {
-          return elements[key].getAttribute('type').then(function(value) {
```
JohanLorenzo, line 85: :warning: This function doesn't check if the password field hides passwords anymore. As the modification might be heavy, I'd recommend to leave this test as is for now and change it in a follow up.

JohanLorenzo:
r+ modulo [this comment](https://github.com/fxbox/foxbox/pull/170/files#r55530084). It's a big problem to deal with, I'd propose to handle it in a follow up and land the rest. If you're fine with it, remove these changes, squash the commits and land it :)

test/selenium/sessions_ui_test.js
isabelrios, line 85: Sorry, not sure I understood the final changes I have to make...
if this test: it('should have the right fields', should be kept as it is, then I do not have to change anything related to confirmAllFieldsArePresent() as the change you suggested in the comment, just need to remove that part of code that I added?

(sorry I do not know how to paste the comment directly :S)

PR #172, opened 2016-03-09
Introduce ServiceSettings::custom_properties for manufacturer, model name and device type (fixes #82).

### LINKED ISSUE #82: Expose manufacturer and model names to the services/list.json
That'd be nice to have the name of the manufacturer and the model ID of each services in list.json.
This can be used to display specific icon to help users to identify devices quicker.

cc @fabricedesre and @cr

azasypkin:


src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -29,6 +30,12 @@ static CAMERA_USERNAME: &'static str = "admin";
 static CAMERA_PASSWORD: &'static str = "password";
 static SNAPSHOT_DIR: &'static str = "snapshots";
 
+const CUSTOM_PROPERTY_MANUFACTURER: &'static str = "manufacturer";
+const CUSTOM_PROPERTY_TYPE: &'static str = "type";
+const CUSTOM_PROPERTY_MODEL: &'static str = "model";
+
+const SERVICE_TYPE: &'static str = "ipcamera";
```
azasypkin, line 37: Don't really know, what type should be for ip cameras....

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -98,12 +105,12 @@ struct IpCameraService<T> {
     properties: ServiceProperties,
     ip: String,
     snapshot_dir: String,
-    name : String,
-    model: String,
```
azasypkin, line 102: I haven't noticed that we use `model` field anywhere, so moved it to custom properties, like I did fir hue.

src/service.rs
```diff
@@ -15,7 +17,9 @@ pub struct ServiceProperties {
     pub name: String,
     pub description: String,
     pub http_url: String,
-    pub ws_url: String
+    pub ws_url: String,
+    #[serde(rename="properties")]
```
azasypkin, line 21: per request from gmarty, but it doesn't really matter since we change format soon anyway, so no preference for me.

azasypkin:
Hey @fabricedesre @cr @dhylands does changes in your code look OK? I need them for the client side.

fabricedesre:
fine for me.

cr:
Looks good to me.

azasypkin:
ping @dhylands

dhylands:
We just need to make corresponding changes in ipcam.py.

The name field (that was removed) was intended to be for what the user names the camera. Is that still true for the name field in the properties?

Currently the client (ipcam.py) was using the service name field to figure our which services were cameras. Presumably, that's what the new type field would get used for, but I don't see if being set anywhere?

Are there other changes to go along with this? Since you removed the model field (which was used elsewhere in ip_camera_adapter/mod,rs) I don't see how it compiles any more.

src/adapters/ip_camera_adapter/mod.rs
dhylands, line 102: It's used in the upnp listener code in ip_camera_adapter/mod.rs

azasypkin:
> The name field (that was removed) was intended to be for what the user names the camera. Is that still true for the name field in the properties?

Mmm, I've only moved `model` to `custom_properties`, `name` is still in `IpCameraService`.

PR #176, opened 2016-03-09
Initial checkin of integration test framework for foxbox

nojunpark:
@JohanLorenzo r?

### DIFF
--- a/test/integration/lib/config/foxbox.js
+++ b/test/integration/lib/config/foxbox.js
@@ -1,0 +1,23 @@
+'use strict';
+
+module.exports = {
+
+  nupnp_server : {
+    id :'111788fffe230b96',
+    port: '8002'
+  },
+
+  credential :  {
+    'email': 'a@b.com',
+    'username': 'admin',
+    'password': '00000000'
+  },
+
+  foxbox : { 
+    url : 'http://localhost:3000'
+  },
+
+  philips_hue : {
+    port : 8001
+  }
+};
--- a/test/integration/lib/html/philips_initial.html
+++ b/test/integration/lib/html/philips_initial.html
@@ -1,0 +1,477 @@
+Bad JSON: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">
+<html>
+<head>
+  <title>hue personal wireless lighting</title>
+
+  <style type="text/css">
+  html, body, div, span, h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+  a, abbr, big, cite, code,
+  em, img, small, strong, b, u, i, center {
+    margin: 0;
+    padding: 0;
+    border: 0;
+    font-size: 100%;
+    font: inherit;
+    vertical-align: baseline;
+  }
+  body {
+    line-height: 1;
+    background-color: #252525;
+    font-family: Verdana;
+  }
+  .philips-header {
+    height: 41px;
+    width: 100%;
+    background-color: white;
+    position: relative;
+  }
+  .philips-logo {
+    position: absolute;
+    left: 23px;
+    top: 14px;
+  }
+  .header {
+    width: 100%;
+    height: 101px;
+    background-color: #000;
+    position: relative;        
+  }
+  .header-logo {
+    position: absolute;
+    top: 17px;
+    left: 20px;
+  }
+  .colorline {
+    bottom: -6px;
+    height: 9px;
+    position: absolute;
+    width: 100%;
+  }
+  .container {
+    width: 908px;
+    margin: 50px auto 0 auto;
+    background-color: white;
+    -webkit-border-radius: 2px 2px 0 0;
+    -moz-border-radius: 2px 2px 0 0;
+    border-radius: 2px 2px 0 0;
+  }
+  .inner-container {
+    padding: 35px;
+  }
+  h1 {    
+    height: 51px;
+    padding: 0 20px;        
+    line-height: 51px;
+    color: white;
+    font-size: 18px;
+    text-transform: uppercase;
+    box-sizing: border-box;
+    -moz-box-sizing: border-box;
+    -webkit-box-sizing: border-box;     
+    -webkit-border-radius: 2px 2px 0 0;
+    -moz-border-radius: 2px 2px 0 0;
+    border-radius: 2px 2px 0 0;
+    background: #505050; /* Old browsers */                
+    background: -moz-linear-gradient(top, #505050 0%, #000000 100%); /* FF3.6+ */
+    background: -webkit-gradient(linear, left top, left bottom, color-stop(0%,#505050), color-stop(100%,#000000)); /* Chrome,Safari4+ */
+    background: -webkit-linear-gradient(top, #505050 0%,#000000 100%); /* Chrome10+,Safari5.1+ */
+    background: -o-linear-gradient(top, #505050 0%,#000000 100%); /* Opera 11.10+ */
+    background: -ms-linear-gradient(top, #505050 0%,#000000 100%); /* IE10+ */
+    background: linear-gradient(to bottom, #505050 0%,#000000 100%); /* W3C */        
+  }
+  h2 {
+    font-size: 16px;
+    margin-bottom: 5px;
+    font-weight: bold;
+  }
+  a {
+    color: black;
+  }
+  a:hover {
+    text-decoration: none;
+  }
+  p {
+    font-size: 12px;
+    line-height: 18px;
+    margin-bottom: 15px;
+  }
+  b {
+    font-weight: bold;      
+  }
+  </style>
+</head>
+<body>
+  <div class="philips-header">
+    <img src="philips-blue.png" class="philips-logo" alt="Philips" />      
+  </div>
+  <div class="header">
+    <img class="header-logo" src="hue-logo.png" alt="hue personal wireless lighting" />
+
+    <img src="hue-color-line.png" class="colorline" />
+    
+  </div>
+  
+
+  <div class="container">
+    <h1>hue personal wireless lighting</h1>
+    <div class="inner-container">
+      <p>
+        Welcome to hue - your personal wireless lighting system. For more information please see our website <a href="http://www.meethue.com">www.meethue.com</a>
+      </p>
+      <h2>Open Source</h2>
+      <p>
+        This product contains software licensed under an open source license. For acknowledgements and license texts, <b>see below</b>. <b>Philips Lighting B.V.</b> hereby offers to deliver, upon request, a copy of the complete corresponding source code for the copyrighted open source software packages used in this product for which such delivery is requested by the respective licenses. This offer is valid up to three years after <b>purchase or software update</b> to anyone in receipt of this information. To obtain source code, please send your request in English with product type to open.source@philips.com. If you prefer not to use email or if you do not receive confirmation receipt within 2 weeks after mailing to this email address, please write in English to "Open Source Team, Philips Intellectual Property & Standards, P.O. Box 220, 5600 AE Eindhoven, The Netherlands". If you do not receive timely confirmation of your letter, please email to the email address above.
+      </p>
+      
+      <h2>SHA-1</h2>
+      <p>
+        Website: <a href="http://www.packetizer.com/security/sha1/">http://www.packetizer.com/security/sha1/</a>
+        <br>
+        License: <a href="sha_license.txt">sha_license.txt</a>
+      </p>
+      <h2>OpenSSL AES Encryption</h2>
+      <p>
+        Website: <a href="https://www.openssl.org/docs/apps/enc.html">https://www.openssl.org/docs/apps/enc.html</a>
+        <br>
+        License: <a href="aesenc_license.txt">aesenc_license.txt</a>
+      </p>
+      <h2>JSON_checker</h2>
+      <p>
+        Website: <a href="http://www.json.org/JSON_checker/">http://www.json.org/JSON_checker/</a>
+        <br>
+        License: <a href="jsonchecker_license.txt">jsonchecker_license.txt</a>
+      </p>
+      <h2>UTF-8 Decoder</h2>
+      <p>
+        Website: <a href="http://bjoern.hoehrmann.de/utf-8/decoder/dfa">http://bjoern.hoehrmann.de/utf-8/decoder/dfa</a>
+        <br>
+        License: <a href="utf8decoder_license.txt">utf8decoder_license.txt</a>
+      </p>
+      
+      <!-- Begin Licensing Information -->
+      <h2>openwrt-runtime </h2>
+      <p>
+        Website: <a href="http://www.openwrt.org">http://www.openwrt.org</a>
+        <br>
+        License: <a href="licenses/license_openwrt-runtime.txt">license_openwrt-runtime.txt</a>
+      </p>
+
+      <h2>zlib 1.2.8</h2>
+      <p>
+        Website: <a href="http://www.zlib.net">http://www.zlib.net</a>
+        <br>
+        License: <a href="licenses/license_zlib_1.2.8.txt">license_zlib_1.2.8.txt</a>
+      </p>
+
+      <h2>uboot 1.1.4</h2>
+      <p>
+        Website: <a href="http://www.denx.de/wiki/U-Boot/WebHome">http://www.denx.de/wiki/U-Boot/WebHome</a>
+        <br>
+        License: <a href="licenses/license_uboot_1.1.4.txt">license_uboot_1.1.4.txt</a>
+      </p>
+
+      <h2>busybox 1.19.4</h2>
+      <p>
+        Website: <a href="http://www.busybox.net/">http://www.busybox.net/</a>
+        <br>
+        License: <a href="licenses/license_busybox_1.19.4.txt">license_busybox_1.19.4.txt</a>
+      </p>
+
+      <h2>firewall 2014-03-20</h2>
+      <p>
+        Website: <a href="https://openwrt.org/">https://openwrt.org/</a>
+        <br>
+        License: <a href="licenses/license_firewall_2014-03-20.txt">license_firewall_2014-03-20.txt</a>
+      </p>
+
+      <h2>fstools 2014-04-07</h2>
+      <p>
+        Website: <a href="https://openwrt.org/">https://openwrt.org/</a>
+        <br>
+        License: <a href="licenses/license_fstools_2014-04-07.txt">license_fstools_2014-04-07.txt</a>
+      </p>
+
+      <h2>ppp-default 2.4.5</h2>
+      <p>
+        Website: <a href="">None</a>
+        <br>
+        License: <a href="licenses/license_ppp-default_2.4.5.txt">license_ppp-default_2.4.5.txt</a>
+      </p>
+
+      <h2>netifd 2014-04-07</h2>
+      <p>
+        Website: <a href="http://wiki.openwrt.org/doc/techref/netifd">http://wiki.openwrt.org/doc/techref/netifd</a>
+        <br>
+        License: <a href="licenses/license_netifd_2014-04-07.txt">license_netifd_2014-04-07.txt</a>
+      </p>
+
+      <h2>iw 3.17</h2>
+      <p>
+        Website: <a href="http://git.sipsolutions.net/iw.git/">http://git.sipsolutions.net/iw.git/</a>
+        <br>
+        License: <a href="licenses/license_iw_3.17.txt">license_iw_3.17.txt</a>
+      </p>
+
+      <h2>ed25519-donna 1.0</h2>
+      <p>
+        Website: <a href="https://github.com/floodyberry/ed25519-donna">https://github.com/floodyberry/ed25519-donna</a>
+        <br>
+        License: <a href="licenses/license_ed25519-donna_1.0.txt">license_ed25519-donna_1.0.txt</a>
+      </p>
+
+      <h2>libiwinfo </h2>
+      <p>
+        Website: <a href="http://wiki.openwrt.org/doc/devel/packages/iwinfo">http://wiki.openwrt.org/doc/devel/packages/iwinfo</a>
+        <br>
+        License: <a href="licenses/license_libiwinfo.txt">license_libiwinfo.txt</a>
+      </p>
+
+      <h2>libgcc 0.9.33.2</h2>
+      <p>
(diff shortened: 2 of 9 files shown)


test/integration/lib/html/philips_initial.html
```diff
@@ -0,0 +1,477 @@
+Bad JSON: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">
```
JohanLorenzo, line 1: Nit: See `Bad JSON:`

I also wonder if this page could be simpler. I mean, I don't see a reason why we have to copy the HTML page given by Philips Hue. Do you think we should do some tests against this page?

test/integration/lib/config/foxbox.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+
+	nupnp_server : {
```
JohanLorenzo: Nit: The rest of the project uses spaces instead of tab. I'd make the PR comply to it.

test/integration/lib/config/foxbox.js
```diff
@@ -0,0 +1,28 @@
...
+		id :"111788fffe230b96",
+		port: "8002"
+	},
+
+	credential :  {
+		"email": "npark@mozilla.com",
```
JohanLorenzo: Nit: Change the email by a dummy one.

test/integration/lib/config/foxbox.js
```diff
@@ -0,0 +1,28 @@
...
+	foxbox : { 
+		url : "http://localhost:3000"
+	},
+
+	philips_hue : {
+	port : 8001
+	},
+
+	header : {
+      "Content-Type" : "application/json"
```
JohanLorenzo: Nit: I don't remember, but I probably changed my mind since the last review. I don't see a use case where the header is depdendant of the environment. I'd move it to a js file instead.

test/integration/lib/json/bridge_status.json
```diff
@@ -0,0 +1,445 @@
...
+    "replacesbridgeid": null,
+    "backup": {
+      "status": "idle",
+      "errorcode": 0
+    },
+    "whitelist": {
+      "2aebf17e5cbd09c3712b116b68325934": {
+        "last use date": "2016-02-29T04:18:37",
+        "create date": "2016-02-26T01:44:55",
+        "name": "Hue#LGE Nexus 4"
```
JohanLorenzo: Nit: Clear the whitelist from personal data :)

test/integration/lib/json/bridge_status.json
```diff
@@ -0,0 +1,445 @@
...
+      },
+      "24955b892a1ccc434a83e3974a77f337": {
+        "last use date": "2016-02-26T06:48:37",
+        "create date": "2016-02-26T06:20:02",
+        "name": "my_hue_app#iphone peter"
+      }
+    }
+  },
+  "schedules": {},
+  "scenes": {
```
JohanLorenzo, line 120: I wonder if we could make the scene part smaller, and just have 2 scenes in it. What do you think?

test/integration/lib/json/bridge_status.json
```diff
@@ -0,0 +1,445 @@
+{
+  "lights": {
+    "1": {
+      "state": {
+        "on": true,
+        "bri": 254,
```
JohanLorenzo: Maybe we could set the HSB values to easy-to-remember ones? Like (255, 16384, 255) ?

test/integration/lib/nupnp_PhilipsHue.js
```diff
@@ -0,0 +1,19 @@
+var express = require ('express');
+
+var philips_nupnp_page = (function() {
+	return {
+		start: function(hue_id,hue_ipaddress,port) {
+			var app = express();
+
+			app.get('/', function (req, res) {
+				res.send([{"id":hue_id,"internalipaddress":hue_ipaddress}]);
+      		});
```
JohanLorenzo: Nit: Bad indentation

test/integration/lib/philipsHue_server.js
```diff
@@ -0,0 +1,83 @@
+var express = require ('express');
+var bodyParser = require('body-parser');
+var path = require('path');
+
+var philipsHue_server = (function() {
+
+	var welcome_page = '/html/philips_initial.html';		// Philips Hub webpage
+	var suc_resp = require('./success_response.js');		// TBD: move this to appropriate module
+	var api_resp = require('./json/bridge_status.json');	// full status response
+	var light_status_resp = require('./json/lights_status.json'); // light status response
```
JohanLorenzo: What is the advantage of requiring local modules here, compared to [there](https://github.com/fxbox/foxbox/pull/176/files#diff-aff726e949c461077c3a77d720cc4ad1R3) ?

test/integration/lib/success_response.js
```diff
@@ -0,0 +1,17 @@
+/* global module */
+(function(module) {
+  'use strict';
+
+  /**
+   * returns the light response
+   */
+  function light_response(light_id,result) {
+  	var response = {};
```
JohanLorenzo: Nit: Mixed tabs and spaces

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
+// This agent refers to PORT where program is runninng.
```
JohanLorenzo: I'm not sure to understand this comment. What does it mean?

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+      return chakram.get(config.get('foxbox.url') + "/services/list")
+      .then(function(listResponse) {
+        lightinfo = listResponse.body;
+        console.log("awesome, lights are:" + lightinfo);
+        expect(listResponse).to.have.status(200);
+      })
+    });
+
+    it("Turn on the second light", function(){
+    var lightID = lightinfo[1].id;
```
JohanLorenzo: This test depends on the previous one, which can be harmful in the future. Instead, I'd go like:

``` js
it("get a list of available philips services",() => { 
  // do the regular test
});

describe('once lights are known', () => {
  var lightsRawData;

  before(() => {
    return chakram.get(config.get('foxbox.url') + "/services/list")
      .then(res => {
        lightsRawData = res.body;
      });
  });

  it("Turn on the second light", () => {
      // do the regular test
  });
})
```

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+  var FOXBOX_STARTUP_WAIT_TIME_IN_MS = 5000;
+  var foxbox_process;
+  before(function(done){
+    this.timeout(500000);
+
+    // Start the nupnp server
+    const hue_location = "localhost:" + config.get('philips_hue.port');
+    nupnp_server.start(config.get('nupnp_server.id'),hue_location,config.get('nupnp_server.port'));
+    
+    // Start the Philips Hue simulator, and set the #2 lightbulb status to false
```
JohanLorenzo: This comment looks redundant with the next 2 lines

test/integration/lib/philipsHue_server.js
```diff
@@ -0,0 +1,83 @@
...
+	
+	var commanded_light_state ={1:true,2:true,3:true};		// TBD: move this to appropriate module
```
JohanLorenzo: Nit: Use `TODO`, `FIXME` or `XXX`. http://c2.com/cgi/wiki?FixmeComment

test/integration/lib/philipsHue_server.js
```diff
@@ -0,0 +1,83 @@
...
+	var last_received_cmd = null;		
+
+	return {
+
+		setlight: function(id,value) {  // Set the on/off status of simulated light
```
JohanLorenzo: I think we can make the meaning of `value` clearer, with something to what we did with BinaryButtons in Gaiatest. What do you think of `turnOnLight(id)`, `turnOffLight(id)` which both use this internal function called `_setLight(id, value)`?

test/integration/lib/philipsHue_server.js
```diff
@@ -0,0 +1,83 @@
+var express = require ('express');
+var bodyParser = require('body-parser');
+var path = require('path');
+
+var philipsHue_server = (function() {
+
+	var welcome_page = '/html/philips_initial.html';		// Philips Hub webpage
```
JohanLorenzo: Nit: This string is only [used once](https://github.com/fxbox/foxbox/pull/176/files#diff-aff726e949c461077c3a77d720cc4ad1R38). If we put the string directly there, people will have enough context to understand it's a path to the welcome page. Hence, I'd remove this var.

test/integration/lib/philipsHue_server.js
```diff
@@ -0,0 +1,83 @@
...
+				}
+				else {
+					res.status(200).json(api_resp);  
+				} 		
+			});
+
+			_server.get('/api/:foxbox_id/lights', function (req, res) {
+				if (req.params.foxbox_id == foxbox_id){
+					res.status(200).json(light_status_resp); 
+				}
```
JohanLorenzo: _No need to add it in this PR_
Improvement: Make the server return a failure if the foxbox_id is not the expected one.

test/integration/lib/philipsHue_server.js
```diff
@@ -0,0 +1,83 @@
...
+					foxbox_id = req.params.foxbox;
+					res.sendStatus(200); 
+				}
+				else {
+					res.status(200).json(api_resp);  
+				} 		
+			});
+
+			_server.get('/api/:foxbox_id/lights', function (req, res) {
+				if (req.params.foxbox_id == foxbox_id){
```
JohanLorenzo: Nit: Use `===` in JS, unless you're sure you want to use `==`

test/integration/lib/philipsHue_server.js
```diff
@@ -0,0 +1,83 @@
...
+			commanded_light_state[id] = value;
+		},
+
+		get last_cmd() {	
+			return last_received_cmd;
+		},
+
+		setup : function(port) {
+			var _server = express();
+			var foxbox_id = 'none';
```
JohanLorenzo: Nit: JS is usually a camel case language. We should use `foxboxId`.

Also, what value provides `'none'` compared to an `undefined` value?

test/integration/lib/upnp_PhilipsHue.js
```diff
@@ -0,0 +1,22 @@
...
+
+class PhilipsHub extends upnpServer {
+
+	constructor() {
+		super(Philips_Config);
+  		this._server.addUSN('urn:schemas-upnp-org:device:Basic:1');
+	}
+
+	startServer (something) {
+  		// Start the server, below is the example of calling superclass method
```
JohanLorenzo: Nit: Redundant comment

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+    philipshue_server.setup(config.get('philips_hue.port'));
+    philipshue_server.setlight(2,false);
+
+    // kick off the foxbox within 10 seconds
+    remove_sqlite = spawn('rm',['users_db.sqlite']);
```
JohanLorenzo: Nit: Use [fs.unlink](http://stackoverflow.com/questions/5315138/node-js-remove-file), which is cross-platform

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+  it("create a new credential",function(){
+
+    return chakram.post(config.get('foxbox.url') + "/users/setup",credential)
+    .then(function(setupResp){
+      expect(setupResp).to.have.status(201);
+    });
+  });
+
+  it("login to foxbox",function(){
+    header['Authorization'] = 'Basic ' + encoded_cred;  // supply the credential used in previous test
```
JohanLorenzo: :warning:  `encoded_cred` is undefined. I suspect this is a leftover of an old commit

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+     });
+    });
+
+    it("get a list of available philips services",function(){   
+
+      return chakram.get(config.get('foxbox.url') + "/services/list")
+      .then(function(listResponse) {
+        lightinfo = listResponse.body;
+        console.log("awesome, lights are:" + lightinfo);
+        expect(listResponse).to.have.status(200);
```
JohanLorenzo, line 88: I'd add more checks on this requests. For instance, make sure we have 3 lights.

test/integration/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/fxbox/foxbox.git"
+  },
+  "keywords": [
+    "foxbox"
+  ],
+
+  "dependencies": {
+    "chai": "^3.5.0",
```
JohanLorenzo: Nit: Express is not installed

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+       chakram.setRequestDefaults({'headers': header});
+     });
+    });
+
+    it("get a list of available philips services",function(){   
+
+      return chakram.get(config.get('foxbox.url') + "/services/list")
+      .then(function(listResponse) {
+        lightinfo = listResponse.body;
+        console.log("awesome, lights are:" + lightinfo);
```
JohanLorenzo: Nit: Clean up. This log line doesn't add much value once executed:

```
awesome, lights are:[object Object],[object Object],[object Object]
```

test/integration/lib/upnp_PhilipsHue.js
```diff
@@ -0,0 +1,22 @@
...
+  		console.log(something + " upnp simulation started");
```
JohanLorenzo: Nit: I didn't see this line being printed out. Is it used?

test/integration/lib/upnpserver_base.js
```diff
@@ -0,0 +1,43 @@
...
+class upnpServer {
+  constructor(parameters) {
+    this.parameters = parameters;
+    console.log(parameters);
+    this._server = new Server(parameters);
+  }
+
+  createServer() {
+    // Define callbacks
+    console.log("Defining callbacks");
```
JohanLorenzo: Nit: I didn't see any console.log being printed out when ran `npm test`

test/integration/lib/philipsHue_server.js
```diff
@@ -0,0 +1,83 @@
...
+				if (req.params.foxbox_id == foxbox_id){
+					// return the appropriate subset of light status
+					res.status(200).json(light_status_resp[req.params.light_id]); 
+				}
+			});
+
+			_server.put('/api/:foxbox_id/lights/:light_id/state', bodyParser.json(),function (req, res) {
+				// set the status of the simulated lightbulb. currently it only handles on/off
+				// TBD: set sat, bri, hue values 
+				console.log(req.body);
```
JohanLorenzo: Nit: This output is a bit confusing, when I run `npm test`:

```
      ✓ get a list of available philips services
{ bri: 254, hue: 14910, on: true, sat: 100 }
```

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+  var key = credential['username']+":"+credential['password'];
+  var encoded_cred = new Buffer(key).toString('base64');
+  var header = config.get('header');
+  var lightinfo; // to store the light status received from foxbox
+  var FOXBOX_STARTUP_WAIT_TIME_IN_MS = 5000;
+  var foxbox_process;
+  before(function(done){
+    this.timeout(500000);
+
+    // Start the nupnp server
```

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+      expect(loginResp).to.have.status(201);
+    });
+  });
+
+  describe ("once logged in", function () {
+    before(function() {
+      return chakram.post(config.get('foxbox.url') + "/users/login",null,{'headers' : header})
+      .then(function(resp){
+       header['Authorization'] = 'Bearer ' + resp.body.session_token;
+       chakram.setRequestDefaults({'headers': header});
```
JohanLorenzo, line 79: I'd add an after function where we reset the headers to a known state.

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,99 @@
...
+    remove_sqlite = spawn('rm',['users_db.sqlite']);
+    foxbox_process = spawn('../../target/debug/foxbox', 
+      ['-c', 'philips_hue;nupnp_url;http://localhost:'+ config.get('nupnp_server.port')+'/']);
+    setTimeout(done, FOXBOX_STARTUP_WAIT_TIME_IN_MS);
+    
+  });
+  
+  after(function(done){
+    foxbox_process.kill();
+    done();
```
JohanLorenzo: Nit: childProcess.kill() [is sync](https://nodejs.org/api/child_process.html#child_process_child_kill_signal). No need to define `done()`

JohanLorenzo:
~~r? @JohanLorenzo~~ 

It looks good overall. I noticed some minor errors here and there, but nothing that creates false positives. Tests are passing on my machine. We have to make them run on Travis-CI, but I'm fine if we handle that in a follow up. This PR is already big enough :smiley: 

Please run `jshint` on folders tests/integration/{lib,test}

test/integration/lib/html/philips_initial.html
nojunpark, line 1: I included  this because the first thing foxbox does is to get '/', and this is what philips hue returns, I suspect foxbox is only concerned about '200 OK' response, but I figured maybe I'll put the whole page just to be safe.

test/integration/lib/json/bridge_status.json
nojunpark: right, foxbox isn't using whitelist so I'll remove the whole entry

test/integration/lib/json/bridge_status.json
nojunpark, line 120: sure, we can always add more scenes later here if we need to

test/integration/lib/json/bridge_status.json
nojunpark: agreed

test/integration/lib/upnpserver_base.js
nojunpark: oh this, it's because upnpserver_base.js is not being used yet, because philips hub is stil Nupnp

test/integration/lib/upnp_PhilipsHue.js
nojunpark: same here, upnp is not used to detect Philips Hue yet, what we use is Nupnp

test/integration/test/philips_light_test.js
nojunpark: actually, it is defined up top, I'll make it as a local var

test/integration/test/philips_light_test.js
nojunpark, line 79: Hmm, should we?  we pretty much need the token throughout the session, and if we decide to use a new account, we have to supply a new token anyways - there isn't a case where we need to explictly clear the header, what do you think?

test/integration/lib/config/foxbox.js
nojunpark: strange, I changed the tab setting to 2 spaces, but here it still shows up as the same.

JohanLorenzo: It doesn't appear in the last commits' changes. Maybe you need to make a modification within the file, so the IDE changes all the rest?

test/integration/lib/upnpserver_base.js
JohanLorenzo: Ok. Could we make the PR smaller and not land Upnp, in this case? So we comply with the YAGNI principle

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,114 @@
...
+    nupnp_server.start(config.get('nupnp_server.id'),hue_location,config.get('nupnp_server.port'));
+    
+    philipshue_server.setup(config.get('philips_hue.port'));
+
+    philipshue_server.turnOffLight(1);
+    philipshue_server.turnOffLight(2);
+    philipshue_server.turnOffLight(3);
+
+    // kick off the foxbox within 10 seconds
+    //remove_sqlite = spawn('rm',['users_db.sqlite']);
```
JohanLorenzo: Nit: Commented code.

test/integration/test/philips_light_test.js
```diff
@@ -0,0 +1,114 @@
...
+      before(function() {
+        return chakram.get(config.get('foxbox.url') + "/services/list")
+        .then(function(listResponse) {
+          lightinfo = listResponse.body;
+        });
+      });
+
+      // Currently, there is no mapping between the foxbox issues id and the philips hue id
+      // until the tag feature is implemented
+      it("Turn on all lights", function(){
```
JohanLorenzo: This could be factorized [this way](http://mochajs.org/#dynamically-generating-tests):

``` js
[0, 1, 2].forEach(id => {
  it('turn light with id ' + id, () => {
     return chakram.put(config.get('foxbox.url') + "/services/" + lightinfo[id].id + "/state", {"on": true})
       .then(res => {
        expect(res).to.have.status(200);
        expect(res.body['result']).equals('success');
        expect(philipshue_server.last_cmd()).to.contain('"on":true');
        expect(philipshue_server.lightStatus(id)).equals(true);
    });
  });
});
```

test/integration/lib/config/foxbox.js
nojunpark: ah, it's just my IDE interprets tabs as 2 spaces.  I have to replace the tab to 2 spaces

test/integration/test/philips_light_test.js
nojunpark: hmm, when I tried this, it looks like the test finishes before waiting for expect statement evaluation, even when there is a return call for chakram API use.  I guess the async nature of expect doesn't like foreach()? 

Another thing is that the order of lightinfo is random, and they do not correspond to the philips hub ID.  so lightinfo[0] may be lightStatus[1] or [2].  That's why I had to separately check all lights are on, in case foxbox generated duplicate IDs for different lights.

You'll notice this if you run my test repeatedly - when the lights are on, in some instances, it is in order of 1 - 3 -2 although in most cases it's 1 - 2 - 3.

JohanLorenzo: Oh that sounds correct. I looked up some ways to deal with async/mocha/forEach and it looks a bit more complex. Let's go with your solution first, and modify it in a follow up.

JohanLorenzo:
r+ ! Please squash review/jshint commits with the others.

PR #178, opened 2016-03-10
Add local config file to .gitignore. r=fabrice

ferjm:


.gitignore
```diff
@@ -25,3 +25,6 @@ node_modules
 
 # Shared copied assets
 /static/**/shared
+
+# Foxbox config
+foxbox.conf
```
fabricedesre, line 30: Ok for now. But I really need to implement this "profile" feature instead...

PR #179, opened 2016-03-10
Move from ngrok to pagekite and add bridge to ws. r=sgiles

ferjm:


arcturus:
Tried the branch and working pretty nice :)

cargo run -- -t knilxof.org:443 -s foxbox -u box22.knilxof.org

But it will need rebase r+!

PR #180, opened 2016-03-10
Major rework of the adapter API

Yoric:
Given the size that the adapter API was taking, I have moved it to [its own repo](https://github.com/fxbox/adapters/). The clock now builds against a nearly full implementation of the Adapter API/Taxonomy API.

I still need to write a gazillion tests for the above repo and to handle the case in which channels are (dis)connected while we are watching them. Opening this PR now to avoid being a blocker. I'll try and write more tests today, after I have eaten, slept and generally behaved somewhat like a human being.


fabricedesre:
This is a core piece of the platform, so I'd rather have that in the main repo. Did you "fork" it just temporarily?

Yoric:
Well, my personal intuition is that keeping small, separate, crates, with well-bounded objectives is a good way to improve maintainability.

As a minor example of this, having foxbox_adapters in a separate crate makes it extremely clear that it's sorely lacking tests :)

fabricedesre:
I don't think that gives us much compared to sub directories in the main repo, and it's annoying when you want to make local changes since you need to work on a clone, change your Cargo.toml, etc.
Also, this crate has no point being reusable outside of foxbox if I'm not mistaken.

Yoric:
Actually, there is kind of a good reason. For the moment, thinkerbell is on its own repo, because it's still quite experimental. As soon as I have time to update it, it will start depending on this API for testing. So if/when we bring adapters into foxbox, we'll need to bring thinkerbell, too.

That's not a real blocker, of course, but at least it suggests that we don't want to bring adapters into foxbox immediately.

Yoric:
Ah, I realize that I forgot to r? @fabricedesre 

Note that the failing check is a dependency that won't build on Travis. Locally, all tests are green.

fabricedesre:
r=me with the clippy change reverted.

Yoric:
I also removed the tests which have been rewritten more completely in taxonomy_adapters.

PR #182, opened 2016-03-10
Add editorconfig

JohanLorenzo:
In #176, the IDE started to introduce tabs. I'd propose to make sure everybody has the same IDE config.

r? @fabricedesre

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.js]
+indent_size = 2
```
fabricedesre, line 15: nit: also indent html with 2 spaces?

JohanLorenzo, line 15: Done.

Issue #142, opened 2016-03-07
Store database and config files in a profile directory.


fabricedesre:
Right now they end up in the directory foxbox was launched from.

jedireza:
Just came across this new create [`preferences`](https://crates.io/crates/preferences) maybe it could help or inspire a solution.

https://crates.io/crates/preferences

hfiguiere:
I'll have a PR for fxbox/users#53 that will allow setting the path for the user DB. It should help implementing this.

fabricedesre:
Working on that today.

fabricedesre:
Fixed by https://github.com/fxbox/foxbox/issues/184

PR #183, opened 2016-03-10
Fix UserDb API usage following changes for issue fxbox/users#53

hfiguiere:
We need fxbox/users#66 to be merged. This reference my own branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,7 +9,7 @@ dependencies = [
  "env_logger 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "foxbox_adapters 0.1.0 (git+https://github.com/fxbox/adapters.git?rev=ba54c80)",
  "foxbox_taxonomy 0.1.2 (git+https://github.com/fxbox/taxonomy.git?rev=4ff36cd)",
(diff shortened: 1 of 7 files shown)


hfiguiere:
@fabricedesre this might be useful for issue #142. Albeit here I hardcode the file name instead of getting it from the configuration.

hfiguiere:
This PR needs some work. Test failures.

ferjm:
FWIW I'd like to take a different approach https://github.com/fxbox/users/pull/66#issuecomment-195036992

hfiguiere:
@fabricedesre @ferjm r? (need the changes for fxbox/users#53 and a readjustment of the Cargo.toml once they land)

hfiguiere:
(rebased again)

Also I support the new stuff from PR #184

hfiguiere:
I have addressed 2 of the 3 comments. I still believe we should not have a default path for the database. Because:
1. if it is relative to cwd() it is error prone. Change the working directory and boom.
2. if it is absolute it is also error prone as there is a strong chance that another library user could try to access it.

hfiguiere:
Now that the other PR has been merged, this is read for @fabricedesre @ferjm r?

ferjm:
LGTM. Thanks!

PR #189, opened 2016-03-11
Introduce basic MJPEG streaming for IP Camera Adapter. r=dhylands - WIP!

azasypkin:


fabricedesre:
That would make the video stream go through the tunnel, which is not the kind of scaling game we want to play.
Instead we need to negotiate a webrtc media channel.

PR #190, opened 2016-03-11
Fix OPTIONS /services/:service/:cmd 501. r=azasypkin

ferjm:
r? @azasypkin


azasypkin:
r+, looks and works good for me. Thanks for the quick fix!

PR #196, opened 2016-03-12
[Wip] taxonomy router

fabricedesre:
@Yoric this is a (non working) initial pass at exposing the taxonomy on http.

Surprise surprise, we have threading issues. I had to comment the |router.get(...)| part so that clippy let me commit that, but what happens is that:
- API should be Send because the http handlers are called on another thread by the http layer.
- Making API Send lead to a snowball of changes in backend.rs, at least using tons of `Arc<Mutex<X>>` instead of `Rc<RefCell<X>>`.

At this point I'm not sure whether the way to go is to make API be Send, or refactor that differently, eg. wrapping the API with channels.

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -9,6 +9,7 @@ extern crate mio;
 use adapters::AdapterManager;
 use config_store::ConfigService;
 use core::marker::Reflect;
+use foxbox_adapters::manager::AdapterManager as AdapterManager2;
 use foxbox_users::UsersManager;
 use http_server::HttpServer;
 use iron::{Request, Response, IronResult};
@@ -106,6 +107,7 @@ impl FoxBox {
 impl Controller for FoxBox {
 
     fn run(&mut self, shutdown_flag: &AtomicBool) {
+
         debug!("Starting controller");
         let mut event_loop = mio::EventLoop::new().unwrap();
 
@@ -122,11 +124,15 @@ impl Controller for FoxBox {
             self.certificate_manager.reload_from_directory(certificate_directory).unwrap_or(());
         }
 
-        HttpServer::new(self.clone()).start();
-        WsServer::start(self.clone(), self.hostname.to_owned(), self.ws_port);
+        // Create the taxonomy based AdapterManager
+        let taxo_manager = AdapterManager2::new();
 
         let mut adapter_manager = AdapterManager::new(self.clone());
-        adapter_manager.start();
+        adapter_manager.start(&taxo_manager);
+
+        HttpServer::new(self.clone()).start(taxo_manager);
+        WsServer::start(self.clone(), self.hostname.to_owned(), self.ws_port);
+
         self.upnp.search(None).unwrap();
 
         event_loop.run(&mut FoxBoxEventLoop {
--- a/src/taxonomy_router.rs
+++ b/src/taxonomy_router.rs
@@ -1,0 +1,273 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+extern crate serde_json;
+
+use controller::Controller;
+use foxbox_adapters::manager::WatchGuard;
+use foxbox_taxonomy::api::{ API, TargetMap };
+use foxbox_taxonomy::values::Value;
+use foxbox_taxonomy::selector::{ GetterSelector, ServiceSelector, SetterSelector };
+use foxbox_taxonomy::services::TagId;
+use foxbox_taxonomy::util::Id;
+use foxbox_users::AuthEndpoint;
+use iron::{ Handler, IronResult, Request, Response };
+use iron::headers::ContentType;
+use iron::method::Method;
+use iron::prelude::Chain;
+use iron::status::Status;
+use serde::Serialize;
+use std::io::Read;
+use std::sync::Mutex;
+
+/// This is a specialized Router for the taxonomy API.
+/// It handles all the calls under the api/v1/ url space.
+pub struct TaxonomyRouter<A> {
+    api: Mutex<A>
+}
+
+impl<A> TaxonomyRouter<A>
+    where A: API<WatchGuard=WatchGuard> + 'static {
+    pub fn new(adapter_api: A) -> Self {
+        TaxonomyRouter {
+            // This locks the full api access when dealing with a http request,
+            // which basically kills http concurrency.
+            // TODO: make it concurrent again.
+            api: Mutex::new(adapter_api)
+        }
+    }
+
+    /// Build a json http response from a Serializable object.
+    fn build_response<S: Serialize>(&self, obj: &S) -> IronResult<Response> {
+        let serialized = itry!(serde_json::to_string(obj));
+        let mut response = Response::with(serialized);
+        response.status = Some(Status::Ok);
+        response.headers.set(ContentType::json());
+        Ok(response)
+    }
+}
+
+impl<A> Handler for TaxonomyRouter<A>
+    where A: API<WatchGuard=WatchGuard> + 'static {
+
+    #[allow(cyclomatic_complexity)]
+    fn handle(&self, req: &mut Request) -> IronResult<Response> {
+        // We are handled urls relative to the mounter set up in http_server.rs
+        // That means that for a full url like http://localhost/api/v1/services
+        // the req.url.path will only contain ["services"]
+        let path = req.url.path.clone();
+
+        /// Generates the code for a generic HTTP call, where we use an empty
+        /// taxonomy selector for GET requests, and a decoded json body for POST ones.
+        /// $call is the method we'll call on the api, like get_services.
+        /// $sel  is the selector type, like ServiceSelector
+        /// $path is a vector describing the url path, like ["service", "tags"]
+        macro_rules! get_post_api {
+            ($call:ident, $sel:ident, $path:expr) => (
+            if path == $path {
+                return {
+                    let api = self.api.lock().unwrap();
+
+                    if req.method != Method::Get && req.method != Method::Post {
+                        // Wrong method, return a 405 error.
+                        return Ok(Response::with((Status::MethodNotAllowed,
+                                                  format!("Bad method: {}", req.method))));
+                    }
+
+                    let result = if req.method == Method::Get {
+                        // On a GET, just send the full taxonomy content for
+                        // this kind of selector.
+                        api.$call(vec![$sel::new()])
+                    } else {
+                        // TODO: switch to serde_json::from_reader(req.body)
+                        let mut payload = String::new();
+                        itry!(req.body.read_to_string(&mut payload));
+                        let selectors: Vec<$sel> =
+                            itry!(serde_json::from_str(&payload));
+
+                        api.$call(selectors)
+                    };
+
+                    self.build_response(&result)
+                }
+            })
+        }
+
+        // Generates the code to process a given HTTP call with a json body.
+        macro_rules! payload_api {
+            ($call:ident, $param:ty, $path:expr, $method:expr) => (
+                if path == $path && req.method == $method {
+                    return {
+                        let api = self.api.lock().unwrap();
+                        let mut payload = String::new();
+                        itry!(req.body.read_to_string(&mut payload));
+                        let selectors: $param =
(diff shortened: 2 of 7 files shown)


Yoric:
Not good :/ I'll see what I can do about `Send`.

Yoric:
Fixed the `Send` in:
- https://github.com/fxbox/taxonomy/commit/b6d0349fe4a6d994a91207eb82e8ab5167639496 (trait)
- https://github.com/fxbox/adapters/commit/a7be98862e04f333bfa291eb02bb2ece7d404fd4 (implementation).

For the moment, I'm taking the easiest route out of this. Instead of a `Mutex`, I now delegate all the back-end to a dedicated thread and I use channels to communicate with the thread. The API remains sync (so I'm doing a `rx.recv()`), but I could easily publish an async API if you prefer.

fabricedesre:
The core constraint we have is that the closure passed to router.get("services", handler) needs to implement Send + Sync + Any because it's a Iron `handler` : http://ironframework.io/doc/iron/middleware/trait.Handler.html

Yoric:
`Sync` closures/traits are _extremely_ limited. Right now, I can't think of any way to do, well, anything useful other than a `map` or a `filter` with a `Sync` closure. I'll see if I can find an idea.

Yoric:
Well, there is `impl<T: ?Sized + Send> Sync for Mutex<T>`. But that would be a bit disappointing...

Yoric:
Wait, does this mean that `Iron` is essentially both single-threaded and synchronous? Or am I missing something?

fabricedesre:
As far as I know it's multi threaded.

Yoric:
I don't see how to get a multi-threaded back-end to work with `Iron` without a big fat `Mutex` around everything. Perhaps we should ask people specialists.

Yoric:
A [hyper-style `Handler`](http://ironframework.io/doc/hyper/server/trait.Handler.html) might be easier to manipulate.

fabricedesre:
With this PR I get /api/v1/services to return : 
[{"tags":[],"id":"service:clock@link.mozilla.org","getters":{"getter:timeofday.clock@link.mozilla.org":{"tags":[],"id":"getter:timeofday.clock@link.mozilla.org","service":"service:clock@link.mozilla.org","mechanism":{"kind":{"CurrentTimeOfDay":[]},"poll":1,"trigger":null,"watch":true,"updated":null},"adapter":"clock@link.mozilla.org","last_seen":null},"getter:timestamp.clock@link.mozilla.org":{"tags":[],"id":"getter:timestamp.clock@link.mozilla.org","service":"service:clock@link.mozilla.org","mechanism":{"kind":{"CurrentTime":[]},"poll":1,"trigger":null,"watch":true,"updated":null},"adapter":"clock@link.mozilla.org","last_seen":null}},"setters":{},"adapter":"clock@link.mozilla.org"}]

This is quite ugly... We should try to clean this up, by not returning `null` properties and empty arrays.

The Iron change has been submitted upstream (https://github.com/iron/iron/pull/436), and we also need a patch on foxbox_adapters (https://github.com/fxbox/adapters/pull/5) and the tip of foxbox_taxonomy.

fabricedesre:
Using serde_json::to_string_pretty() helps with readability/debugging:

```
[
  {
    "tags": [],
    "id": "service:clock@link.mozilla.org",
    "getters": {
      "getter:timestamp.clock@link.mozilla.org": {
        "tags": [],
        "id": "getter:timestamp.clock@link.mozilla.org",
        "service": "service:clock@link.mozilla.org",
        "mechanism": {
          "kind": {
            "CurrentTime": []
          },
          "poll": 1,
          "trigger": null,
          "watch": true,
          "updated": null
        },
        "adapter": "clock@link.mozilla.org",
        "last_seen": null
      },
      "getter:timeofday.clock@link.mozilla.org": {
        "tags": [],
        "id": "getter:timeofday.clock@link.mozilla.org",
        "service": "service:clock@link.mozilla.org",
        "mechanism": {
          "kind": {
            "CurrentTimeOfDay": []
          },
          "poll": 1,
          "trigger": null,
          "watch": true,
          "updated": null
        },
        "adapter": "clock@link.mozilla.org",
        "last_seen": null
      }
    },
    "setters": {},
    "adapter": "clock@link.mozilla.org"
  }
]
```

fabricedesre:
The iron changes landed upstream and I cherry picked them in https://github.com/fxbox/foxbox/commit/c9cb6cb06a58bba56643e902d0cc6a12e8ef0fdf

fabricedesre:
This POST request as an example of selecting a single service:
`curl -X POST -d '[{"id":{"Exactly":"service:clock@link.mozilla.org"}}]' http://localhost:3000/api/v1/services`

Yoric:
\o/

fabricedesre:
The taxonomy_router starts to be somewhat clean. Still incomplete, and lacks testing.

fabricedesre:
@JohanLorenzo I disabled kcov in this PR because it SIGSEVs. I can also reproduce locally, but haven't tried with the latest tip version. That's pretty sad...

@Yoric : Can you take a look? I'd like to land that soon to unblock people working on the adapters.
I'll add tests before merging anyway.

fabricedesre:
Added a couple of tests. They fail intermittently locally, because the json serialization is sometimes listing the getters in a different order. I'm not sure what to do about that.
I tried to not compare the json strings and instead compare Vec<Service> objects but we'll need to implement PartiaEq on Service for that to work. @Yoric what do you think?

Yoric:
That's weird, because I don't see any place where we could have non-determinism.

I have a small fear about adding `PartialEq` to `Service`. I expect that if we do, someone is going to start using it as a shortcut for `service_1.id == service_2.id`, whereas tests need it to also check the getters/setters.

How annoying would it be for you to rather implement comparing `Vec<Service>` inside the tests?

src/controller.rs
```diff
@@ -95,16 +96,23 @@ impl FoxBox {
...
+
+        // Create the taxonomy based AdapterManager
+        let taxo_manager = AdapterManager2::new();
+
         let mut adapter_manager = AdapterManager::new(self.clone());
-        adapter_manager.start();
+        adapter_manager.start(&taxo_manager);
+
+        HttpServer::new(self.clone()).start(taxo_manager);
+        WsServer::start(self.clone(), self.hostname.to_owned(), self.ws_port);
```
Yoric, line 134: We will eventually need to connect the WsServer to the `AdapterManager`, right?

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+use iron::{ Handler, IronResult, Request, Response };
+use iron::headers::ContentType;
+use iron::method::Method;
+use iron::prelude::Chain;
+use iron::status::Status;
+use serde::Serialize;
+use std::io::Read;
+use std::sync::Mutex;
+
+// This is a specialized Router for the taxonomy API.
```
Yoric: Nit: `///`, for the sake of cargo doc?

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+pub struct TaxonomyRouter<A> {
+    api: Mutex<A>
```
Yoric, line 27: Can you add a comment here (and possibly open an issue) regarding the performance issue?

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    #[allow(cyclomatic_complexity)]
+    fn handle(&self, req: &mut Request) -> IronResult<Response> {
+        // We are handled urls relative to the mounter set up in http_server.rs
+        // That means that for a full url like http://localhost/api/v1/services
+        // the req.url.path will only contain ["services"]
+        let path = req.url.path.clone();
+
+        // Generates the code for a generic HTTP call, where we use an empty
+        // selector for GET requests, and a decoded json body for POST ones.
```
Yoric: What does "selector" mean in this context? I assume that's Taxonomy selectors, right? If so, could you document that `sel` is the name of a type of `Selector`s?

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+                    let api = self.api.lock().unwrap();
+
+                    if req.method != Method::Get && req.method != Method::Post {
+                        // Wrong method, return a 405 error.
+                        return Ok(Response::with((Status::MethodNotAllowed,
+                                                  format!("Bad method: {}", req.method))));
+                    }
+
+                    let result = if req.method == Method::Get {
+                        // On a GET, just send the full service list.
```
Yoric: And what does "service" mean in this context? Taxonomy Service?

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+        // Generates the code to process a given HTTP call with a json body.
+        macro_rules! payload_api {
+            ($call:ident, $param:ty, $path:expr, $method:expr) => (
+                if path == $path && req.method == $method {
+                    return {
+                        let api = self.api.lock().unwrap();
+                        let mut payload = String::new();
+                        itry!(req.body.read_to_string(&mut payload));
+                        let selectors: $param =
+                            itry!(serde_json::from_str(&payload));
```
Yoric: Do I read correctly that we're parsing twice to json?

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+
+        // Keep these urls in sync with the AuthEndpoint(s) in the create() method.
+
+        // Selectors queries.
+        get_post_api!(get_services, ServiceSelector, ["services"]);
+        get_post_api!(get_getter_channels, GetterSelector, ["channels", "getters"]);
+        get_post_api!(get_setter_channels, SetterSelector, ["channels", "setters"]);
+
+        // Fetching and getting values.
+        payload_api!(fetch_values, Vec<GetterSelector>, ["channels", "get"], Method::Post);
```
Yoric: Should probably be `GET` instead of `POST`.

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+        payload_api!(send_values, TargetMap<SetterSelector, Value>, ["channels", "set"], Method::Post);
```
Yoric: Should probably be `PUT` instead of `POST`.

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+
+        // Adding tags.
```
Yoric, line 150: These, too, should probably be `PUT` instead of `POST`.

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+                      tags => Vec<Id<TagId>>,
+                      ["channels", "getter", "tags"], Method::Delete);
+        payload_api2!(remove_setter_tags,
+                      setters => Vec<SetterSelector>,
+                      tags => Vec<Id<TagId>>,
+                      ["channels", "getter", "tags"], Method::Delete);
+
+        // Fallthrough, returning a 404.
+        Ok(Response::with((Status::NotFound,
+                           format!("Unknwon url: {}", req.url))))
```
Yoric: Typo in the string.

src/taxonomy_router.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+}
+
+pub fn create<T, A>(controller: T, adapter_api: A) -> Chain
+    where A: API<WatchGuard=WatchGuard> + 'static,
+          T: Controller {
+    let router = TaxonomyRouter::new(adapter_api);
+
+    let auth_endpoints = if cfg!(feature = "authentication") && !cfg!(test) {
```
Yoric, line 189: I don't know that code yet, so can't review that part.

Yoric:
r+ with a few nits/suggestions.

JohanLorenzo:
kcov: Thanks for the heads up. I didn't manage to repro locally on your latest revision. I'll look into it.

src/controller.rs
fabricedesre, line 134: Yes, and it will use register_channel_watch().

src/taxonomy_router.rs
fabricedesre: I did that initially, and that's correct semantically, but it's also very unusual to have body payloads with `GET` requests. I'll switch though.

src/taxonomy_router.rs
fabricedesre, line 150: `PUT` usually means "replace" and `POST` "Create/add" so I'll keep them as Post.

PR #200, opened 2016-03-14
Adapters should not handle options requests. r=azasypkin

ferjm:
r? @azasypkin


azasypkin:
Looks good to me, r+!

PR #209, opened 2016-03-15
Store image snapshots in the profile directory

fabricedesre:
r? @dhylands 
I could not test locally. That should put the snapshot directory under $XDG_DATA_HOME or $HOME/.local/share.


dhylands:
r+

I tested and it seems to be working fine under linux.

Issue #210, opened 2016-03-15
Beautify logger output and log to file


cr:
The debug output is becoming very noisy, and it's getting hard to distinguish the lines. Two improvements would be helpful at this point:
- Having a logger implementation that beautifies the the console output with termcolors and neat formatting would make terminal logging much more readable.
- Logging to a file would make postmortem analysis easier.

This would be a concise little project suitable for Rust beginners, so help would be much appreciated. The general approach for implementing a custom logger is sketched at https://doc.rust-lang.org/log/log/index.html#logger-implementations.

PR #211, opened 2016-03-15
Colorizing logger output

cr:
Addresses the colorful part of issue #210.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -185,18 +185,43 @@ fn main() {
         libc::signal(SIGINT, mem::transmute(handle_sigint));
     }
 
-    let format = |record: &LogRecord| {
-        let t = time::now();
-        format!("{}.{:03}: {}{:5} {}",
-            time::strftime("%Y-%m-%d %H:%M:%S", &t).unwrap(),
-            t.tm_nsec / 1000_000,
-            tid_str(),
-            record.level(),
-            record.args()
-        )
-    };
     let mut builder = LogBuilder::new();
-    builder.format(format).filter(None, LogLevelFilter::Info);
+    let istty = unsafe { libc::isatty(libc::STDERR_FILENO as i32) } != 0;
+    if istty {
+        // Colorized output formatter
+        let format = |record: &LogRecord| {
+            let t = time::now();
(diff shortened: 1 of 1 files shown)


dhylands:
I'd recommend a few additional changes:
1 - Call isatty (or whatever the rust equivalent is) and only colorize if isatty returns true (so that you don't get all of the escape codes when redirecting stdout to a file.).
2 - Add a --nocolor option for those cases/platforms where colorizing doesn't work

src/main.rs
```diff
@@ -166,18 +166,43 @@ fn main() {
...
+    if istty {
+        // Colorized output formatter
+        let format = |record: &LogRecord| {
+            let t = time::now();
+            let level_color = match record.level() {
+                log::LogLevel::Error => "\x1b[1;31m",  // bold red
+                log::LogLevel::Warn  => "\x1b[1;33m",  // bold yellow
+                log::LogLevel::Info  => "\x1b[1;32m",  // bold green
+                log::LogLevel::Debug => "\x1b[1;34m",  // bold blue
+                log::LogLevel::Trace => "\x1b[1;35m"   // bold magenta
```
dhylands, line 199: You're faster than I am :)

I was going to suggest putting a no-color escape at the end of the line so that println!'s don't get colored the same as the last log message.

And persaonly, I prefer Info to uncolored.

cr, line 199: Line 182 resets the color with **\x1b[0m** right after the log level keyword is printed, so that keyword is the only part of the line that is color coded. The color is unable to bleed over into _println!_s.

We can still add color themes later if you don't like the green. ;)

dhylands, line 199: Ahh - ok - I missed that.

I'm used to the colored-logcat output for adb, which assigns a color to each PID and colorizes the left portion of the display using that, and then colors the right portion using the log level.
https://www.dropbox.com/s/2bxgmp9wzh0r0um/Screenshot%20from%202016-03-15%2019%3A11%3A17.png?dl=0
The colored logcat script is here: https://github.com/dhylands/colored-logcat

samgiles:
r+

PR #212, opened 2016-03-15
Make IP Camera adapter to accept "filename" query string parameter. r=dhylands

azasypkin:


Cargo.lock
```diff
@@ -9,7 +9,7 @@ dependencies = [
  "env_logger 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
```
azasypkin, line 9: Please ignore changes in `Cargo.lock` and `Cargo.toml` - they are here to include `users` patch that allows passing `auth` token as query string parameter - I'll get rid of these changes for final PR.

src/adapters/ip_camera_adapter/mod.rs
```diff
@@ -261,11 +262,22 @@ impl<T: Controller> Service for IpCameraService<T> {
             "snapshot" => self.cmd_snapshot(),
             "list" => self.cmd_list(),
             "get" => {
-                if let Some(filename) = req.url.query.clone() {
-                    // For now we assume that the query is the filename
-                    return self.cmd_get(&filename);
+                if let Some(query) = req.url.query.as_ref() {
+                    // Creating a fake URL to get the query parsed.
```
azasypkin, line 266: Here I use the same hack as used in [1] to avoid adding custom query string processing code or one more dependency [2].

Please, tell me if you'd rather prefer adding [2] that has dedicated query string parsing capabilities.

Btw, this change is needed since I use this `cmd` directly in `img.src` so query string will include `auth` query string parameter.

[1] https://github.com/fxbox/foxbox/blob/996e84123e464eb69b36b9d960211c7b66150461/src/ws_server.rs#L48
[2] https://crates.io/crates/urlencoded

PR #214, opened 2016-03-15
Use iron-cors. r=fabrice

ferjm:


ferjm:
r? @fabricedesre 

I moved all the CORS related code that we were sharing with the users repo to a [external crate](https://github.com/fxbox/iron-cors). I will also be using it for the registration server.

fabricedesre:
Very nice!

PR #216, opened 2016-03-15
Make the ipcam.py authentication part better

dhylands:
r? @azasypkin 

This is just a change to the ipcam.py script.


azasypkin:
Looks good and works well, r+

PR #219, opened 2016-03-16
Add WebPush support

aosmond:


src/webpush/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+            Ok(Response::with(status::Ok))
+        });*/
+
+        let auth_endpoints = if cfg!(feature = "authentication") {
+            vec![
+                AuthEndpoint(vec![Method::Post], "subscribe".to_owned())
+            ]
+        } else {
+            vec![]
+        };
```
aosmond, line 115: I just realized here that I did not provide any way for the user to query their current subscriptions. I will add that.

PR #220, opened 2016-03-16
Run tests with a separate profile, fixes #213

### LINKED ISSUE #213: Selenium tests need to use a new temporary profile
Right now if you run
`npm test`
the tests pass, but if you run that command again, they will fail.

This is most likely because the tests assume that the database is fresh (ie. with no existing users in it), and after the first run there are users so the default behavior for some pages change.

mikehenrty:


tools/execute-selenium-tests
```diff
@@ -0,0 +1,27 @@
...
+$FOXBOX_BINARY -p ${DEBUG_PORT} --profile ${DEBUG_PROFILE} &
+FOXBOX_PID=$!
+while ! nc -z localhost ${DEBUG_PORT}; do sleep 3; done
+popd
+
+# run tests
+BOX_PORT=${DEBUG_PORT} ${PROJECT_HOME}/node_modules/.bin/mocha ${PROJECT_HOME}/test/selenium/*_test.js
+
+# clean up foxbox process
+kill $FOXBOX_PID
```
hfiguiere, line 32: Can we also delete the temporary profile? Or is there a good reason to not...

hfiguiere:
This should also help for issue #237.

@mikehenrty do you mind rebasing and updating the PR so we can get this reviewed? (it doesn't apply cleanly). I'd love to see this landed.

tools/execute-selenium-tests
mikehenrty, line 32: I was thinking about keeping it, in case someone wanted to run Link against it and do some manual debugging of what happened.

mikehenrty:
I'll pick this up today, thanks Hub

mikehenrty:
@ferjm, @npark-mozilla, review?

src/controller.rs
```diff
@@ -82,9 +82,10 @@ impl FoxBox {
                hostname: Option<String>,
                http_port: u16,
                ws_port: u16,
-               tls_option: TlsOption) -> Self {
-        let profile_service = ProfileService::new(None);
+               tls_option: TlsOption,
+               directory: Option<String>) -> Self {
```
ferjm: Instead of a `Option` we can make `directory` a `enum` like:

``` rust
pub enum ProfilePath {
  Default,
  Custom(String)
}
```

and change https://github.com/fxbox/foxbox/blob/d8ccbfa0da2d7c1a0fb74c38fd481210f3216e98/src/profile_service.rs#L27 to use it as well.

src/main.rs
```diff
@@ -102,13 +102,14 @@ use std::sync::atomic::{ AtomicBool, Ordering, ATOMIC_BOOL_INIT };
...
 docopt!(Args derive Debug, "
-Usage: foxbox [-v] [-h] [-l <hostname>] [-p <port>] [-w <wsport>] [-r <url>] [-i <iface>] [-t <tunnel>] [-s <secret>] [--disable-tls] [--remote-name <hostname>] [-c <namespace;key;value>]...
+Usage: foxbox [-v] [-h] [-l <hostname>] [-p <port>] [-w <wsport>] [-d <profile_dir>] [-r <url>] [-i <iface>] [-t <tunnel>] [-s <secret>] [--disable-tls] [--remote-name <hostname>] [-c <namespace;key;value>]...
 
 Options:
     -v, --verbose            Toggle verbose output.
     -l, --local-name <hostname>    Set local hostname. Linux only. Requires to be a member of the netdev group.
     -p, --port <port>        Set port to listen on for http connections. [default: 3000]
     -w, --wsport <wsport>    Set port to listen on for websocket. [default: 4000]
+    -d, --profile <path>     Set profile path [default: $HOME/.local/share/foxbox]
```
ferjm: Can we use `$XDG_DATA_HOME` instead of another command line option?

Check https://github.com/fxbox/foxbox/blob/d8ccbfa0da2d7c1a0fb74c38fd481210f3216e98/src/profile_service.rs#L28

@fabricedesre wdyt?

mikehenrty:
Whoops, just found a bug. Running Link without specifying a profile folder creates a `$HOME` folder in the cwd. Not good, I'm investigating now.

Edit: Fixed.

src/main.rs
mikehenrty: My thinking here is that `XDG_DATA_HOME` is a special thing, and that someone might want to specify a profile outside of their `XDG_DATA_HOME` at some point.

tools/execute-selenium-tests
hfiguiere, line 32: Fair enough. Or maybe only in case of failure. That can be fixed later if it is too much cluttering.

src/main.rs
hfiguiere: No, you don't want to mess with `$XDG_DATA_HOME`.

nojunpark:
@JohanLorenzo r? as well

hfiguiere:
@fabricedesre we need this PR for #240 r?

tools/execute-selenium-tests
```diff
@@ -0,0 +1,27 @@
...
+CURRENT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_HOME="$CURRENT_PATH/.."
+FOXBOX_BINARY="$PROJECT_HOME/target/debug/foxbox"
+
+DEBUG_HOST=${BOX_LOCAL_NAME:-'localhost'}
+DEBUG_PORT=${BOX_PORT:-3331}
+DEBUG_PROFILE="$HOME/.local/share/foxbox-debug"
+
+# clean up residual profile
+rm -rf ${DEBUG_PROFILE}
```
JohanLorenzo: Nit: Double-quote `$DEBUG_PROFILE`. If somebody has a HOME path that contains a space, this line won't work.

tools/execute-selenium-tests
```diff
@@ -0,0 +1,27 @@
...
+
+# build and launch foxbox daemon
+pushd ${PROJECT_HOME}
+cargo build
+$FOXBOX_BINARY -l ${DEBUG_HOST} -p ${DEBUG_PORT} --disable-tls --profile ${DEBUG_PROFILE} &
```
JohanLorenzo: What is the pro of `cargo build` + call executable, compared to `cargo run`?

tools/execute-selenium-tests
```diff
@@ -0,0 +1,27 @@
...
+FOXBOX_PID=$!
+popd
+
+# run tests
+BOX_PORT=${DEBUG_PORT} ${PROJECT_HOME}/node_modules/.bin/mocha ${PROJECT_HOME}/test/selenium/*_test.js
```
JohanLorenzo: Nit: Double-quotes on `${PROJECT_HOME}/node_modules/.bin/mocha` and `${PROJECT_HOME}/test/selenium/`. The project might be in a folder that contains a space

.travis.yml
```diff
@@ -32,5 +32,5 @@ script:
     - cargo build # Linter is also executed here
     - cargo test
     - jshint static/main/js/*.js static/setup/js/*.js test/selenium/*.js test/integration/lib/*.js test/integration/test/*.js
-    - npm run test-integration # starts foxbox and kills it within the test script
-    - (cargo run -- -l $BOX_LOCAL_NAME -p $BOX_PORT --disable-tls &> /dev/null &) ; npm run test-selenium
+    - npm run test-integration
+    - npm run test-selenium
```
JohanLorenzo, line 36: Nice clean up :)

src/main.rs
```diff
@@ -102,13 +102,14 @@ use std::sync::atomic::{ AtomicBool, Ordering, ATOMIC_BOOL_INIT };
...
 docopt!(Args derive Debug, "
-Usage: foxbox [-v] [-h] [-l <hostname>] [-p <port>] [-w <wsport>] [-r <url>] [-i <iface>] [-t <tunnel>] [-s <secret>] [--disable-tls] [--remote-name <hostname>] [-c <namespace;key;value>]...
+Usage: foxbox [-v] [-h] [-l <hostname>] [-p <port>] [-w <wsport>] [-d <profile_dir>] [-r <url>] [-i <iface>] [-t <tunnel>] [-s <secret>] [--disable-tls] [--remote-name <hostname>] [-c <namespace;key;value>]...
 
 Options:
     -v, --verbose            Toggle verbose output.
     -l, --local-name <hostname>    Set local hostname. Linux only. Requires to be a member of the netdev group.
     -p, --port <port>        Set port to listen on for http connections. [default: 3000]
     -w, --wsport <wsport>    Set port to listen on for websocket. [default: 4000]
+    -d, --profile <path>     Set profile path
```
JohanLorenzo: If I didn't see the PR, I wouldn't be 100% sure of what a profile represents. Maybe we could make the message more explicit? What do you think of `Set profile path to store user's data`?

fabricedesre: Please switch to -P instead of -d

src/controller.rs
fabricedesre: Yep, good idea.

fabricedesre:
Fine with me if you address all the comments from various people ;)

tools/execute-selenium-tests
npark-mozilla: Strange, so when I execute the foxbox, it dies with the error 
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Custom(Custom { kind: Other, error: StringError("failed to lookup address information: nodename nor servname provided, or not known") }) }', ../src/libcore/result.rs:746

This was fixed when I deleted the -l localhost parameter and issued following command instead:
/Users/g5njpark/GitRepo/foxbox/tools/../target/debug/foxbox -p 3331 --disable-tls --profile /Users/g5njpark/.local/share/foxbox-debug

Ah, so -l is for linux only.  I guess it's good for the travis since it's on Ubuntu, but would there be a quick way to check for the environment in ./execute-selenium-tests?  if so we can also use npm run selenium-tests on mac.

nojunpark:
Strange, so when I execute the foxbox, it dies with the error 
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Custom(Custom { kind: Other, error: StringError("failed to lookup address information: nodename nor servname provided, or not known") }) }', ../src/libcore/result.rs:746

This was fixed when I deleted the -l localhost parameter and issued following command instead:
/Users/g5njpark/GitRepo/foxbox/tools/../target/debug/foxbox -p 3331 --disable-tls --profile /Users/g5njpark/.local/share/foxbox-debug

Ah, so -l is for linux only.  I guess it's good for the travis since it's on Ubuntu, but would there be a quick way to check for the environment in ./execute-selenium-tests?  if so we can also use npm run selenium-tests on mac.

Aside from this nit, r+ for me.

hfiguiere:
@npark-mozilla the ````--profile```` isn't about the Firefox profile at all.

hfiguiere:
(deleted my last comment, please ignore the email)

tools/execute-selenium-tests
mikehenrty: Since `cargo run` first builds and then forks the foxbox daemon process, it is harder to get the PID of the actual foxbox process (so we can kill it later). We also then have to explicitly wait for the daemon to spin up (using something like netcat) before continuing to the next line. Doing it this way was just cleaner.

tools/execute-selenium-tests
mikehenrty: @JohanLorenzo doesn't the double quotes on the definition of DEBUG_PROFILE above cover this?

mikehenrty:
@fabricedesre or @ferjm, r?

src/controller.rs
```diff
@@ -289,9 +290,12 @@ describe! controller {
     before_each {
         use stubs::service::ServiceStub;
         use tls::TlsOption;
+        use profile_service::ProfilePath;
 
         let service = ServiceStub;
-        let controller = FoxBox::new(false, Some("foxbox".to_owned()), 1234, 5678, TlsOption::Disabled);
+        let controller = FoxBox::new(
+          false, Some("foxbox".to_owned()), 1234, 5678,
+                      TlsOption::Disabled, ProfilePath::Default);
```
jedireza, line 298: nit: line up args

``` rust
let controller = FoxBox::new(false,
                             Some("foxbox".to_owned()),
                             1234,
                             5678,
                             TlsOption::Disabled,
                             ProfilePath::Default);
```

src/controller.rs
```diff
@@ -316,7 +320,9 @@ describe! controller {
             assert_eq!(controller.get_http_root_for_service("1".to_string()),
                        "http://foxbox.local:1234/services/1/");
 
-            let controller = FoxBox::new(false, Some("foxbox".to_owned()), 1234, 5678, TlsOption::Enabled);
+            let controller = FoxBox::new(false, Some("foxbox".to_owned()),
+                                         1234, 5678, TlsOption::Enabled,
+                                         ProfilePath::Default);
```
jedireza, line 325: nit: i like one arg per line when they're broken up

src/service_router.rs
```diff
@@ -69,8 +69,11 @@ describe! service_router {
         use iron_test::request;
         use mount::Mount;
         use tls::TlsOption;
+        use profile_service::ProfilePath;
 
-        let controller = FoxBox::new(false, Some("localhost".to_owned()), 1234, 5678, TlsOption::Disabled);
+        let controller = FoxBox::new(false, Some("localhost".to_owned()),
+                                     1234, 5678, TlsOption::Disabled,
+                                     ProfilePath::Default);
```
jedireza, line 76: nit: i like one arg per line when they're broken up

src/profile_service.rs
```diff
@@ -104,7 +110,8 @@ fn test_custom_profile() {
     let profile_path = String::from(profile_dir.into_path()
                                     .to_str().unwrap());
 
-    let profile = ProfileService::new(Some(profile_path.to_owned()));
+    let profile = ProfileService::new(ProfilePath::Custom(profile_path
+                                                          .clone()));
```
jedireza, line 114: nit: since we're using another line, can we make it more readable? maybe making a variable out of one of those inner values?

``` rust
let custom_path = ProfilePath::Custom(profile_path.clone());
let profile = ProfileService::new(cloned_path);
```

src/controller.rs
mikehenrty, line 298: Hmmm, this isn't the convention I've seen. Do you guys have a style guide somewhere?

src/profile_service.rs
mikehenrty, line 114: The reason we are doing the clone is because the ownership of this string gets passed to the profile service, and we need to do a comparison with this string in our assert statement here below. So doing a clone before assigning the string to a variable runs into the same ownership problem this clone is meant to address.

mikehenrty, line 114: Oh I see what you mean. That would work too. Although the convention seems to be not assigning state variables for this ownership management stuff. A style guide would help here, but I guess you guys are still figuring that out.

src/controller.rs
jedireza, line 298: I see it in Servo
https://github.com/servo/webrender/blob/2a71ccb0249601b25ef156f9a585a001c8d606cb/src/render_backend.rs#L152-L158

PR #222, opened 2016-03-16
Upgrade foxbox_users to use bcrypt() for password hashing.

hfiguiere:


hfiguiere:
@fabricedesre r?

hfiguiere:
Apparently the selenium tests are failing. But I have never been able to run them locally :-/

hfiguiere:
OK I can now reproduce the failure :-/ Will update the PR with a fix.

hfiguiere:
If I do the test manually it works. :-(

hfiguiere:
Now the failure is indeed the systemic intermittent from bug #237

hfiguiere:
@fabricedesre @ferjm @JohanLorenzo  r? (Johan for the Selenium tests)

test/selenium/lib/view/successful_page/accessors.js
```diff
@@ -12,7 +12,14 @@ function SuccessfulPageAccessor(driver) {
...
 SuccessfulPageAccessor.prototype = {
    get successMessageLocator() {
-        return this.driver.findElement(SELECTORS.successMessage);
+        var driver = this.driver;
+        return driver.wait(
+            webdriver.until.elementLocated(SELECTORS.successMessage))
+            .then(function (element) {
+                return driver.wait(webdriver.until.elementIsVisible(element));
+            }).then(function () {
+                return driver.findElement(SELECTORS.successMessage);
```
JohanLorenzo: Nit: This call is not needed `driver.wait(webdriver.until.elementIsVisible(element))` already returns the element.

test/selenium/lib/view/successful_page/accessors.js
```diff
@@ -12,7 +12,14 @@ function SuccessfulPageAccessor(driver) {
 
 SuccessfulPageAccessor.prototype = {
    get successMessageLocator() {
-        return this.driver.findElement(SELECTORS.successMessage);
+        var driver = this.driver;
```
JohanLorenzo: Nit: This line is not necessary. Arrow functions can be used with this version of node.

.travis.yml
```diff
@@ -30,7 +30,7 @@ before_script:
 
 script:
     - cargo build # Linter is also executed here
-    - cargo test
+    - RUST_TEST_THREADS=1 cargo test
```
JohanLorenzo: What do you think of adding a reference to #237?

hfiguiere: The commit does. Also I'm working on fixing it.

test/selenium/lib/view/successful_page/accessors.js
hfiguiere: Which version is that?

hfiguiere: ah right, elsewhere it does. Will use arrows.

JohanLorenzo: We currently [support 4.2](https://github.com/fxbox/foxbox/pull/222/files/fd5531433c1ccdf70e862dfcf602f584301b9cb5#diff-354f30a63fb0907d4ad57269548329e3R28)

hfiguiere: Old habits. I didn't realise Node 4 had support for arrows now. Since I'm using it now. PR has been updated to use arrows.

test/selenium/lib/view/successful_page/accessors.js
hfiguiere: It doesn't. It returns a Promise. If I don't have the findElement() it doesn't work.

hfiguiere:
Last be not least. @fabricedesre r?

This will probably require people to reset the passwords. Better do that now than later.

fabricedesre:
the test failure looks real. Can you check that?

hfiguiere:
Looks like a continuation of the unreliable selenium tests.

hfiguiere:
Apparently not. Since they have passed... Not sure exactly what has happened.

Issue #230, opened 2016-03-17
README instruction to run selenium test don't work.


hfiguiere:
README instructions to run selenium test don't work.
1.  npm test doesn't work

```
$ npm test

> foxbox@1.0.0 test /home/hub/source/mozilla/foxbox
> echo "Error: no test specified"

Error: no test specified
```

Travis uses `npm run test-selenium`
1. Even then it doesn't say to start foxbox either.

PR #205, opened 2016-03-14
Add integration test into travis

nojunpark:

PR #231, opened 2016-03-17
Issue #230 - Fix documentation about tests.

hfiguiere:
@JohanLorenzo feedback? I had to second guess how to run them.


JohanLorenzo:
Good catch! `npm test` was changed in #205 and we forgot to update it. Thanks for fixing it.

f+

PR #232, opened 2016-03-17
Fix the test hostname detection to not add .local. if it is localhost

hfiguiere:
If you set BOX_LOCAL_NAME to localhost test won't work. This is was in the previous documentation.
Also add a final `.` to `.local.` because we don't want any domain name happened if there is a failure with the mDNS registration.

### DIFF
--- a/test/selenium/sessions_ui_test.js
+++ b/test/selenium/sessions_ui_test.js
@@ -9,8 +9,8 @@ var webdriver = require('selenium-webdriver'),
 
 var HOST_URL = util.format(
     'http://%s:%s/',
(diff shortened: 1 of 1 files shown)


hfiguiere:
@JohanLorenzo r?

hfiguiere:
Let's wait on the test to pass :-/ Looks like this breaks here. Can't test locally.

hfiguiere:
@JohanLorenzo r? (this time it passes - figured out how to run locally with mDNS registration)

JohanLorenzo:
Looks good to me.

PR #235, opened 2016-03-17
Add integration tests that check pagekite site, and ping endpoint

nojunpark:


nojunpark:
In this PR, I have added:
- a test that checks the ping endpoint of the pagekite site, and verifies by using the values from it.
- a test that turns on philips hue lights remotely by using pagekite site
- cleanup of servers so the tests can run in fast succession.
  @JohanLorenzo , r?

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+var helper = (function() {
+
+    function removeDB() {
+      var fs = require('fs');
```
JohanLorenzo: Nit: Could be put at the top of the file

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,19 @@
...
+      var filePath = process.env.HOME +
```
JohanLorenzo: Nit: [path.join()](https://nodejs.org/api/path.html#path_path_join_path1_path2) could be used. It normalizes the path to be cross-platform.

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,19 @@
...
+
+var helper = (function() {
+
+    function removeDB() {
+      var fs = require('fs');
+      var filePath = process.env.HOME + 
+      '/.local/share/foxbox/users_db.sqlite' ; 
+      fs.exists(filePath,function(exists){
+        if(exists) {
+          fs.unlinkSync(filePath);
```
JohanLorenzo: Using a sync method is usually a bad practice in Node, as it block the main (and only) thread. I think we can return a promise to make `removeDb` async.

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+var helper = (function() {
+
+    function removeDB() {
```
JohanLorenzo: Nit: Maybe we could precise which database is deleted? `removeUsersDatabase`?

test/integration/lib/nupnp_PhilipsHue.js
```diff
@@ -3,19 +3,25 @@
...
       
-      app.listen(port, function () {
+      instance = _app.listen(port, function () {
         console.log('Philips nupnp app listening on port ' + port);
       });
     }
-  };
+
+    function stop() {
+      instance.close();
```
JohanLorenzo: `close()` is [an async method](https://nodejs.org/api/http.html#http_server_close_callback).

test/integration/lib/philipsHue_server.js
```diff
@@ -107,12 +108,16 @@ var philipsHue_server = (function() {
...
 
-      _server.listen(port, function () {
+      instance = _server.listen(port, function () {
         console.log('Hue simulator listening on port ' + port);
       });
     }
 
-  return {setup, lastCmd, lightResponse, turnOnLight, 
+    function stop(done){
+      instance.close();
```
JohanLorenzo: Same comment about async

test/integration/test/philips_light_test.js
```diff
@@ -4,59 +4,71 @@ const spawn = require('child_process').spawn;
...
+  var loginURL = config.get('foxbox.url') + '/users/login';
+  var serviceURL = config.get('foxbox.url') + '/services/';
+  var serviceListURL = serviceURL + 'list';  
+  
   before(function(done){
+    var FOXBOX_STARTUP_WAIT_TIME_IN_MS = 3000;
     this.timeout(500000);
 
-    const hue_location = 'localhost:' + config.get('philips_hue.port');
+    // start nupnp server
```

test/integration/test/pagekite_ping_test.js
```diff
@@ -0,0 +1,94 @@
...
+    // start nupnp server
+    nupnp_server.start(config.get('nupnp_server.id'),
+      config.get('philips_hue.url') +':'+config.get('philips_hue.port'),
+      config.get('nupnp_server.port'));
+    
+    // configure philips hub simulator
+    philipshue_server.setup(config.get('philips_hue.port'));
+
+    // start foxbox
+    foxbox_process = spawn('./target/debug/foxbox',
```
JohanLorenzo: I think the tests start to know too much details. We can extract the logic of how a process should be started in a separated module. Something like:

``` js
FoxboxManager.prototype = {
  shutdown: function() { // implementation },
  standaloneStart: function() {  // implementation },
  fullOptionStart: function() {
   return new Promise((resolve, reject) {
     this.process = spawn('./target/debug/foxbox', 
       ['-c',  config.get('nupnp_server.param')+';'+ 
       config.get('nupnp_server.url')+':'+ 
       config.get('nupnp_server.port')+'/',
       '-r',config.get('pagekite.r'),
       '-t',config.get('pagekite.t'),
       '-s',config.get('pagekite.s'),
       '--remote-name',config.get('pagekite.externalURL'),
       '--disable-tls'] ); // TODO TLS not yet supported

     setTimeout(resolve, FOXBOX_STARTUP_WAIT_TIME_IN_MS);
   });
  },
}
```

and then in the test:

``` js
before(() => {
  // some other steps
  return foxboxManager.fullOptionStart();
});
```

What do you think?

test/integration/test/pagekite_ping_test.js
```diff
@@ -0,0 +1,94 @@
...
+      '--remote-name',config.get('pagekite.externalURL'),
+      '--disable-tls'] );
+
+    // give time until foxbox is operational
+    setTimeout(done, FOXBOX_STARTUP_WAIT_TIME_IN_MS);  
+  });
+  
+  after(function(){
+    foxbox_process.kill('SIGINT');
+    require('../lib/foxboxHelper.js').removeDB();
```
JohanLorenzo: Maybe this could also be incorporated in [FoxboxManager](https://github.com/fxbox/foxbox/pull/235/files#r56685935)

test/integration/test/pagekite_access_test.js
```diff
@@ -0,0 +1,142 @@
...
+  var serviceURL = 'http://' + config.get('pagekite.externalURL')+ '/services/';
+  var serviceListURL = serviceURL + 'list';  
+  
+  before(function(done){
+
+    var FOXBOX_STARTUP_WAIT_TIME_IN_MS = 3000;
+    this.timeout(500000);
+
+    // start nupnp server
+    nupnp_server.start(config.get('nupnp_server.id'),
```
JohanLorenzo: As the nupnp_server is started the same way in the 3 tests, I'd put the default configuration in that module.

test/integration/test/pagekite_access_test.js
```diff
@@ -0,0 +1,142 @@
...
+      config.get('philips_hue.url') +':'+config.get('philips_hue.port'),
+      config.get('nupnp_server.port'));
+    
+    // configure philips hub simulator
+    philipshue_server.setup(config.get('philips_hue.port'));
```
JohanLorenzo: Same comment as the nupnp_server just above

test/integration/test/pagekite_access_test.js
```diff
@@ -0,0 +1,142 @@
...
+  it('login to foxbox remotely',function(){
+    var encoded_cred = new Buffer(credential.username+
+      ':'+credential.password).toString('base64');
+
+    // supply the credential used in previous test
+    header.Authorization = 'Basic ' + encoded_cred;  
+    
+    return  chakram.post(loginURL,null,{'headers' : header})
+    .then(function(loginResp){
+      header.Authorization = 'Bearer ' + loginResp.body.session_token;
```
JohanLorenzo: I believe this line can be changed to something like:

``` js
expect(loginResp.body.session_token).to.be.a('string');
```

test/integration/test/pagekite_access_test.js
```diff
@@ -0,0 +1,142 @@
...
+  // TODO: a new test should be added once it is implemented
+  it('Create account locally',function(){
+
+    return chakram.post(setupURL, credential)
+    .then(function(setupResp){
+      expect(setupResp).to.have.status(201);
+    });
+  });
+
+  it('login to foxbox remotely',function(){
```
JohanLorenzo, line 28: I just realized the tests are similar to the first one uploaded. Maybe we could factorize them with something similar to:
http://stackoverflow.com/questions/29312331/how-do-i-factor-out-common-before-calls-with-mocha

test/integration/test/pagekite_ping_test.js
```diff
@@ -0,0 +1,94 @@
...
+    var pingUrl = config.get('pagekite.r') + '/ping';
+    
+    return  chakram.get(pingUrl,{'json': true})
+    .then(function(pingResp){
+      local_ip = pingResp.body[0].local_ip;
+      tunnel_url = pingResp.body[0].tunnel_url;
+      
+      expect(pingResp).to.have.status(200);
+      expect(local_ip).is.not.equal(undefined);
+      expect(tunnel_url).is.not.equal(undefined);
```
JohanLorenzo: I think these 2 assertions could be smarter. We can use regexp: http://chaijs.com/api/bdd#matchregexp

test/integration/test/philips_light_test.js
```diff
@@ -4,59 +4,71 @@ const spawn = require('child_process').spawn;
...
 
-describe('Initiate the connection with foxbox as Philips Hue hub',function(){
-  var credential = config.get('credential');
-  var FOXBOX_STARTUP_WAIT_TIME_IN_MS = 3000;
+describe('Initiate the local connection',function(){
+  var credential = config.get('credential'); 
   var lightinfo; // to store the light status received from foxbox
   var foxbox_process;
 
+  var setupURL = config.get('foxbox.url') + '/users/setup';
```
JohanLorenzo, line 14: I wonder if we shouldn't put the URLs in FoxboxManager too

JohanLorenzo:
~~r? @JohanLorenzo~~
Nice to see some tests done through the bridge! I think we could make some improvements in regards to sync/async. That might be a cause of intermittent-cy

test/integration/lib/foxboxHelper.js
nojunpark: I thought about it, but actually, the main needs to be blocked in order to proceed to the next test.  IF the db is not removed before moving on to the next test, the test would fail.  Also, this is at the end of the test, so nothing other than the upcoming test would be blocked by it.

test/integration/test/pagekite_access_test.js
nojunpark, line 28: This is great, I was looking for something like this, thanks!

package.json
```diff
@@ -34,5 +34,8 @@
     "path": "^0.12.7",
     "config-js": "^1.1.9",
     "express": "^4.13.4"
+  },
+  "dependencies": {
+    "promise": "^7.1.1"
```
JohanLorenzo: This dependency doesn't seem to be used in the PR. I don't see anywhere it's required.

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,57 @@
...
+       config.get('nupnp_server.url')+':'+ 
+       config.get('nupnp_server.port')+'/',
+       '-r',config.get('pagekite.r'),
+       '-t',config.get('pagekite.t'),
+       '-s',config.get('pagekite.s'),
+       '--remote-name',config.get('pagekite.externalURL'),
+       '--disable-tls'], {stdio: 'inherit'} ); // TODO TLS not yet supported
+      setTimeout(function(){resolve();}, FOXBOX_STARTUP_WAIT_TIME_IN_MS);
+    });
+     start.then(function(val) {return callback();});
```
JohanLorenzo: What's the advantage of using a callback in this case?

test/integration/lib/philipsHue_server.js
```diff
@@ -107,14 +108,21 @@ var philipsHue_server = (function() {
...
-    }
+    instance = _server.listen(port, function () {
+      console.log('Hue simulator listening on port ' + port);
+    });
+  }
+
+  function stop(callback) {
+    instance.close(function(){
+      console.log('philips hue server closed');
+      return callback;
```
JohanLorenzo: I'm not sure what this statement is supposed to do. Could you explain it to me?

test/integration/lib/philipsHue_server.js
```diff
@@ -107,14 +108,21 @@ var philipsHue_server = (function() {
...
-  return {setup, lastCmd, lightResponse, turnOnLight, 
+  return {setup, stop, lastCmd, lightResponse, turnOnLight, 
     turnOffLight, lightStatus};
 
-})();
+  })();
 
-module.exports = philipsHue_server;
\ No newline at end of file
+  module.exports = philipsHue_server;
```
JohanLorenzo: Nit: Unnecessary indentation

test/integration/lib/testPrepper.js
```diff
@@ -0,0 +1,54 @@
...
+  function beforeTest(done) { 
+    // start nupnp server
+    nupnp_server.start(config.get('nupnp_server.id'),
+      config.get('philips_hue.url') +':'+config.get('philips_hue.port'),
+      config.get('nupnp_server.port'));
+    
+    // configure philips hub simulator
+    philipshue_server.setup(config.get('philips_hue.port'));
+
+    // start foxbox
```
JohanLorenzo: Nit: Unnecessary comment

test/integration/lib/testPrepper.js
```diff
@@ -0,0 +1,54 @@
...
+
+var testPrepper = (function() {
+
+  function beforeTest(done) { 
+    // start nupnp server
+    nupnp_server.start(config.get('nupnp_server.id'),
+      config.get('philips_hue.url') +':'+config.get('philips_hue.port'),
+      config.get('nupnp_server.port'));
+    
+    // configure philips hub simulator
```
JohanLorenzo: Nit: Unnecessary comment

test/integration/lib/testPrepper.js
```diff
@@ -0,0 +1,54 @@
...
+var config = new Config('./test/integration/lib/config/foxbox.js');
+var philipshue_server = require('../lib/philipsHue_server.js');
+var nupnp_server = require('../lib/nupnp_PhilipsHue.js');
+var foxboxManager = require('../lib/foxboxHelper.js');
+var config = new Config('./test/integration/lib/config/foxbox.js');
+
+var testPrepper = (function() {
+
+  function beforeTest(done) { 
+    // start nupnp server
```
JohanLorenzo: Nit: Unnecessary comment

test/integration/test/pagekite_access_test.js
```diff
@@ -0,0 +1,104 @@
...
+      });
+
+      // Currently, there is no mapping between the foxbox 
+      // issues id and the philips hue id until the tag feature is implemented
+      it('remotely Turn on all lights', function(){
+
+        return chakram.put(serviceURL + lightinfo[0].id + 
+          '/state', {'on': true})
+        .then(function(cmdResponse) {
+         expect(cmdResponse).to.have.status(200);
```
JohanLorenzo: I think we can extract these 3 expectations and the command in functions. This would make the test easier to read, I think. 

``` js
function expectLightIsOn(response) {
   expect(response).to.have.status(200);
   expect(response.body.result).equals('success');
   expect(Prepper.philipshue_server.lastCmd()).to.contain('"on":true');
}

function turnOnLight(index) {
  return chakram.put(serviceURL + lightinfo[2].id + '/state', {'on': true});})
}
```

``` js
.then(function(cmdResponse) {
  expectLightIsOn(cmdResponse);
  return turnOnLight(1);
})
.then(function(cmdResponse) {
  expectLightIsOn(cmdResponse);
  return turnOnLight(2);
})
```

test/integration/test/pagekite_access_test.js
```diff
@@ -0,0 +1,104 @@
...
+         expect(cmdResponse.body.result).equals( 'success');
+         expect(Prepper.philipshue_server.lastCmd()).to.contain('"on":true');
+         
+         // check no lights are turned off now
+         expect(Prepper.philipshue_server.lightStatus(1) && 
+          Prepper.philipshue_server.lightStatus(2) && 
+          Prepper.philipshue_server.lightStatus(3)).equals(true);
+       });
+      });
+});
```
JohanLorenzo: Nit: Bad indentation

test/integration/lib/foxboxHelper.js
nojunpark: so callback has to be in this location.  if done() is called within above promise, it executes instantly, if there is a way around it, please let me know.

test/integration/lib/philipsHue_server.js
nojunpark: so this releases the port.  otherwise, on the next test run, the test will fail because the port is still occupied. I have to make sure that the listener instance is closed, by waiting on the callback

test/integration/test/pagekite_access_test.js
nojunpark: we can do expectLightIsOn, but we can't to turonlight yet, because we don't know which light we have turned it on at this point ,without the use of tags.  this could be added later when the tag feature is complete.
EDIT: ah wait, my mistake, will do this.

test/integration/lib/testPrepper.js
```diff
@@ -9,46 +10,47 @@ var config = new Config('./test/integration/lib/config/foxbox.js');
...
-    };
-    nupnp_server.stop(getKill());
-    philipshue_server.stop(getKill());
-    foxboxManager.removeUsersDB();
+    
     foxboxManager.killFoxBox();
+    nupnp_server.stop();
+    philipshue_server.stop();
+    foxboxManager.removeUsersDB();
+    console.log("aftertest done");
```
nojunpark: This is where I'm having a problem.  This console.log gets executed after the console.log in before() (line 25 of this file).  How can I make sure that it waits sufficiently? @JohanLorenzo 

EDIT: this seems to be fixed...

nojunpark:
hmm, i removed all callbacks and it seems to be working reliably... could you check whether it runs on your machine okay as well?  @JohanLorenzo

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,57 @@
...
+var config = new Config('./test/integration/lib/config/foxbox.js');
+var FOXBOX_STARTUP_WAIT_TIME_IN_MS = 3000;
+var foxboxInstance;
+
+var helper = (function() {
+
+  function removeUsersDB() {
+    var filePath = path.join(process.env.HOME, 
+      '/.local/share/foxbox/users_db.sqlite'); 
+    fs.unlinkSync(filePath);
```
JohanLorenzo, line 17: _This can be done in a follow up_

I got this error (because of a bug in PageKite -- works fine when I fixed it):

```
  3) Verify validity of pagekite ping endpoint "after all" hook: afterTest:
     Error: ENOENT: no such file or directory, unlink '/home/jlorenzo/.local/share/foxbox/users_db.sqlite'
      at Error (native)
      at Object.fs.unlinkSync (fs.js:932:18)
      at Object.removeUsersDB (test/integration/lib/foxboxHelper.js:16:8)
      at Context.afterTest (test/integration/lib/testPrepper.js:41:19)
```

As `after all` should be responsible for the end state, I don't think we should return an error if no file was initially present. 

I'd recommend to mute the exception in `after()`

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,56 @@
...
+       config.get('nupnp_server.url')+':'+ 
+       config.get('nupnp_server.port')+'/',
+       '-r',config.get('pagekite.r'),
+       '-t',config.get('pagekite.t'),
+       '-s',config.get('pagekite.s'),
+       '--remote-name',config.get('pagekite.externalURL'),
+       '--disable-tls'], {stdio: 'inherit'} ); // TODO TLS not yet supported
+      setTimeout(function(){resolve();}, FOXBOX_STARTUP_WAIT_TIME_IN_MS);
+    });
+    start.then(function(val) {return callback();});
```
JohanLorenzo: Nit: With a new look, a Promise is actually not really necessary. You could do something like:

``` js
function fullOptionStart(callback) {
  foxboxInstance = spawn('./target/debug/foxbox', 
    ['-c',  config.get('nupnp_server.param')+';'+ 
    config.get('nupnp_server.url')+':'+ 
    config.get('nupnp_server.port')+'/',
    '-r',config.get('pagekite.r'),
    '-t',config.get('pagekite.t'),
    '-s',config.get('pagekite.s'),
    '--remote-name',config.get('pagekite.externalURL'),
    '--disable-tls'], {stdio: 'inherit'} ); // TODO TLS not yet supported
  setTimeout(callback, FOXBOX_STARTUP_WAIT_TIME_IN_MS);
}
```

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,56 @@
...
+  function localStart(callback) {
+    var start = new Promise(function (resolve, reject) {
+      foxboxInstance = spawn('./target/debug/foxbox', 
+        ['-c',  config.get('nupnp_server.param')+';'+ 
+        config.get('nupnp_server.url')+':'+ 
+        config.get('nupnp_server.port')+'/',
+        '--disable-tls'] );
+      setTimeout(function(){resolve();}, FOXBOX_STARTUP_WAIT_TIME_IN_MS);
+    });
+    start.then(function(val) {return callback();});
```
JohanLorenzo: Nit: Same comment as [this one](https://github.com/fxbox/foxbox/pull/235/files#r57878878).

test/integration/lib/foxboxHelper.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  function killFoxBox() {
+    foxboxInstance.kill('SIGINT');
+  }
+
+return {removeUsersDB, fullOptionStart, localStart, killFoxBox};
```
JohanLorenzo: Nit: Bad indentation

test/integration/lib/nupnp_PhilipsHue.js
```diff
@@ -1,21 +1,28 @@
...
       
-      app.listen(port, function () {
+      instance = _app.listen(port, function () {
         console.log('Philips nupnp app listening on port ' + port);
+      });     
+    }
+
+    function stop() {
+      instance.close(function(){
+        console.log('nupnp server closed');
```
JohanLorenzo: Nit: Use a callback (or a Promise):

``` js
function stop(callback) {
  instance.close(function(){
    console.log('nupnp server closed');
    callback();
  });
}
```

test/integration/lib/philipsHue_server.js
```diff
@@ -107,14 +108,20 @@ var philipsHue_server = (function() {
...
-      _server.listen(port, function () {
-        console.log('Hue simulator listening on port ' + port);
-      });
-    }
+    instance = _server.listen(port, function () {
+      console.log('Hue simulator listening on port ' + port);
+    });
+  }
+
+  function stop() {
```
JohanLorenzo, line 116: Nit: Same comment as [this one](https://github.com/fxbox/foxbox/pull/235/files#r57879230)

JohanLorenzo:
r+ . These first tests for pagekite look good to me modulo a few nits. Thanks for the PR! :+1:

test/integration/lib/foxboxHelper.js
nojunpark: turns out localstart is not being used at the moment, i'll remove it for now.

PR #239, opened 2016-03-18
Take a fingerprint of the cert and store along with the CertificateRecord r=michielbdejong

samgiles:
This is to be used to send to the DNS API - which I've began to outline in this PR.


michielbdejong:
As discussed on irc, please add a unit test that checks that the fingerprint calculated is correct, given a cert in a file on disk. You could reuse https://github.com/fxbox/dns-server/blob/master/test/dns-query_test.js#L37, then we know it's compatible with the way the DNS server calculates the fingerprint.

Otherwise, r+!

samgiles:
Added suggested test.

PR #240, opened 2016-03-18
Ensure test are run with their own independent profile.

hfiguiere:
On top of PR #222 and PR #220 

@fabricedesre r?


hfiguiere:
If we land #220 we can land this and I'll rebase #222

hfiguiere:
The second last commit is not required but nice to have. I needed to validate it was working.

hfiguiere:
This is on top of PR #220 only. Nothing else.

hfiguiere:
Rebased now. @fabricedesre r?

PR #241, opened 2016-03-18
Version bump

Yoric:


Yoric:
Fixed a (de)serialization error in Taxonomy, bumped version.

PR #244, opened 2016-03-21
Implement WebPush as an adapter using new taxonomy

aosmond:


Yoric:
\o/ Has it reached a stage at which we can test it?

Cargo.lock
```diff
@@ -220,7 +221,7 @@ source = "git+https://github.com/fxbox/users.git?rev=3db32f0#3db32f0057bc70c78e8
 dependencies = [
  "iron 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "jwt 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
```
fabricedesre: hm, what's leading to this downgrade?

aosmond: Hm, maybe a bad merge from master. I'll revert that.

src/adapters/webpush/db.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+pub struct WebPushDb {
+    db: Connection,
+}
+
+impl WebPushDb {
+    pub fn new(path: &str) -> Self {
+        let db = Connection::open(path).unwrap();
+        db.execute("CREATE TABLE IF NOT EXISTS subscriptions (
```
fabricedesre, line 38: Please add comments to explain what the tables are used for.

src/adapters/webpush/db.rs
```diff
@@ -0,0 +1,215 @@
...
+                    user_id     INTEGER,
+                    resource    TEXT NOT NULL
+            )", &[]).unwrap();
+
+        WebPushDb {
+            db: db
+        }
+    }
+
+    pub fn subscribe(&self, user_id: i32, sub: &Subscription) -> rusqlite::Result<c_int> {
```
fabricedesre, line 55: nit: We should have a custom type for user_id, but that can be done in a followup.

src/adapters/webpush/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+use std::sync::Arc;
+use std::thread;
+use transformable_channels::mpsc::*;
+
+header! { (Encryption, "Encryption") => [String] }
+header! { (EncryptionKey, "Encryption-Key") => [String] }
+
+static ADAPTER_NAME: &'static str = "WebPush adapter (built-in)";
+static ADAPTER_VENDOR: &'static str = "team@link.mozilla.org";
+static ADAPTER_VERSION: [u32;4] = [0, 0, 0, 0];
```
fabricedesre, line 44: We need to decide what we expect from version numbers...

src/adapters/webpush/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+        let enc = match self::crypto::encrypt(&self.public_key, message.clone()) {
+            Some(x) => x,
+            None => {
+                warn!("notity subscription {} failed for {}", self.push_uri, message);
+                return;
+            }
+        };
+
+        let client = Client::new();
+        let res = match client.post(&self.push_uri)
```
fabricedesre, line 96: This is a blocking call. Since this method doesn't return anything, can you move that to a short lived thread?

src/adapters/webpush/mod.rs
Yoric, line 44: For the moment, version numbers are entirely unused and entirely up to renegotiation.

src/adapters/webpush/mod.rs
aosmond, line 96: Hm. I wonder what is best to do here. I spawn a thread (see WebPush::notify) which sequentially executes all of the push notifications. My thinking was if we have a lot of subscriptions active, we would want to avoid hammering the push service by spawning N threads all at once. On the other hand, if one user uses another push service and that one hangs, it would block all of the others from completing...

src/adapters/webpush/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+        set.drain(..).map(|id| {
+            let user_id = 1; // FIXME: currently logged in user
+
+            macro_rules! getter_api {
+                ($getter:ident, $getter_id:ident, $getter_type:ident) => (
+                    if id == self.$getter_id {
+                        match self.$getter(user_id) {
+                            Ok(data) => {
+                                let rsp = $getter_type::new(data);
+                                return (id, Ok(Some(Value::Json(Arc::new(Json(serde_json::to_value(&rsp)))))));
```
fabricedesre, line 186: why isn't Ok(Some(serde_json::to_value(&resp))) enough?

src/adapters/webpush/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+
+            setter_api!(update_resources, setter_resource_id, ResourceGetter);
+            setter_api!(subscribe, setter_subscribe_id, SubscriptionGetter);
+            setter_api!(unsubscribe, setter_unsubscribe_id, SubscriptionGetter);
+            setter_api!(notify, setter_notify_id, NotifySetter);
+            (id.clone(), Err(Error::InternalError(InternalError::NoSuchSetter(id))))
+        }).collect()
+    }
+
+    fn register_watch(&self, mut watch: Vec<(Id<Getter>, Option<Range>)>,
```
fabricedesre, line 233: nit for myself: we should make that be the default implementation.

src/controller.rs
```diff
@@ -99,7 +99,7 @@ impl FoxBox {
             config: Arc::new(ConfigService::new(&profile_service.path_for("foxbox.conf"))),
             upnp: Arc::new(UpnpManager::new()),
             users_manager: Arc::new(UsersManager::new(&profile_service.path_for("users_db.sqlite"))),
-            profile_service: Arc::new(profile_service)
+            profile_service: Arc::new(profile_service),
```
fabricedesre: nit: sure, but why? ;)

src/adapters/webpush/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+                                return (id, Ok(Some(Value::Json(Arc::new(Json(serde_json::to_value(&rsp)))))));
+                            },
+                            Err(_) => return (id, Err(Error::InternalError(InternalError::InvalidInitialService)))
+                        };
+                    }
+                ) 
+            }
+
+            getter_api!(user_subscriptions, getter_subscription_id, SubscriptionGetter);
+            getter_api!(resources, getter_resource_id, ResourceGetter);
```
fabricedesre: that may be a bit clearer to name these get_user_subscriptions and get_resources

src/adapters/webpush/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+                            Err(_) => return (id, Err(Error::InternalError(InternalError::InvalidInitialService)))
+                        }
+                    }
+                ) 
+            }
+
+            setter_api!(update_resources, setter_resource_id, ResourceGetter);
+            setter_api!(subscribe, setter_subscribe_id, SubscriptionGetter);
+            setter_api!(unsubscribe, setter_unsubscribe_id, SubscriptionGetter);
+            setter_api!(notify, setter_notify_id, NotifySetter);
```
fabricedesre: here also, set_XXX

fabricedesre:
That looks great overall! \o/

I'd mostly like to see more comments.

Cargo.toml
```diff
@@ -28,9 +28,11 @@ mount = "0.0.10"
 nix = { git = "https://github.com/nix-rust/nix.git", rev = "138080" } # Until 0.5.1 is released
 openssl = "0.7.6"
 openssl-sys = "0.7.6"
+rand = "0.3"
 router = "0.1.0"
 rust-crypto = "0.2.34"
 rustc-serialize = "0.3"
+rusqlite = "*"
```
fabricedesre: switch to = "0.6.0"

src/adapters/webpush/mod.rs
fabricedesre, line 96: Ok, keep it like this for now then.

samgiles, line 96: If we do need short lived threads it, would be good to have a shared thread pool.  Servo uses http://kimundi.github.io/scoped-threadpool-rs/scoped_threadpool/index.html and I've used it in some personal projects.

src/adapters/webpush/mod.rs
aosmond, line 186: It isn't able to infer the type and automatically wrap it in an Value/Arc/Json if I do that. Just does exactly what I tell it to :).

src/controller.rs
aosmond: Will remove, left over from my removal of the old impl pre-taxonomy.

PR #252, opened 2016-03-23
Generate self-signed certificate, and post the domain with the fingerprint to the registration server.

samgiles:
This PR enables a few other things, would like to land this in stages.

This will generate a self-signed cert, even if we use --disable-tls, because we need to send a unique value (or certificate fingerprint) to the DNS API and the Registration server, and we can use the cert to authenticate the POSTs to these APIs.

This posts the 'domain' to the registration server, although it won't yet do anything with it until we land and deploy: https://github.com/fxbox/registration_server/pull/18 which requires a change in the app.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -106,7 +106,7 @@ use tls::TlsOption;
 use traits::Controller;
 
 docopt!(Args derive Debug, "
-Usage: foxbox [-v] [-h] [-l <hostname>] [-p <port>] [-w <wsport>] [-d <profile_path>] [-r <url>] [-i <iface>] [-t <tunnel>] [-s <secret>] [--disable-tls] [--remote-name <hostname>] [-c <namespace;key;value>]...
+Usage: foxbox [-v] [-h] [-l <hostname>] [-p <port>] [-w <wsport>] [-d <profile_path>] [-r <url>] [-i <iface>] [-t <tunnel>] [-s <secret>] [--disable-tls] [--remote-name <hostname>] [--dns-domain <domain>] [-c <namespace;key;value>]...
 
 Options:
     -v, --verbose            Toggle verbose output.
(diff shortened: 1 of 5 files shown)


samgiles:
To clarify: this can land before the linked PR.  Everything should still work as it does now with this PR, will take a bit of integration to land these changes.

src/main.rs
```diff
@@ -115,11 +115,12 @@ Options:
     -p, --port <port>        Set port to listen on for http connections. [default: 3000]
     -w, --wsport <wsport>    Set port to listen on for websocket. [default: 4000]
     -d, --profile <path>     Set profile path to store user data.
-    -r, --register <url>     Change the url of the registration endpoint. [default: http://localhost:4242/register]
+    -r, --register <url>     Change the url of the registration endpoint. [default: http://localhost:4242]
     -i, --iface <iface>      Specify the local IP interface.
     -t, --tunnel <tunnel>    Set the tunnel endpoint's hostname. If omitted, the tunnel is disabled.
     -s, --tunnel-secret <secret>    Set the tunnel shared secret. [default: secret]
         --disable-tls               Run as a plain HTTP server, disabling encryption.
+        --link-domain <domain>      Set the top level domain for public DNS [default: knilxof.org]
```
fabricedesre: Use --dns-domain instead. No reason to have the project name here.

src/tls/certificate_manager.rs
```diff
@@ -78,8 +91,72 @@ impl CertificateManager {
...
+            .add_name("CN".to_owned(), hostname.clone())
+            .set_sign_hash(Type::SHA256);
+
+        let gen_result = generator.generate();
+        if let Ok((cert, pkey)) = gen_result {
+
+            // Write the cert and pkey PEM files
+
+            // Ensure the directory exists
+            let _ = fs::create_dir_all(directory.clone());
```
fabricedesre: try!(fs::create_dir_all(...))

fabricedesre:
r=me with nits addressed

src/tls/certificate_manager.rs
samgiles: I wanted to ignore the result, because it doesn't matter if it already exists.  I'll check `is_dir` or something, then `try!` `create_dir_all`

fabricedesre: Or you can do something similar to https://github.com/fxbox/foxbox/blob/master/src/profile_service.rs#L61

src/main.rs
```diff
@@ -268,7 +273,8 @@ describe! main {
             assert_eq!(args.flag_local_name, None);
             assert_eq!(args.flag_port, 3000);
             assert_eq!(args.flag_wsport, 4000);
-            assert_eq!(args.flag_register, "http://localhost:4242/register");
```
michielbdejong, line 270: Right, I just ran into this as well, current master branch tries to post to http://localhost:4242/register/register :)

PR #254, opened 2016-03-23
Add Thinkerbell as an adapter.

mcav:
This depends on the following PRs:
- https://github.com/fxbox/taxonomy/pull/35
- https://github.com/fxbox/thinkerbell/pull/54

### DIFF
--- a/src/adapters/mod.rs
+++ b/src/adapters/mod.rs
@@ -13,10 +13,14 @@ pub mod webpush;
 
 mod ip_camera;
 
-use foxbox_taxonomy::adapter::AdapterManagerHandle;
+/// An adapter providing access to Thinkerbell.
+mod thinkerbell;
+
 use foxbox_taxonomy::api::API;
+use foxbox_taxonomy::adapter::AdapterManagerHandle;
 
 use self::philips_hue::PhilipsHueAdapter;
+use self::thinkerbell::ThinkerbellAdapter;
 use service::ServiceAdapter;
 use traits::Controller;
 
@@ -42,6 +46,7 @@ impl<T: Controller> AdapterManager<T> {
         clock::Clock::init(&adapter_manager).unwrap(); // FIXME: We should have a way to report errors
         webpush::WebPush::init(c, &adapter_manager).unwrap();
         ip_camera::IPCameraAdapter::init(adapter_manager.clone(), self.controller.clone()).unwrap();
+        ThinkerbellAdapter::init(adapter_manager.clone()).unwrap(); // FIXME: no unwrap!
     }
 
     fn start_adapter(&mut self, adapter: Box<ServiceAdapter>) {
--- a/src/adapters/thinkerbell/mod.rs
+++ b/src/adapters/thinkerbell/mod.rs
@@ -1,0 +1,412 @@
+//! An adapter providing access to the Thinkerbell rules engine.
+
+use foxbox_taxonomy::adapter::{ AdapterManagerHandle, Adapter, WatchEvent, AdapterWatchGuard };
+use foxbox_taxonomy::api::{ API, Error, InternalError, ResultMap };
+use foxbox_taxonomy::services::{ Setter, Getter, AdapterId, ServiceId, Service, Channel, ChannelKind };
+use foxbox_taxonomy::util::Id;
+use foxbox_taxonomy::values::{ Range, Duration, Type, Value, TypeError, OnOff };
+use foxbox_thinkerbell::compile::ExecutableDevEnv;
+use foxbox_thinkerbell::manager::{ ScriptManager, ScriptId, Error as ScriptManagerError };
+use foxbox_thinkerbell::run::ExecutionEvent;
+
+use timer;
+use transformable_channels::mpsc::*;
+
+use std::collections::{ HashMap, HashSet };
+use std::path::Path;
+use std::sync::{ Arc, Mutex };
+use std::thread;
+
+static ADAPTER_NAME: &'static str = "Thinkerbell adapter (built-in)";
+static ADAPTER_VENDOR: &'static str = "team@link.mozilla.org";
+static ADAPTER_VERSION: [u32;4] = [0, 0, 0, 0];
+
+/// ThinkerbellAdapter hooks up the rules engine (if this, then that) as an adapter.
+///
+/// Each "rule", or "script", is a JSON-serialized structure according to Thinkerbell conventions.
+///
+/// This adapter exposes a root service, with one AddThinkerbellRule setter (to add a new rule).
+/// Each rule that has been added is exposed as its own service, with the following getters/setters:
+/// - Set Enabled (setter) -- toggles whether or not the script is enabled
+/// - Get Enabled (getter) -- returns whether or not the script is enabled
+/// - Remove (setter) -- removes the script
+///
+/// This adapter performs most actions by delegating channel messages to its main thread.
+#[derive(Clone)]
+pub struct ThinkerbellAdapter<T> where T: API + Clone + Send + 'static {
+
+    /// The sending end of the channel for sending messages to ThinkerbellAdapter's main loop.
+    tx: RawSender<ThinkAction>,
+
+    /// A reference to the AdapterManager.
+    adapter_manager: T,
+
+    /// The ID of this adapter (permanently fixed)
+    adapter_id: Id<AdapterId>,
+
+    /// The ID of the root service's "Add Rule" setter.
+    setter_add_rule_id: Id<Setter>,
+}
+
+/// Thinkerbell requires an execution environment following this API.
+#[derive(Clone)]
+struct ThinkerbellExecutionEnv<T> {
+    adapter_manager: T,
+    // FIXME: Timer's not clonable, so we should only use one, right? Does this have to be mutexed?
+    timer: Arc<Mutex<timer::Timer>>
+}
+
+impl<T> ExecutableDevEnv for ThinkerbellExecutionEnv<T>
+    where T: API + Clone + Send + 'static {
+    // Don't bother stopping watches.
+    type WatchGuard = T::WatchGuard;
+    type API = T;
+
+    fn api(&self) -> Self::API {
+        self.adapter_manager.clone()
+    }
+
+    type TimerGuard = timer::Guard;
+    fn start_timer(&self, duration: Duration, sender: Box<ExtSender<()>>) -> Self::TimerGuard {
+        let timer = self.timer.lock().unwrap(); // FIXME: There's no way to report this error...
+        timer.schedule_with_delay(duration.into(), move || {
+            let _ = sender.send(());
+        })
+    }
+}
+
+/// Convert a ScriptManagerError into an API Error.
+/// We can't implement From<T> because ScriptManagerError is in a different crate.
+fn sm_error(e: ScriptManagerError) -> Error {
+    Error::InternalError(InternalError::GenericError(format!("{:?}", e)))
+}
+
+impl<T> Adapter for ThinkerbellAdapter<T> where T: AdapterManagerHandle + API + Clone + Send + 'static {
+    fn id(&self) -> Id<AdapterId> {
+        self.adapter_id.clone()
+    }
+
+    fn name(&self) -> &str {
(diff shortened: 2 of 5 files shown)


Yoric:
Oh, sorry, I hadn't seen that PR. We really need a way to follow them.

src/adapters/mod.rs
```diff
@@ -35,12 +41,14 @@ impl<T: Controller> AdapterManager<T> {
...
     pub fn start<A>(&mut self, adapter_manager: &A)
-        where A: AdapterManagerHandle {
+        where A: AdapterManagerHandle + API + Clone + 'static {
         let c = self.controller.clone(); // extracted here to prevent double-borrow of 'self'
         self.start_adapter(Box::new(PhilipsHueAdapter::new(c.clone())));
+
         self.start_adapter(Box::new(IpCameraAdapter::new(c.clone())));
-        clock::Clock::init(adapter_manager).unwrap(); // FIXME: We should have a way to report errors
-        webpush::WebPush::init(c, adapter_manager).unwrap();
+        clock::Clock::init(&adapter_manager.clone()).unwrap(); // FIXME: We should have a way to report errors
```
Yoric: I believe (I hope!) that this part has been rewritten in the meantime. Cc @fabricedesre .

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+/// - Set Enabled (setter) -- toggles whether or not the script is enabled
+/// - Get Enabled (getter) -- returns whether or not the script is enabled
+/// - Remove (setter) -- removes the script
+///
+/// This adapter performs most actions by delegating channel messages to its main thread.
+#[derive(Clone)]
+pub struct ThinkerbellAdapter<T> where T: API + Clone + Send + 'static {
+    /// The sending end of the channel for sending messages to ThinkerbellAdapter's main loop.
+    tx: RawSender<ThinkAction>,
+    /// A reference to the AdapterManager.
```
Yoric, line 41: Nit: newlines before `///`.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+
+static ADAPTER_NAME: &'static str = "Thinkerbell adapter (built-in)";
+static ADAPTER_VENDOR: &'static str = "team@link.mozilla.org";
+static ADAPTER_VERSION: [u32;4] = [0, 0, 0, 0];
+
+/// ThinkerbellAdapter hooks up the rules engine (if this, then that) as an adapter.
+///
+/// Each "rule", or "script", is a JSON-serialized structure according to Thinkerbell conventions.
+///
+/// This adapter exposes a root service, with one AddThinkerbellRule setter (to add a new rule).
```
Yoric: Note for the future: we'll probably want to add some kind of `PropertyBag` to getters/setters to convey information such as which `get enabled` matches with `set enabled`.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+/// Each rule that has been added is exposed as its own service, with the following getters/setters:
+/// - Set Enabled (setter) -- toggles whether or not the script is enabled
+/// - Get Enabled (getter) -- returns whether or not the script is enabled
+/// - Remove (setter) -- removes the script
+///
+/// This adapter performs most actions by delegating channel messages to its main thread.
+#[derive(Clone)]
+pub struct ThinkerbellAdapter<T> where T: API + Clone + Send + 'static {
+    /// The sending end of the channel for sending messages to ThinkerbellAdapter's main loop.
+    tx: RawSender<ThinkAction>,
```
Yoric, line 39: I like `ThinkAction` :)

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+    adapter_id: Id<AdapterId>,
+    /// The ID of the root service's "Add Rule" setter.
+    setter_add_rule_id: Id<Setter>,
+}
+
+/// Thinkerbell requires an execution environment following this API.
+#[derive(Clone)]
+struct ThinkerbellExecutionEnv<T> {
+    adapter_manager: T,
+    // FIXME: Timer's not clonable, so we should only use one, right? Does this have to be mutexed?
```
Yoric, line 55: Unless `ThinkerbellExecutionEnv` needs to be `Sync` (and I'm almost sure that's not the case), you don't need to `Mutex` it.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+    type WatchGuard = T::WatchGuard;
+    type API = T;
+
+    fn api(&self) -> Self::API {
+        self.adapter_manager.clone()
+    }
+
+    type TimerGuard = timer::Guard;
+    fn start_timer(&self, duration: Duration, sender: Box<ExtSender<()>>) -> Self::TimerGuard {
+        let timer = self.timer.lock().unwrap(); // FIXME: There's no way to report this error...
```
Yoric, line 71: I believe that you don't need that Mutex anyway.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+        ADAPTER_VENDOR
+    }
+
+    fn version(&self) -> &[u32;4] {
+        &ADAPTER_VERSION
+    }
+
+    fn fetch_values(&self, set: Vec<Id<Getter>>) -> ResultMap<Id<Getter>, Option<Value>, Error> {
+        set.iter().map(|id| {
+            let (tx, rx) = channel();
```
Yoric, line 103: For better parallelization, I'd suggest two steps:
1. first, for all items of `set`, create `(tx, rx) and make the`RespondToGetter`calls, putting the`rx` in an array;
2. then, once step 1. is complete, listen to all the elements of the array.

Also using `drain(..)` instead of `iter()` will let you save one `id.clone()`.

All of this can be done in a followup, since we're in a hurry. My bad (and github's) for not seeing your PR.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+            let _ = self.tx.send(ThinkAction::RespondToGetter(tx, id.clone()));
+            match rx.recv() {
+                Ok(result) => (id.clone(), result),
+                // If an error occurs, the channel died!
```
Yoric: (which also means that the thread has died)

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+                }
+            })
+            .collect()
+    }
+
+    fn register_watch(&self, mut watch: Vec<(Id<Getter>, Option<Range>)>,
+        _: Box<ExtSender<WatchEvent>>) ->
+            ResultMap<Id<Getter>, Box<AdapterWatchGuard>, Error> {
+        watch.drain(..).map(|(id, _)| {
+            (id.clone(), Err(Error::GetterDoesNotSupportWatching(id)))
```
Yoric, line 133: Maybe in the future, we'll be able to watch whether a rule is enabled or disabled. Not doing it now is ok, though.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+    fn main(
+        &self,
+        rx: Receiver<ThinkAction>,
+        mut script_manager: ScriptManager<ThinkerbellExecutionEnv<T>, RawSender<(Id<ScriptId>, ExecutionEvent)>>
+    ) {
+        // Store an in-memory list of all of the rules (their getters, setters, etc.).
+        // We need to track these to respond to getter/setter requests.
+        let mut rules = Vec::new();
+
+        'recv: while let Ok(action) = rx.recv() {
```
Yoric: This is officially the first time I see labels in use in Rust :)

Yoric: Also, I believe that the idiomatic manner of doing this is

``` rust
for action in rx {
  ...
}
```

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+            match action {
+                // After a script has been started (elsewhere), start a Service for that script.
```
Yoric: A full service for the script? I thought we only wanted a pair of channels, but that works, too.

Nit: I don't understand your "elsewhere" comment.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+impl<T> ThinkerbellAdapter<T> where T: AdapterManagerHandle + API + Clone + 'static {
+
+    fn main(
+        &self,
+        rx: Receiver<ThinkAction>,
+        mut script_manager: ScriptManager<ThinkerbellExecutionEnv<T>, RawSender<(Id<ScriptId>, ExecutionEvent)>>
+    ) {
+        // Store an in-memory list of all of the rules (their getters, setters, etc.).
+        // We need to track these to respond to getter/setter requests.
+        let mut rules = Vec::new();
```
Yoric: Shouldn't this rather be a `HashMap<Id<Getter>, your data>`?

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+                        },
+                        Err(e) => {
+                            error!("Unable to add Thinkerbell Rule Service: {:?}", e);
+                        }
+                    };
+                },
+                // After a script has been stopped (elsewhere), remove the Service for that script.
+                ThinkAction::RemoveRuleService(script_id) => {
+                    for ref rule in &rules {
+                        if rule.script_id == script_id {
```
Yoric, line 187: These lookups in the full vector don't feel me with hope for the future. I'd rather have a `HashMap` lookup.

This can wait for a followup, if you do it soon.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+                        } else if getter_id == rule.getter_source_id {
+                            match script_manager.get_source(&rule.script_id) {
+                                Ok(source) => {
+                                    let _ = tx.send(Ok(Some(Value::String(Arc::new(source.to_owned())))));
+                                },
+                                Err(e) => {
+                                    let _ = tx.send(Err(sm_error(e)));
+                                }
+                            };
+                            continue 'recv;
```
Yoric, line 214: For instance, this would be avoided by using the `HashMap`. Ok, maybe you'll need several `HashMap`s, but still.

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+                                    expected: Type::ThinkerbellRule,
+                                    got: value.get_type()
+                                })));
+                            }
+                        }
+                    } else {
+                        // The rest of the rules are script/rule-specific.
+                        // NOTE: This linear search is not ideal, but tracking getters/setters in maps
+                        // would be far more complex until we have a simpler way to track state within
+                        // getter/setter API requests. In any case, this loop should be plenty fast for now.
```
Yoric, line 240: If you have an API suggestion, please file it. Now that everybody has tested the API, it's the right time to improve it before we proceed :)

src/adapters/thinkerbell/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+                    match self.add_rule_service(script_id) {
+                        Ok(rule) => {
+                            rules.push(rule);
+                        },
+                        Err(e) => {
+                            error!("Unable to add Thinkerbell Rule Service: {:?}", e);
+                        }
+                    };
+                },
+                // After a script has been stopped (elsewhere), remove the Service for that script.
```
Yoric: I don't think it makes sense to remove the Service without removing the script.

*\* Update *\* Oh, you do remove it. Can you make the documentation clearer that the script has already been removed by that time?

Yoric:
Again, apologies for the delay. If you could get rid of the `Mutex` before landing, this would be great. Plus I'll expect followups for the rest :)

fabricedesre:
\o/

mcav:
Two followups as issues: #274, #275

PR #258, opened 2016-03-24
Registration generic message

michielbdejong:
r? @samgiles 

Here too, took [your PR](https://github.com/fxbox/foxbox/pull/252) and added one commit.


samgiles:
I'm going to address this in the other PR as I need to send a tunnel configured value too and will include this in the message.

michielbdejong:
> I need to send a tunnel configured value too and will include this in the message.

Sounds good!

PR #266, opened 2016-03-24
Use ws_as_addrs to create the Websocket listener rather than passing in port and hostname.

samgiles:
Trivial change to make the creation of the listener consistent with the HttpServer.

@mcav r?


mcav:
lgtm

PR #270, opened 2016-03-25
Enable CORS support for binary response. r=fabrice

azasypkin:
r? @fabricedesre


fabricedesre:
Right, I forgot to change that. Thanks!

PR #271, opened 2016-03-25
Include the scheme in the message posted to the registration server

samgiles:
Ideally the scheme needs to come from the controller, but I don't understand what is going on in the controller with `get_http_root_for_service` and the web socket equivalent as it is right now (the hostname is also potentially going to be wrong in there). The controller API needs reviewing before I make the right change as it's used in an adapter.  But for now: 

The garbage will do

@fabricedesre  r?


ferjm:
This one is obsoleted by https://github.com/fxbox/foxbox/pull/281 right? Can we close it?

samgiles:
yup

PR #273, opened 2016-03-25
Fix snapshot directory creation for IP camera

aosmond:


src/adapters/ip_camera/api.rs
```diff
@@ -62,12 +62,18 @@ pub struct IpCamera {
...
-    pub fn new(udn: &str, url: &str, root_snapshot_dir: &str) -> Self {
-        IpCamera {
+    pub fn new(udn: &str, url: &str, root_snapshot_dir: &str) -> Result<Self, Error> {
+        let camera = IpCamera {
             udn: udn.to_owned(),
             url: url.to_owned(),
             snapshot_dir: format!("{}/{}", root_snapshot_dir, udn)
+        };
+        // Create a directory to store snapshots for this camera.
+        if let Err(err) = fs::create_dir_all(&camera.snapshot_dir) {
```
fabricedesre, line 72: create_dir_all() also fails if the directory already exists. That should be just ignored, no? We do that in https://github.com/fxbox/foxbox/blob/master/src/profile_service.rs#L61 for instance.

aosmond, line 72: Hm. I just copied from the original code: https://github.com/fxbox/foxbox/commit/343415a93b7d0bd27c630bfbeca21a2e871767cf#diff-4e892450a9b75896d13d064d5d81d11aL242

I'll check if it is failing but it isn't printing the warning after rediscovery...

aosmond, line 72: And by rediscovery I mean, discovery (create dir), restart foxbox, and discovery (attempt to create dir again), instead of rediscovery in the same process instance (would ignore).

aosmond, line 72: Didn't seem to be a problem but I'll check just in case. Maybe some platforms behave differently.

PR #279, opened 2016-03-25
Text To Speech adapter using espeak

fabricedesre:


fabricedesre:
Will land once I have the Rpi2 toolchain updated

PR #284, opened 2016-03-30
Resolve #243 - Removing the big Mutex in taxonomy_router

### LINKED ISSUE #243: Figure out how to take advantage of threads in the Taxonomy Router
The architecture of Iron requires all `Handler`s to be functions that implement `Fn(Request) -> Response + Sync`.

This has two drawbacks:
- since the function is `Sync`, we cannot easily manipulate e.g. channels;
- since the function is sync, we cannot simply use asynchronous functions from within it.

For the moment, for deliverable 1, the Taxonomy Router simply grabs a lock that it holds until it has provided a response. This is pretty bad for performance, as we cannot have two simultaneous requests to the Taxonomy Router executed concurrently.

Opening this issue to discuss better ways to implement this.

Cc @fabricedesre

Yoric:
To be compatible with Iron's thread pool, the taxonomy_router needs to
be Sync. This used to be implemented by putting everything in a big
Mutex.

Changes in Taxonomy have made the AdapterManager Sync and
thread-friendly. This patch propagates the changes to all Adapters
currently in FoxBox and to the taxonomy_router itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_script:
 
 script:
     - cargo build # Linter is also executed here
-    - cargo test
+    - RUST_BACKTRACE=1 cargo test
     - jshint static/main/js/*.js static/setup/js/*.js test/selenium/*.js test/integration/lib/*.js test/integration/test/*.js
     - npm run test-integration
     - npm run test-selenium
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,13 +2,13 @@
 name = "foxbox"
 version = "0.1.0"
 dependencies = [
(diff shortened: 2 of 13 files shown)


Yoric:
Mmmh... The Selenium tests that work for me locally fail on Travis. Moreover, they don't seem to actually get even near to my code, so I have a few difficulties understanding what's wrong.

Cc @fabricedesre, @JohanLorenzo .

Yoric:
Just to be more mysterious, increasing the number of threads from 8 to 32 makes it pass.

fabricedesre:
Closing since https://github.com/fxbox/foxbox/commit/9ea8942f91a572ea1b1599187f0aac5d13664943 landed

Issue #246, opened 2016-03-21
Transmit user identification from http to the services


fabricedesre:
The webpush adapter/service is hardcoding its user id because right now we don't surface the current authenticated user from the http layer to the getters and setters. Let's add that.

PR #287, opened 2016-03-30
Issue #246. Transmit user identification from http to the services. r…

ferjm:
…=fabrice

### DIFF
--- a/src/adapters/webpush/mod.rs
+++ b/src/adapters/webpush/mod.rs
@@ -41,6 +41,8 @@ header! { (EncryptionKey, "Encryption-Key") => [String] }
 static ADAPTER_NAME: &'static str = "WebPush adapter (built-in)";
 static ADAPTER_VENDOR: &'static str = "team@link.mozilla.org";
 static ADAPTER_VERSION: [u32;4] = [0, 0, 0, 0];
+// This user identifier will be used when authentication is disabled.
+static NO_AUTH_USER_ID: i32 = -1;
 
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub struct Subscription {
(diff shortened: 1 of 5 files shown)


ferjm:
This needs more work, but I'd love to have some early feedback. Please, ignore the various hacks and `XXX` comments.

@fabricedesre is this the API you had in mind?

I [modified the taxonomy crate](https://github.com/ferjm/taxonomy/commit/63f9c95d8da7277f3afafd0bd9d930b0d1e6866e) to give the user to `fetch_value()` and `send_value()`.

fabricedesre:
Thanks! Overall looks fine to me.
In the case where there is no valid user id found in the taxonomy router, we should just return a 401 error - That may be defense in depth if the auth middleware does it, but that make things clear.

src/adapters/webpush/mod.rs
```diff
@@ -172,9 +172,12 @@ impl<C: Controller> Adapter for WebPush<C> {
         &ADAPTER_VERSION
     }
 
-    fn fetch_values(&self, mut set: Vec<Id<Getter>>) -> ResultMap<Id<Getter>, Option<Value>, Error> {
+    fn fetch_values(&self, mut set: Vec<Id<Getter>>, user: User) -> ResultMap<Id<Getter>, Option<Value>, Error> {
         set.drain(..).map(|id| {
-            let user_id = 1; // FIXME: currently logged in user
+            let user_id = match user {
+                User::None => 0,
```
fabricedesre: Not sure about this one. Should we just bail out? @aosmond ?
Same thing in send_values

ferjm: I've changed this behavior. Now, in case that we have authentication enabled and we have not available user ID, we bail out with a error (we should never get there anyway). In case that we have authentication disabled, we use an ID with value `-1`.

ferjm:
Rebased, again.

ferjm:
@aosmond could you take a look at this patch, please? Thanks.

aosmond:
r+ for the adapter changes

PR #288, opened 2016-03-30
update README.md with cmdline parameters

nojunpark:


fabricedesre:
This is not the full list of command line flags. We at least need to copy the content of https://github.com/fxbox/foxbox/blob/master/src/main.rs#L111

nojunpark:
Added now, should we mention anything else here?

PR #292, opened 2016-03-31
separate tests that can/cannot run on travis

nojunpark:
Some tests require remote (external) servers to be up. exclude them from travis tests


nojunpark:
@JohanLorenzo r?

package.json
```diff
@@ -8,6 +8,7 @@
   "scripts": {
     "test-selenium": "./tools/execute-selenium-tests",
     "test-integration": "npm run clean_db && ./node_modules/.bin/mocha test/integration/test/*.js",
+    "test-integration-travis": "npm run clean_db && ./node_modules/.bin/mocha test/integration/test/travis_*.js",
```
JohanLorenzo, line 11: I think we should execute the tests by default on travis, and only exclude the specific ones.

PR #296, opened 2016-04-01
Change ip camera latest image to be Binary type

dhylands:
This allows ipcam.py to work properly.

r? @Yoric


Yoric:
Sorry about the confusion.

Do you really need all these `print`?

Other than that, r=me.

dhylands:
Without the prints, ipcam.py fails in mysterious ways when somebody goes and changes the getter name. With the prints, it makes it more obvious why something failed.

Yoric:
Well, I do get that I broke something. On the other hand, I suspect that using `debug!` or `info!` from `log` (which is already used in Foxbox) would be a bit nicer.

dhylands:
Merged

dhylands:
The prints are in python code, not rust code

PR #297, opened 2016-04-01
[RFC] Resolves #37 Improve support for binary f?fabrice

Yoric:
This patch converts the taxonomy_router to return multipart data when
the result of a fetch (or, theoretically, anything else) contains
Binary.

Other changes were required to get foxbox to compile.

r? @fabricedesre 

Note: Not tested. I need to come up with a test scenario.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,10 @@ clippy = "0.0.48"
 docopt = "0.6.78"
 docopt_macros = "0.6.80"
 env_logger = "0.3.2"
-foxbox_taxonomy = { git = "https://github.com/fxbox/taxonomy.git", rev = "8cad391" }
-foxbox_thinkerbell = { git = "https://github.com/fxbox/thinkerbell.git", rev = "ec60634" }
+foxbox_taxonomy = { git = "https://github.com/fxbox/taxonomy.git", rev = "c8f05af" }
+foxbox_thinkerbell = { git = "https://github.com/fxbox/thinkerbell.git", rev = "84a0a30" }
+# foxbox_taxonomy = { path = "../taxonomy" }
+# foxbox_thinkerbell = { path = "../thinkerbell" }
 foxbox_users = { git = "https://github.com/fxbox/users.git", rev = "73e8943" }
 get_if_addrs = "0.3.1"
 hyper = "0.7.2"
@@ -24,6 +26,7 @@ iron-cors = { git = "https://github.com/fxbox/iron-cors.git", rev = "96ede73" }
 libc = "0.2.7"
 log = "0.3"
 mio = { git = "https://github.com/carllerche/mio.git" }
+mime = "*"
 mount = "0.0.10"
 nix = { git = "https://github.com/nix-rust/nix.git", rev = "138080" } # Until 0.5.1 is released
 openssl = "0.7.6"
@@ -51,9 +54,15 @@ version = "0.2.6"
 default-features = true
 features = ["ssl"]
 
+[dependencies.multipart]
+version = "^0.5.0"
+default-features = false
+features = ["iron"]
+
 [dev-dependencies]
 stainless = "0.1.4"
 iron-test = "0.2.0"
 regex = "0.1.55"
 rand = "0.3.14"
 tempdir = "0.3.4"
+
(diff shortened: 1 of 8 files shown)


fabricedesre:
Does that still send back a simple, non-multipart http response when we have a single binary value? My understanding is no, and I think we want to keep that for webiness (but to be honest I haven't tried to send a single-part multipart message to ``).

Cargo.toml
```diff
@@ -14,8 +14,10 @@ clippy = "0.0.48"
 docopt = "0.6.78"
 docopt_macros = "0.6.80"
 env_logger = "0.3.2"
-foxbox_taxonomy = { git = "https://github.com/fxbox/taxonomy.git", rev = "8cad391" }
-foxbox_thinkerbell = { git = "https://github.com/fxbox/thinkerbell.git", rev = "ec60634" }
+foxbox_taxonomy = { git = "https://github.com/fxbox/taxonomy.git", rev = "c8f05af" }
+foxbox_thinkerbell = { git = "https://github.com/fxbox/thinkerbell.git", rev = "84a0a30" }
+# foxbox_taxonomy = { path = "../taxonomy" }
+# foxbox_thinkerbell = { path = "../thinkerbell" }
```
fabricedesre, line 20: nit: remove

Cargo.toml
```diff
@@ -24,6 +26,7 @@ iron-cors = { git = "https://github.com/fxbox/iron-cors.git", rev = "96ede73" }
 libc = "0.2.7"
 log = "0.3"
 mio = { git = "https://github.com/carllerche/mio.git" }
+mime = "*"
```
fabricedesre, line 29: set a specific version

src/taxonomy_router.rs
```diff
@@ -208,7 +201,7 @@ impl Handler for TaxonomyRouter {
         // Fetching and getting values.
         // We can't use a GET http method here because the Fetch() DOM api
         // doesn't allow bodies with GET and HEAD requests.
-        payload_api!(fetch_values, Vec<GetterSelector>, ["channels", "get"], Method::Put, binary);
+        payload_api!(fetch_values, Vec<GetterSelector>, ["channels", "get"], Method::Put, simple);
```
fabricedesre, line 204: It seems that we don't need the simple!() macro anymore and we can re-inline its content in payload_api!()

Yoric:
> Does that still send back a simple, non-multipart http response when we have a single binary value? My understanding is no, and I think we want to keep that for webiness (but to be honest I haven't tried to send a single-part multipart message to ).

I don't think that it makes sense. This would mean that the same request will sometimes return a single binary value and sometimes a multipart. If we want support for single binary value, I believe that we need to come up with another mechanism.

Maybe `/api/v1/getters/${URLEncoded(id)}`?

fabricedesre:
Yes, I could live with that. But we really need something more webby than the current api. Simple things are too hard currently.

Yoric:
Filed that discussion at https://github.com/fxbox/foxbox/issues/302.

Yoric:
@JohanLorenzo Do you have suggestions for how to test this? I assume that Selenium is the right choice for that, right?

JohanLorenzo:
Discussed offline with @Yoric. At first, I don't think Selenium is suited for this type of job. Yoric mentioned we just want to make sure the request is correctly handled by `XMLHttpRequest`. If we wanted to go the Selenium way, we'd require to write an extra HTML page that contains a JS call that uses XMLHttpRequest. It seems overkill, knowing that Selenium tests are usually meant for UI testing.

I think we can reduce the number of components by just letting node crafting the XMLHttpRequest. It seem [a library](https://www.npmjs.com/package/xmlhttprequest) does that emulation for us.

As a consequence, I'd put this test at the same level as the Chakram tests, @npark-mozilla wrote. What do you think @npark-mozilla ?

nojunpark:
So chakram doc doesn't explicitly say that it supports multipart response, but it would be quite odd if it doesn't.  I'll put it in my backlog to try this out.  I suppose what I can do is check for boundaries, and the content type and content disposition for each component of the response.

JohanLorenzo:
I found this PR thanks to https://github.com/pulls/mentioned. There hasn't been any activity for more than 3 years. I'm closing it for the sake of cleanup (Gaia is not a supported project anymore). If you think it still has some use, please reopen it.

PR #298, opened 2016-04-01
Disable tests that uses remote network

nojunpark:


nojunpark:
@JohanLorenzo r?

nojunpark:
Merging based on conversation with @JohanLorenzo earlier

PR #304, opened 2016-04-04
add OpenZWave adapter

julienw:


Cargo.lock
```diff
@@ -5,7 +5,7 @@ dependencies = [
  "chrono 0.2.21 (registry+https://github.com/rust-lang/crates.io-index)",
```
julienw, line 5: updated Cargo.lock after adding the new dependency to openzwave-adapter.

Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "foxbox"
 version = "0.1.0"
 authors = ["fabrice <fabrice@desre.org>"]
 build = "build.rs"
+exclude = [ "config/**" ]
```
julienw, line 6: The openzwave library writes in this directory, we don't want to retrigger a build after each run. ;)

config/openzwave/.gitignore
```diff
@@ -0,0 +1,4 @@
+# This directory will be populated by openzwave
+
+*
+!.gitignore
```
julienw, line 4: this is where the openwave adapter outputs logs and the network persistent XML file. This is also where we can put preconfigured config files.

julienw:
Ready to merge

julienw:
ah crap I think i'll need to either build openzwave in the crate or install it for Travis.

julienw:
pending https://github.com/fxbox/openzwave-rust/pull/22

Cargo.toml
ferjm, line 6: We already have a profile folder outside of the src tree where this configuration could fit better and we also have a way to specify the location of this profile folder through a [cmd line arg](https://github.com/fxbox/foxbox/blob/master/src/main.rs#L118). It would be nice if the adapter could take this location and use it to store its configuration.

ferjm:
Is the plan to keep this adapter as a separate crate? In that case, what makes it different from the rest of adapters? I'm not saying that we shouldn't have the adapters outside foxbox, but looking at the current status, it seems that the way to do it for now is inside of foxbox itself.

Cargo.toml
ferjm, line 6: File a follow up issue for this, please.

julienw, line 6: done: https://github.com/fxbox/foxbox/issues/309

.travis.yml
```diff
@@ -15,7 +15,7 @@ env:
 
 before_install:
   - sudo apt-get -qq update
-  - sudo apt-get install -y avahi-daemon libavahi-client-dev libavahi-common-dev libavahi-common3 libdbus-1-dev libupnp-dev libespeak-dev
+  - sudo apt-get install -y avahi-daemon libavahi-client-dev libavahi-common-dev libavahi-common3 libdbus-1-dev libupnp-dev libespeak-dev libudev-dev
```
ferjm, line 18: I am afraid that libudev is not available on OS X (or I couldn't find it), so we'll need to disable this adapter for OS X builds for now.

src/adapters/mod.rs
```diff
@@ -28,6 +28,8 @@ use self::thinkerbell::ThinkerbellAdapter;
 use service::ServiceAdapter;
 use traits::Controller;
 
+use openzwave::Adapter as OpenzwaveAdapter;
```
ferjm, line 31: Add `#[cfg(target_os = "linux")]` here.

src/adapters/mod.rs
```diff
@@ -62,6 +64,8 @@ impl<T: Controller> AdapterManager<T> {
         webpush::WebPush::init(c, manager).unwrap();
         ip_camera::IPCameraAdapter::init(manager, self.controller.clone()).unwrap();
         ThinkerbellAdapter::init(manager).unwrap(); // FIXME: no unwrap!
+        OpenzwaveAdapter::init(manager).unwrap();
```
ferjm, line 67: And here

src/main.rs
```diff
@@ -58,6 +58,9 @@ extern crate ws;
 extern crate multicast_dns;
 extern crate xml;
 
+// adapters
+extern crate openzwave_adapter as openzwave;
```
ferjm, line 62: `#[cfg(target_os = "linux")]`

.travis.yml
julienw, line 18: we don't need it for OSX :) (udev is purely a linux thing). On OSX it uses some CoreFoundation and IOKit stuff that I don't really know.

ferjm:
r=me

Thanks for doing this work!

PR #313, opened 2016-04-07
Use latest firefox instead of 31.0esr

JohanLorenzo:
Selenium tests don't pass locally anymore. This is due to Firefox 46+ which doesn't requires addon signin. That's how selenium took control of Firefox. Now we have to use Marionette. This can simply be done by upgrading to the latest Selenium

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ cache:
         - node_modules
         - $TRAVIS_BUILD_DIR/target
 
+addons:
+  firefox: latest
+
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
As you can see in the commits, the first one shows that the latest Firefox (release) doesn't work with the version of selenium we had. Then we you upgrade (3rd commit), tests are back passing.

Tell me if it works for you locally @ferjm . r? @ferjm 

PS: I'll squash the commits once the review is done.

ferjm:
It works \o/ Thank you Johan!

PR #321, opened 2016-04-07
Prevent cargo from building every time

JohanLorenzo:
When I debugged #313, I noticed cargo was rebuilding nearly every time I ran `npm run test-selenium`. After some investigation, it turned out npm wrote `npm-debug.log` each time tests failed.

Then, in order to optimize the execution time, I propose to stop cargo from building when some file are changed.

r? @ferjm

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,12 @@ name = "foxbox"
 version = "0.1.0"
 authors = ["fabrice <fabrice@desre.org>"]
 build = "build.rs"
-exclude = [ "config/**" ]
+# Files that shouldn't retrigger a new rust build if they are changed
+exclude = [ "config/**", "**/*.conf",
+            "node_modules/**/*", "test/**/*", "tools/**/*", # Non-rust scripts
+            "**/*.log", "**/*.sqlite", # Application generated data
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -3,7 +3,12 @@ name = "foxbox"
 version = "0.1.0"
 authors = ["fabrice <fabrice@desre.org>"]
 build = "build.rs"
-exclude = [ "config/**" ]
+# Files that shouldn't retrigger a new rust build if they are changed
+exclude = [ "config/**", "**/*.conf",
+            "static/**/*", "node_modules/**/*", "test/**/*", "tools/**/*", # Non-rust scripts
```
ferjm: Just to double check, this does not include changes to `static/shared`, right? Note that we need to run the build script every time something changes in this folder.

ferjm:
r=me assuming that changes to `static/shared` still trigger the build.

Cargo.toml
JohanLorenzo: Good catch. It actually does. I removed it.

JohanLorenzo:
I removed `static/**/*` so `static/shared` still triggers a new Rust compilation.

PR #332, opened 2016-04-11
Rewriting Philips Hue adapter for taxonomy

cr:


docs/scripts/huedemo.sh
```diff
@@ -1,58 +1,67 @@
 #!/bin/bash
 
-# You must disable authenticaion before using this script.
```
Yoric, line 3: Nit: Might be useful to explain what this file is for. Also, why in `docs/`?

src/adapters/mod.rs
```diff
@@ -16,7 +16,7 @@ pub mod tts;
 mod ip_camera;
 
 /// An adapter dedicated to the Philips Hue
-mod philips_hue;
+pub mod philips_hue;
```
Yoric: Why `pub`?

src/adapters/philips_hue/discovery.rs
```diff
@@ -0,0 +1,167 @@
...
+use transformable_channels::mpsc::*;
+use upnp::{ UpnpListener, UpnpManager, UpnpService };
+
+pub struct Discovery<C> {
+    adapter: PhilipsHueAdapter<C>,
+    upnp_manager: Arc<Mutex<Arc<UpnpManager>>>,
+}
+
+impl<C: Controller> Discovery<C> {
+    // pub fn new(manager: Arc<AdapterManager>, controller: C,
```
Yoric: Nit: dead code

docs/scripts/huedemo.sh
cr, line 3: Keeping the tradition. That's where the old demo script was located, too.

src/adapters/philips_hue/discovery.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        if !url.ends_with(":80/") {
+            debug!("Unsupported HTTP port, expected 80.");
+            return false;
+        }
+
+        let ip: &str = url.rsplit("//").nth(0).unwrap_or("");
+        let ip: &str = ip.split(':').nth(0).unwrap_or("");
+
+        // CAVE: assuming raw IPv4 addresses
```
Yoric, line 130: Nit: I believe that you're the first person I have seen who documents her code in latin :)
I'm not sure everybody reads it, though.

src/adapters/philips_hue/discovery.rs
```diff
@@ -0,0 +1,167 @@
...
+        if ip.len() < 7 || ip.len() > 15 {
+            debug!("Unexpected IP address format");
+        }
+
+        let mut id = serial.clone();
+        if id.len() == 12 {
+            // Turn serial into actual Hue ID by inserting "fffe" mid string
```
Yoric, line 137: Is this documented somewhere?

src/adapters/mod.rs
cr: You're right, it's no longer required since I switched all local `use` declarations to use `super::`. Fixed.

src/adapters/philips_hue/discovery.rs
cr: Oops! Removed.

src/adapters/philips_hue/discovery.rs
```diff
@@ -0,0 +1,167 @@
...
+            id.insert(6, 'e');
+            id.insert(6, 'f');
+            id.insert(6, 'f');
+            id.insert(6, 'f');
```
Yoric: Nit: The following might be more readable.

``` rust
let prefix = &id[0..6];
let suffix = &id[6..];
let id = format!("{}fffe{}", prefix, suffix);
```

src/adapters/philips_hue/discovery.rs
cr, line 130: It really was a thing in the late 90's and early 00's. It bled over from medical papers, IIRC. Hard to believe people won't figure it out.

src/adapters/philips_hue/http.rs
```diff
@@ -2,8 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
```
Yoric, line 2: Nit: Any chance you could explain what this module is about? I guess that the Philips Hue exposes a http server and you need a http client to communicate with it, right?

src/adapters/philips_hue/discovery.rs
cr, line 137: I don't think it's documented, but one finds out eventually when comparing nUPnP and UPnP IDs after wondering for a moment why your hubs get added twice. UPnP returns the pure MAC addresses while nUPnP places that `fffe` insert for unknown reasons.

src/adapters/philips_hue/hub.rs
```diff
@@ -0,0 +1,111 @@
...
+pub struct Hub<C> {
+    pub adapter: PhilipsHueAdapter<C>,
+    pub id: String,
+    pub ip: String,
+    pub api: Arc<HubApi>,
+}
+
+impl<C: Controller> Hub<C> {
+    pub fn new(adapter: PhilipsHueAdapter<C>, id: &str, ip: &str) -> Self {
+        // Get API token from config store, default to a random UUID.
```
Yoric, line 31: It would be useful to explain what that API token is for.

src/adapters/philips_hue/discovery.rs
Yoric, line 137: In that case, could you document this somewhere?

src/adapters/philips_hue/hub.rs
```diff
@@ -0,0 +1,111 @@
...
+            api: Arc::new(HubApi::new(id, ip, &token)),
+        }
+    }
+    pub fn start(&self) {
+        info!("Starting Hue Hub Service for {}", self.id);
+        let adapter = self.adapter.clone();
+        let id = self.id.clone();
+        let api = self.api.clone();
+
+        thread::spawn(move || {
```
Yoric, line 53: I don't see any way to kill that thread.

Yoric, line 53: Would it make sense to rewrite this using `timer` instead of a thread that waits forever?

src/adapters/philips_hue/hub.rs
```diff
@@ -0,0 +1,111 @@
...
+                        adapter.controller.adapter_notification(
+                            json_value!({ adapter: "philips_hue", message: "PairingSuccess",
+                                hub: id }));
+                    } else {
+                        warn!("Pairing timeout with Philips Hue Bridge ID {}", id);
+                        adapter.controller.adapter_notification(
+                            json_value!({ adapter: "philips_hue", message: "PairingTimeout",
+                                hub: id }));
+                        // Giving up for this Hub.
+                        // Re-try pairing every hour.
```
Yoric, line 89: Could you file a followup to make it not-hardcoded? Perhaps we should add an API to adapters (not just this one) to force re-attempting to pair?

src/adapters/philips_hue/hub.rs
```diff
@@ -0,0 +1,111 @@
...
+                    hs.config.ipaddress);
+
+                let light_ids = api.get_lights();
+                for light_id in light_ids {
+                    debug!("Found light {} on hub {}", light_id, id);
+                    adapter.send(HueAction::AddLight(id.to_owned(), light_id.to_owned()));
+                }
+
+                loop { // Forever
+                    thread::sleep(Duration::from_millis(60*1000));
```
Yoric, line 112: That looks pretty strange. What's the goal of this infinite loop that does nothing?

src/adapters/philips_hue/discovery.rs
cr, line 137: The plan is to collect info on our targets in the Wiki at some point. Making a note to document it there when the time comes.

src/adapters/philips_hue/discovery.rs
cr: Agreed. Changed.

src/adapters/philips_hue/http.rs
cr, line 2: Correct. I thought this much was obvious. I'll add a line of text to clarify.

src/adapters/philips_hue/hub.rs
cr, line 31: Ok, done.

src/adapters/philips_hue/hub.rs
cr, line 53: This is a stub that will be extended to take over some monitoring function, for example polling sensor values or checking whether lights are going offline.

Stopping these threads will become important for proper teardown, but none of that logic is in place in the adapter management layers, yet.

cr, line 53: There's actually a lot more about expected adapter behavior to define here:
- Are adapters expected to quit if no target devices are detected?
- Should adapters leave a minimal detection stub running for when devices pop up?
- What's the penalty on resources if many unused adapters keep lingering in the background?
- Should the user instead send an event to a Foxbox that adapters to re-run discovery? What's that interface going to look like for adapters?

src/adapters/philips_hue/hub.rs
cr, line 89: Yes, this is part of a larger discussion as outlined above.

src/adapters/philips_hue/hub.rs
cr, line 112: It's a stub that will take over hub monitoring functions (polling etc.) in the future. See https://github.com/fxbox/foxbox/pull/332#discussion_r59225523

src/adapters/philips_hue/discovery.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    pub fn do_upnp_discovery(&self) {
+        let upnp = self.upnp_manager.lock().unwrap();
+        // Still wondering which one of these triggers a bridge response:
+        // upnp.search(Some("urn:schemas-upnp-org:device:basic:1".to_owned())).unwrap();
+        // upnp.search(Some("urn:schemas-upnp-org:device:Basic:1".to_owned())).unwrap();
+        // upnp.search(Some("urn:schemas-upnp-org:device:libhue:idl".to_owned())).unwrap();
+        // upnp.search(Some("upnp:rootdevice".to_owned())).unwrap();
+        // upnp.search(Some("uuid:2f402f80-da50-11e1-9b23-00178825681a".to_owned())).unwrap();
+        upnp.search(None).unwrap();
```
fabricedesre: Does that actually do anything?

src/adapters/philips_hue/discovery.rs
```diff
@@ -0,0 +1,166 @@
...
+impl<C: Controller> PhilipsHueUpnpListener<C> {
+    pub fn new(adapter: PhilipsHueAdapter<C>) -> Box<Self> {
+        Box::new(PhilipsHueUpnpListener {
+            adapter: adapter
+        })
+    }
+}
+
+impl<C: Controller> UpnpListener for PhilipsHueUpnpListener<C> {
+    // This will called each time that the device advertises itself using UPNP.
```
fabricedesre: nit: This will be called ...

src/adapters/philips_hue/discovery.rs
```diff
@@ -0,0 +1,166 @@
...
+        macro_rules! try_get {
+            ($hash:expr, $key:expr) => (match $hash.get($key) {
+                Some(val) => val,
+                None => return false
+            })
+        }
+
+        let model_name = try_get!(service.description, "/root/device/modelName");
+
+        if !model_name.starts_with("Philips hue bridge") {
```
fabricedesre: Can you move this hardcoded string to a static?

src/adapters/philips_hue/discovery.rs
```diff
@@ -0,0 +1,166 @@
...
+        debug!("UPnP announcement for Philips Hue bridge serial {} model {} at {}",
+            serial, model, url);
+
+        if !url.starts_with("http://") {
+            debug!("Unsupported URL scheme, expected HTTP.");
+            return false;
+        }
+
+        if !url.ends_with(":80/") {
+            debug!("Unsupported HTTP port, expected 80.");
```
fabricedesre, line 123: why don't we support ports other than 80? Also, since it's the default one, is it always explicitely set? Looks like you want proper url parsing there instead.

src/adapters/philips_hue/hub.rs
```diff
@@ -0,0 +1,116 @@
...
+                    hs.config.modelid, hs.config.bridgeid, hs.config.swversion,
+                    hs.config.ipaddress);
+
+                let light_ids = api.get_lights();
+                for light_id in light_ids {
+                    debug!("Found light {} on hub {}", light_id, id);
+                    adapter.send(HueAction::AddLight(id.to_owned(), light_id.to_owned()));
+                }
+
+                loop { // Forever
```
Yoric, line 110: Infinite infinite loop? Pretty weird.

src/adapters/philips_hue/hub.rs
Yoric, line 53: Well, in this case, the adapter cannot be safely instantiated in the background since the thread will never, ever quit.

src/adapters/philips_hue/lights.rs
```diff
@@ -0,0 +1,146 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
Yoric, line 1: It would be nice to have a `//!` comment in each module explaining what the module is about.

src/adapters/philips_hue/lights.rs
```diff
@@ -0,0 +1,146 @@
...
+        services: LightServiceMap) -> Result<(), Error>
+    {
+        let adapter_id = create_adapter_id();
+        let status = self.api.get_light_status(&self.light_id);
+        if status.lighttype == "Extended color light" {
+            info!("New Philips Hue Extended Color Light service for light {} on bridge {}",
+                self.light_id, self.hub_id);
+            let mut service = Service::empty(self.service_id.clone(), adapter_id.clone());
+            service.properties.insert(CUSTOM_PROPERTY_MANUFACTURER.to_owned(),
+                                                status.manufacturername.to_owned());
```
fabricedesre: nit: indentation is strange

src/adapters/philips_hue/discovery.rs
cr: Callbacks are triggered even without the search, but then I need to assume that some other adapter's search is also triggering the Hue bridges, but disabling the search here would mean introducing an unwanted dependency.

src/adapters/philips_hue/mod.rs
```diff
@@ -2,279 +2,287 @@
...
+    {
+        set.drain(..).map(|id| {
+            let light = match self.services.lock().unwrap().getters.get(&id) {
+                Some(light) => light.clone(),
+                None => return (id.clone(), Err(Error::InternalError(InternalError::NoSuchGetter(id))))
+            };
+
+            if id == light.get_available_id {
+                let rsp = light.get_available();
+                if rsp {
```
fabricedesre: inline light.get_available() here?

src/adapters/philips_hue/mod.rs
```diff
@@ -2,279 +2,287 @@
...
+                    return (id, Ok(Some(Value::OnOff(OnOff::On))));
+                } else {
+                    return (id, Ok(Some(Value::OnOff(OnOff::Off))));
+                }
             }
-        }
+            if id == light.get_power_id {
+                let rsp = light.get_power();
+                if rsp {
```
fabricedesre: same here

src/adapters/philips_hue/discovery.rs
cr: Fixed.

src/adapters/philips_hue/mod.rs
```diff
@@ -2,279 +2,287 @@
...
-    response.headers.set(AccessControlAllowOrigin::Any);
-    Ok(response)
-}
+            if id == light.set_power_id {
+                match value {
+                    Value::OnOff(OnOff::On)  => { light.set_power(true); },
+                    Value::OnOff(OnOff::Off) => { light.set_power(false); },
+                    _ => {
+                        return (id, Err(Error::TypeError(TypeError {
+                                        got:value.get_type(),
```
fabricedesre: nit: space after "got:"

src/adapters/philips_hue/discovery.rs
cr: `search(None)` triggers a search for _ssdp:all_. I added a comment to clarify this.

src/adapters/philips_hue/discovery.rs
cr: Moved both.

src/adapters/philips_hue/discovery.rs
cr, line 123: No, we don't need proper parsing for **UPnP** here. That check is probably even a bit over the top, because it's unlikely that Hue bridges will switch port any time soon. **nUPnP** doesn't even support listing the port at all.

src/adapters/philips_hue/lights.rs
cr: Fixed to use prevalent style (+1 tab)

src/adapters/philips_hue/hub.rs
cr, line 110: As mentioned before, this will be filled with purpose soon.

src/adapters/philips_hue/lights.rs
cr, line 1: Ok, I have added a documentation header to every module.

src/adapters/philips_hue/mod.rs
cr: Inlined.

src/adapters/philips_hue/mod.rs
cr: Inlined, too.

src/adapters/philips_hue/mod.rs
cr: Neatly spaced now.

src/adapters/philips_hue/lights.rs
```diff
@@ -0,0 +1,146 @@
...
+impl Light {
+    pub fn new(api: Arc<HubApi>, hub_id: &str, light_id: &str)
+        -> Self
+    {
+        Light {
+            api: api,
+            hub_id: hub_id.to_owned(),
+            light_id: light_id.to_owned(),
+            service_id: create_light_id(&hub_id, &light_id),
+            get_available_id: create_getter_id("available", &hub_id, &light_id),
```
Yoric, line 46: Any chance you could document what "available" means?

src/adapters/philips_hue/hub.rs
Yoric, line 53: Right now, the main problem I see is that you cannot safely start an adapter (in particular for unit testing) as it will have side-effects until the process quits.

Regarding your questions.

> Are adapters expected to quit if no target devices are detected?

I don't think so.

> Should adapters leave a minimal detection stub running for when devices pop up?

Yes.

> What's the penalty on resources if many unused adapters keep lingering in the background?

Time will tell.

> Should the user instead send an event to a Foxbox that adapters to re-run discovery? What's that interface going to look like for adapters?

I believe that we should also support this. Maybe something as simple as

``` rust
/// Try again to detect any new connected devices. This method is called by the user from the
/// UX.
fn discover(&self) {
  // By default, do nothing.
}
```

Yoric:
r=me, but please file followup bugs for all the stuff that's TODO but not necessarily indicated in the code.

src/adapters/philips_hue/lights.rs
cr, line 46: Allright, but I'm a bit lost for words here, because it means just that. It goes to "Off" when the device is not plugged in.

cr:
We really need to fix the integration test before landing.

src/adapters/philips_hue/lights.rs
Yoric, line 46: Well, I had no clue that "available" meant "device is plugged", so this is useful :)

cr, line 46: We came across this topic in the past: This `availability` extension might evolve into something that every service exposes, because it signifies the distinction between "Foxbox knows about this device, but it's currently unreachable" and "Foxbox can actually talk to this device".

Yoric, line 46: Fair enough. I'm just saying that, by reading the code, I have no clue what "available" means, so it deserves documentation.

cr:
> No need to disable authentication by default ;)

Oh gee. Fixed by https://github.com/fxbox/foxbox/pull/362

PR #333, opened 2016-04-11
Travis: Add OS X and optional nightly build

### LINKED ISSUE #307: Travis: Expand our builds to mac and most recent nightlies
As @julienw proposed, we could run the tests on more than just 1 environment:
- Mac: many devs are running under this env. 
- Latest nightly: This would enable us to know when a nightly build fails. It would give us a backlog when somedody wants to rust up. This build should allow failure

There is [a working example](https://github.com/nix-rust/nix/blob/master/.travis.yml) on how to set up travis under these envs and allowing failures.

We'll be able to make these changes, after the next rust up happened. The main reason is that we'll be able to start with a green build on latest nightly

JohanLorenzo:
Fix #307


JohanLorenzo:
The test has been deactivated on Mac OS only. r? @samgiles

samgiles:
Nice! r+ from me.

Issue #325, opened 2016-04-08
I'd like to update rules with the Thinkerbell adapter


Yoric:
STR:

```
curl -X PUT http://localhost:3000/api/v1/channels/set -d '{"select":{"kind":"AddThinkerbellRule"},"value":{"ThinkerbellRule":{"name":"Hello, Thinkerbell","source":"{\"rules\":[{\"conditions\":[{\"source\":[{\"id\":\"OpenZWave/72057594126794752 (Sensor)\"}],\"kind\":\"OpenClosed\",\"range\":{\"Eq\":{\"OpenClosed\":\"Open\"}}}],\"execute\":[{\"destination\":[{\"kind\":\"Log\"}],\"kind\":\"Log\",\"value\":{\"String\":\"Closed\"}}]}]}"}}}'
```

(do it twice)

Expected results: the second rule should replace the first one.

Actual results:

```
[2016-04-08 16:09:46.655] ERROR Unable to add Thinkerbell Rule Service: InternalError(DuplicateService(Id { id: Atom('thinkerbell/Hello, Thinkerbell' type=dynamic), phantom: Phantom { phantom: PhantomData } }))
```

This is pretty annoying for demo development, debugging, etc. @mcav, could you take a look at this? 

Cc @ferjm

mcav:
Yeah, I'll take this.

mcav:
Resolved in #334.

Yoric:
Now I get the following response: `{"thinkerbell-add-rule":{"Error":{"InternalError":{"GenericError":"RunError(StartStopError(AlreadyRunning))"}}}}`.

PR #334, opened 2016-04-11
Allow recreating Thinkerbell rules without removing the old ones first.

mcav:
@Yoric It turns out, even before this patch, you could overwrite existing rules; the error you received was misleading. This patch corrects that error.

However, both before _and_ with this patch, I see the following REST API error when adding rules (even with the curl command in #325):

```
{"thinkerbell-add-rule":{"Error":{"InternalError":{"GenericError":"ParseError(\"MissingField { name: \\\"name\\\", at: \\\".\\\" }\")"}}}}
```

I don't know where that's coming from. Either that's a new-ish phenomenon (as of the past three days), or we've been ignoring the JSON response.

r? @Yoric

### DIFF
--- a/src/adapters/thinkerbell/mod.rs
+++ b/src/adapters/thinkerbell/mod.rs
@@ -163,6 +163,7 @@ struct ThinkerbellRule {
 
 impl ThinkerbellAdapter {
 
+    #[allow(cyclomatic_complexity)]
     fn main(
         &self,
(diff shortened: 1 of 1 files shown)


Yoric:
@mcav The format of rules has changed a tiny bit: rules themselves now need a field `name`.

Yoric:
r=me, but I'd really like unit tests, too, possibly in a followup.

PR #349, opened 2016-04-12
Philips Hue now uses LightOn for lights being on/off

Yoric:
This will make the Recipes more readable.

r? @cr


fabricedesre:
lgtm

Issue #358, opened 2016-04-13
[Workaround] thread 'Registrar' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 2, message: "No such file or directory" } }'


Yoric:
I have encountered this twice. Unfortunately, I didn't have `RUSTC_BACKTRACE=1`.

This has happened once during startup and once during runtime. The latter occurrence matched with me changing Wifi network, I don't know if there is causality.

Cc @ferjm , @samgiles

Yoric:
Ah, well, since my latest rebase, it seems 100% reproducible on startup:

```
$ cargo run
[2016-04-13 17:05:20.148] INFO  nUPNP server: Starting registration with http://knilxof.org:4242/register
[2016-04-13 17:05:20.149] INFO  Got ip address: 10.243.30.225
thread 'Registrar' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 2, message: "No such file or directory" } }', ../src/libcore/result.rs:746
stack backtrace:
   1:        0x10b8ec388 - std::sys::backtrace::tracing::imp::write::h714760a4c8c0cdd8
   2:        0x10b8f14c5 - std::panicking::default_hook::_$u7b$$u7b$closure$u7d$$u7d$::hff309ab1d83ffd90
   3:        0x10b8f10fe - std::panicking::default_hook::h08ad3bb09872855b
   4:        0x10b8da946 - std::sys_common::unwind::begin_unwind_inner::h406d5f1a330b854b
   5:        0x10b8db7de - std::sys_common::unwind::begin_unwind_fmt::h57ea3fbee1a40196
   6:        0x10b8eb657 - rust_begin_unwind
   7:        0x10b918390 - core::panicking::panic_fmt::ha6b3c19493c123b3
   8:        0x10abf3cab - core::result::unwrap_failed::he41fdbb4a3e1914b
   9:        0x10b086800 - _<core..result..Result<T, E>>::unwrap::h73493af4a956f6b0
  10:        0x10b08661f - foxbox::registration::Registrar::get_fingerprint::h07f9b00e3e21d9f6
  11:        0x10b086961 - foxbox::registration::Registrar::get_common_name::hc66bdf16d17299a5
  12:        0x10b086b2e - foxbox::registration::Registrar::get_local_dns_name::h8311cdd8bc200fb0
[...]

PR #360, opened 2016-04-13
Enable WebPush on Mac

aosmond:


README.md
```diff
@@ -215,10 +215,12 @@ You'll need some dependencies installed to build.
 $ brew install openssl libupnp sqlite
 ```
 
-This is required to build the openssl crate using homebrew's openssl:
+This is required to build/link the openssl crate and foxbox using homebrew's openssl:
 
 ``` bash
 $ export DEP_OPENSSL_INCLUDE=/usr/local/opt/openssl/include/
+$ export OPENSSL_LIB_DIR=/usr/local/opt/openssl/lib
+$ export EXTRA_LDFLAGS=-L/usr/local/opt/openssl/lib
```
ferjm, line 223: Do we need to add these exports to the .travis.yml file?

Yoric:
@aosmond I'm trying to understand: why should this fix the issue?

Yoric:
As far as I can tell, this is not going to change anything on my computer. I have built rust-openssl manually (using their `cargo build`) and it already uses my version of openssl that lives in `/usr/local/opt/openssl`.

If you wish to test which version is used by Foxbox, I have a patch pending review in rust-openssl: https://github.com/sfackler/rust-openssl/pull/380#issuecomment-209640698

aosmond:
@Yoric That doesn't seem to be the case for others since they have build issues. Not being a Mac user I find speculating difficult. Feel free to update the README :).

nojunpark:
I think when brew is used to install openssl (like the readme suggests) , such export is necessary, but I guess in case of macports it might be different. Then again I had a few issues with installing openssl using macports so I went back to homebrew.

Yoric:
OK. Unfortunately, it won't fix #358, though.

nojunpark:
I think #358 doesn't happen with brew install method - I remember seeing occasional unwrap() before, but now it's not happening.

Yoric:
Well, we found out that #358 is due to using the wrong version of curl (and not handling failures too gracefully).

PR #369, opened 2016-04-14
Ignore pkcs12 files generated by LE

ferjm:


samgiles:
Well spotted, thanks!  Not sure this file should end up where it is though, and may go away entirely if I land this: https://github.com/fxbox/foxbox/pull/376

r+ though, we should have this in now.

Issue #370, opened 2016-04-14
Tunnel url is no longer registered with the nUPNP server


ferjm:

ferjm:
This was fixed at https://github.com/fxbox/foxbox/pull/378

PR #378, opened 2016-04-14
Issue #370. Tunnel url is no longer registered with the nUPNP server.…

ferjm:
… r=sgiles

r? @samgiles 

Not a beautiful fix, but works for now for the demo

### DIFF
--- a/src/tunnel_controller.rs
+++ b/src/tunnel_controller.rs
@@ -6,7 +6,7 @@
 use std::io::prelude::*;
 use std::process::{ Child, Command };
 use std::io::Result;
-use url::Url;
+use url::{ SchemeData, Url };
 use managed_process::ManagedProcess;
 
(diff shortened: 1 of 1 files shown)


src/tunnel_controller.rs
```diff
@@ -102,19 +102,17 @@ impl Tunnel {
...
-                        Some(host.serialize())
-                    } else {
-                        None
-                    }
-                },
-                Err(_) => None
-            }
+        // We only care about the domain part, but Url won't parse without
+        // a proper scheme, so we append http just to make parse work.
+        match Url::parse(&format!("http://{}", self.config.tunnel_url)) {
```
samgiles: Should we check that `tunnel_url` doesn't already have a scheme?

samgiles:
r+ - with a change to check that `tunnel_url` doesn't already have a scheme. :) Thanks!

PR #351, opened 2016-04-13
Switch to using webpush taxonomy type for notify setter

aosmond:

PR #380, opened 2016-04-14
WebPushNotify now uses ChannelKind::WebPushNotify

Yoric:
r? @aosmond 

This should fix the issue encountered by @ferjm when attempting to test WebPush with Thinkerbell.

Cc @gmarty and @arcturus in case they rely upon the old channelkind.

### DIFF
--- a/src/adapters/webpush/mod.rs
+++ b/src/adapters/webpush/mod.rs
@@ -306,12 +306,23 @@ impl<C: Controller> WebPush<C> {
             )
         }
 
+        try!(adapt.add_setter(Channel {
+            tags: HashSet::new(),
+            adapter: id.clone(),
+            id: setter_notify_id,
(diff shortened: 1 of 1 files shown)


ferjm:
I'm testing this patch, but you may want @aosmond to take a look at the code.

Yoric:
Well, r? @aosmond .

aosmond:
@Yoric r-. This isn't complete because we would need to change the send_values like I did in https://github.com/fxbox/foxbox/pull/351 no? I rebased it so it can merge once travis passes....

Yoric:
@aosmond I don't really see the difference between our PRs. So pick one and land it :)

ferjm:
FWIW the patch works :)

Yoric:
Landed as discussed with @aosmond .

PR #382, opened 2016-04-14
Update the Hue integration test to match the new taxnomy

nojunpark:
This also breaks the non-travis tests, (while would fail anyways) I'll file a separate issue to deal with that one.

### DIFF
--- a/test/integration/test/philips_light_test.js
+++ b/test/integration/test/philips_light_test.js
@@ -7,15 +7,16 @@ var config = new Config('./test/integration/lib/config/foxbox.js');
 var Prepper = require('../lib/testPrepper.js');
 
 Prepper.makeSuite('Control lights locally',function(){
-  var lightinfo;
   var header = new Config('./test/integration/lib/config/header.js');
   var credential = config.get('credential');
-
   var setupURL = config.get('foxbox.url') + '/users/setup';
   var loginURL = config.get('foxbox.url') + '/users/login';
-  var serviceURL = config.get('foxbox.url') + '/services/';
-  var serviceListURL = serviceURL + 'list';
-
+  var channelURL = config.get('foxbox.url') + '/api/v1/channels';
+  var getURL = channelURL + '/get';
+  var setURL = channelURL + '/set';
+  var getterPayload = [{'kind': 'LightOn'}];
+  var lights;
+  
   it('create a new credential',function(){
 
     return chakram.post(setupURL,credential)
@@ -39,6 +40,7 @@ Prepper.makeSuite('Control lights locally',function(){
 
   describe ('once logged in', function () {
     before(function() {
+      
       return chakram.post(loginURL,null,{'headers' : header})
       .then(function(resp){
        header.Authorization = 'Bearer ' + resp.body.session_token;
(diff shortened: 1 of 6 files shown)


nojunpark:
@JohanLorenzo r?

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
     });
 
     it('check 3 bulbs are registered',function(){
 
-      return chakram.get(serviceListURL)
+      // collect all getters for the lightbulbs
+      var getURL = channelURL + '/get';
+      var getterPayload = [{'kind': 'LightOn'}];
+      
+      return chakram.put(getURL,getterPayload)
```
JohanLorenzo, line 54: _Improvement for a next PR_

I think we could add more context by putting this call in a class `FoxboxLights`. This class would be like a page class in WebUI testing.

We could name the function `fetchAllGetters`. This would allow to get rid of [this comment](https://github.com/fxbox/foxbox/pull/382/files#diff-cfe836b96d14aa0a2382eca5bb6a8e3eR59) too. What do you think?

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
       .then(function(listResponse) {
-        expect(listResponse.body.length).equals(3);
+        lightKeys = Object.keys(listResponse.body);
+        expect(lightKeys.length).equals(3);
```
JohanLorenzo: I wonder if we could be more explicit on why `lightKeys` must have a size of 3. How about:
```js
var numberOfBulbs = Object.keys(listResponse.body).length;
expect(numberOfBulbs).equals(3);
```

?

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
 
-      return chakram.get(serviceListURL)
+      // collect all getters for the lightbulbs
+      var getURL = channelURL + '/get';
+      var getterPayload = [{'kind': 'LightOn'}];
+      
+      return chakram.put(getURL,getterPayload)
       .then(function(listResponse) {
-        expect(listResponse.body.length).equals(3);
+        lightKeys = Object.keys(listResponse.body);
```
JohanLorenzo: I believe we can rename `lightKeys` into `lights`, then

``` js
expect(lights.length).equals(3);
```

would be easier to understand. The name would also be closer to what we manipulate.

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
+        expect(lightKeys.length).equals(3);
         expect(listResponse).to.have.status(200);
+
+        //check all lights are reported to be off
+        for (var i =0; i < lightKeys.length; i++){
```
JohanLorenzo: Nit: We can use a simpler loop:

``` js
lights.forEach(light => {
  expect(light.OnOff).equals('Off'); 
})
```

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
+      var getterPayload = [{'kind': 'LightOn'}];
+      
+      return chakram.put(getURL,getterPayload)
       .then(function(listResponse) {
-        expect(listResponse.body.length).equals(3);
+        lightKeys = Object.keys(listResponse.body);
+        expect(lightKeys.length).equals(3);
         expect(listResponse).to.have.status(200);
+
+        //check all lights are reported to be off
```

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
-        Prepper.philipshue_server.turnOffLight(2);
-        Prepper.philipshue_server.turnOffLight(3);
-
-        return chakram.get(serviceListURL)
-        .then(function(listResponse) {
-          lightinfo = listResponse.body;
-        });
-      });
+    // Currently, there is no mapping between the foxbox
+    // id and the philips hue id until the tag feature is implemented
```
JohanLorenzo: :+1: Nice, thanks for the background

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
+    it('Turn on all lights one by one', function(){
+      var setURL = channelURL + '/set';
+      var lightID = lightKeys[0].replace('getter','setter');
```
JohanLorenzo: :warning: This value depends on the previous test. I'd suggest to fetch it in a `before()`, instead.

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
-
-         // check no lights are turned off now
-         expect(Prepper.philipshue_server.lightStatus(1) &&
-          Prepper.philipshue_server.lightStatus(2) &&
-          Prepper.philipshue_server.lightStatus(3)).equals(true);
-       });
-      });
-});
-});
+        for (var i =0; i < lightKeys.length; i++){
```
JohanLorenzo: See [this comment](https://github.com/fxbox/foxbox/pull/382/files#r59844218)

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
+    // id and the philips hue id until the tag feature is implemented
+    it('Turn on all lights one by one', function(){
+      var setURL = channelURL + '/set';
+      var lightID = lightKeys[0].replace('getter','setter');
+      var payload = {'select': {'id': lightID}, 'value': { 'OnOff': 'On' } };
+      
+      return chakram.put(setURL,payload)
+      .then(function(cmdResponse) {
+       expect(cmdResponse).to.have.status(200);
+       expect(Object.keys(cmdResponse.body)[0]).equals(lightID);
```
JohanLorenzo: Maybe we could simplify this expectation with:

``` js
expect(cmdResponse.body[lightID]).to.be.defined;
```

?

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
-        });
-      });
+    // Currently, there is no mapping between the foxbox
+    // id and the philips hue id until the tag feature is implemented
+    it('Turn on all lights one by one', function(){
+      var setURL = channelURL + '/set';
+      var lightID = lightKeys[0].replace('getter','setter');
+      var payload = {'select': {'id': lightID}, 'value': { 'OnOff': 'On' } };
+      
+      return chakram.put(setURL,payload)
```
JohanLorenzo: _Improvement for a following PR_

Just like for [the getters](https://github.com/fxbox/foxbox/pull/382/files#r59843302), I think this call is at the same level of abstraction than the other functions in `FoxboxLights`, It could be something like:

``` js
return foxboxLights.turnOffLight(0)
.then(response => {
  // perform checks
  return foxboxLights.turnOffLight(1)
}) // and so on
```

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
+        expect(Object.keys(cmdResponse.body)[0]).equals(lightID);
+        lightID = lightKeys[2].replace('getter','setter');
+        payload = {'select': {'id': lightID}, 'value': { 'OnOff': 'On' } };
+        return chakram.put(setURL,payload);
+      })
+      .then(function(cmdResponse) {
+        expect(cmdResponse).to.have.status(200);
+        expect(Object.keys(cmdResponse.body)[0]).equals(lightID);
+        // check no lights are turned off now
+        expect(Prepper.philipshue_server.lightStatus(1) &&
```
JohanLorenzo: This logic is also [present here](https://github.com/fxbox/foxbox/pull/382/files#diff-cfe836b96d14aa0a2382eca5bb6a8e3eR51) and [there](https://github.com/fxbox/foxbox/pull/382/files#diff-cfe836b96d14aa0a2382eca5bb6a8e3eR131). I'd factorize it in `philipshue_server`. For example:

``` js
get areAllLightsOn() {
  return [1, 2, 3].every(i => this.lightStatus(i))
}
```

then in the test:

``` js
expect(Prepper.philipshue_server.areAllLightsOn).to.be.true;
```

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
-      });
-});
-});
+        for (var i =0; i < lightKeys.length; i++){
+          expect(listResponse.body[lightKeys[i]].OnOff).equals('On');  
+        }
+      }); 
+    });
+
+    it('Turn off all lights at once', function() {
```
JohanLorenzo: :warning: This test depends on the result of the previous one. I'd force the lights to be on (whatever the previous result) in a `before()`. What do you think?

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
+        Prepper.philipshue_server.lightStatus(2) &&
+        Prepper.philipshue_server.lightStatus(3)).equals(false);
+
+        //check all lights are reported to be off
+        var getURL = channelURL + '/get';
+        var getterPayload = [{'kind': 'LightOn'}];  
+        return chakram.put(getURL,getterPayload);
+      })
+      .then(function(listResponse) {
+        lightKeys = Object.keys(listResponse.body);
```
JohanLorenzo: Same comment about lights vs lightKeys

test/integration/test/philips_light_test.js
```diff
@@ -38,57 +35,117 @@ Prepper.makeSuite('Control lights locally',function(){
...
+        expect(lightKeys.length).equals(3);
+        expect(listResponse).to.have.status(200);
+
+        for (var i =0; i < lightKeys.length; i++){
```
JohanLorenzo: `forEach`

JohanLorenzo:
It's nice to have new tests to turn on and off all the lights! There are a some small nits plus 2 things which might make the tests hard to understand their failure (one depend on the results of the other). I'd recommend to fix the latter before landing.

test/integration/test/philips_light_test.js
nojunpark, line 54: yes, definitely refactoring is needed, and I was thinking about something that looks like WebUI testing.  I was thinking about doing it in this PR, but I thought I need to wait until Yoric's decentralized taxonomy is in place, since that'll definitely require me to do the rework.

test/integration/test/philips_light_test.js
nojunpark: true, instead of .to.be.defined, i think i can use equals(null)

test/integration/test/philips_light_test.js
nojunpark: correct, will do for the next PR.

test/integration/test/philips_light_test.js
nojunpark: good idea!

test/integration/test/philips_light_test.js
nojunpark: Actually, it doesn't.  it'll turn off regardless (if it is off, send the off command again) so it doesn't 'depend' on the previous test, although having it turned on before would make the test more meaningful. But I'm thinking even sending the off command to the already off test has some merit.

test/integration/test/philips_light_test.js
nojunpark: true, will do

nojunpark:
@JohanLorenzo r?  also, above travis failure seems to be cause by the openssl issue.  those three extra export statements should be added before each build i think.

test/integration/test/philips_light_test.js
```diff
@@ -47,48 +49,93 @@ Prepper.makeSuite('Control lights locally',function(){
...
+          expect(cmdResponse.body[lightID]).equals(null);
+          expect(Prepper.philipshue_server.areAllLightsOn()).to.be.true;
+          
+          //check all lights are reported to be on
+          return chakram.put(getURL,getterPayload);
+        })
+        .then(function(listResponse) {
+          lights = Object.keys(listResponse.body);
+          expect(lights.length).equals(3);
+          expect(listResponse).to.have.status(200);
```
JohanLorenzo: I just realized the 200 checks would be best done before we manipulate the body. For example, the server fails with a 50x: you'll get an error saying that there are no 3 lights, whereas the problem is more serious.

JohanLorenzo:
Looks good to me, with the follow ups planned. r+

PR #383, opened 2016-04-15
Extending the clock to support triggering events every n seconds

Yoric:


Yoric:
Ah, well, got the commit message wrong.

Issue #359, opened 2016-04-13
Set up page is not scaled properly in adaptable design view


isabelrios:
@ferjm, as per our conversation: 
Please see screenshot attached
<img width="318" alt="captura de pantalla 2016-04-13 a las 16 59 11" src="https://cloud.githubusercontent.com/assets/1897507/14498158/b178ff12-0199-11e6-8bc1-db3117b8158d.png">
Thanks!

ferjm:
https://github.com/ferjm/foxbox/commit/066a92e7ad4d31bdf662a7a9388d33fa0e959908

PR #384, opened 2016-04-15
Issue #359. Set up page is not scaled properly in adaptable design vi…

ferjm:
…ew. r=gmarty

r? @gmarty

### DIFF
--- a/static/shared/css/common.css
+++ b/static/shared/css/common.css
@@ -44,8 +44,15 @@ button {
 .container {
   position: relative;
   width: 70%;
-  padding: 20% 0 0;
   margin: auto;
+}
+
+.signin-container {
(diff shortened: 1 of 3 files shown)


static/shared/css/common.css
```diff
@@ -58,3 +65,24 @@ button {
...
+
+.signup-container {
+  padding: 2% 0 0;
+}
+
+input {
+  margin-top: 2rem;
+}
+
+}
```
gmarty, line 88: This `@media` section needs to be indented.

gmarty:
Looks good to me with a nit but Travis doesn't seem happy.

I think we have visual for the box login page, I'll dig them up and that'd be nice to integrate it post-demo.

PR #389, opened 2016-04-15
Add a check for group membership in the dialout and netdev groups.

dhylands:
Not needed for the demo.

I'm also tempted to say that this should be fatal, but made it non-fatal initially.

r? @fabricedesre


fabricedesre:
Let's make it fatal.

dhylands:
Unfortunately, if I make these fatal, then travis fails since its not a member of those groups.

fabricedesre:
We can do that in travis, like https://github.com/fxbox/foxbox/blob/master/tools/travis-linux.sh#L23

PR #391, opened 2016-04-15
Allow ZWave device to come from the configuration.

dhylands:
r? @julienw 

This needs https://github.com/fxbox/openzwave-adapter/pull/24

The user can either provide a command line option:

```
cargo run -- --disable-tls --config 'openzwave;device;/dev/ttyACM0'
```

or they can add an entry to foxbox.conf:

```
{
  "foxbox": {
    "certificate_directory": "certs/"
  },
  "philips_hue": {
    "nupnp_enabled": "true",
    "nupnp_url": "http://www.meethue.com/api/nupnp",
    "token_001788fffe230cda": "f0970f2995fb487eb3a7ec0c7760959c"
  },
  "openzwave": {
    "device": "/dev/ttyACM0"
  }
}
```

### DIFF
--- a/src/adapters/mod.rs
+++ b/src/adapters/mod.rs
@@ -69,7 +69,8 @@ impl<T: Controller> AdapterManager<T> {
         let scripts_path = &self.controller.get_profile().path_for("thinkerbell_scripts.sqlite");
         ThinkerbellAdapter::init(manager, scripts_path).unwrap(); // FIXME: no unwrap!
         let profile_openzwave = &self.controller.get_profile().path_for("openzwave");
-        OpenzwaveAdapter::init(manager, profile_openzwave).unwrap();
+        let openzwave_device = self.controller.clone().get_config().get("openzwave", "device");
+        OpenzwaveAdapter::init(manager, profile_openzwave, openzwave_device).unwrap();
 
         self.start_tts(manager);
     }


julienw:
Why not, but in the long run we'll need to have all of this exposed to the user through the web interface. And not only this: a lot of things are exposed through zwave adapters (for example: door lock codes) or other parameters to the zwave library (example: the secure network key).

I think we need something better and more generic than "just" parameters to the init function.

src/adapters/mod.rs
```diff
@@ -69,7 +69,8 @@ impl<T: Controller> AdapterManager<T> {
         let scripts_path = &self.controller.get_profile().path_for("thinkerbell_scripts.sqlite");
         ThinkerbellAdapter::init(manager, scripts_path).unwrap(); // FIXME: no unwrap!
         let profile_openzwave = &self.controller.get_profile().path_for("openzwave");
-        OpenzwaveAdapter::init(manager, profile_openzwave).unwrap();
+        let openzwave_device = self.controller.clone().get_config().get("openzwave", "device");
```
julienw, line 72: not sure why you clone ? It builds for me without the clone here

julienw:
r=me with the `clone()` call removed.

julienw:
> I think we need something better and more generic than "just" parameters to the init function.

Filed https://github.com/fxbox/RFC/issues/15

dhylands:
The motivation for this change was to allow people to have some way to use foxbox without having to edit:
https://github.com/fxbox/openzwave-stateful-rust/blob/master/src/lib.rs#L21-L29

julienw:
Yes I get it :)

PR #395, opened 2016-04-16
initial attempt to create a push server sim and tests

nojunpark:


package.json
```diff
@@ -10,7 +10,7 @@
     "test-integration": "npm run clean && ./node_modules/.bin/mocha test/integration/test/*.js && ./node_modules/.bin/mocha test/integration/test/no_travis/*.js",
     "test-integration-travis": "npm run clean && ./node_modules/.bin/mocha test/integration/test/*.js",
     "clean": "npm run clean_db && npm run clean_picture",
-    "clean_db": "rm -f ~/.local/share/foxbox/users_db.sqlite",
+    "clean_db": "rm -f ~/.local/share/foxbox/*.sqlite",
```
JohanLorenzo: I wonder if we should delete the whole folder, so we'd start with a clean and known state. What do you think?

package.json
```diff
@@ -38,5 +38,8 @@
     "config-js": "^1.1.9",
     "express": "^4.13.4",
     "iso-date": "^1.0.0"
+  },
+  "dependencies": {
+    "ecdh-es": "0.0.8"
```
JohanLorenzo: Nit: Unused dependency

package.json
Yoric: Just to be sure: will this be called only on Travis? Because otherwise, I'd loathe to lose all my local configuration just because I have run a single test.

test/integration/lib/foxboxHelper.js
```diff
@@ -12,9 +12,22 @@ var foxboxInstance;
 var helper = (function() {
 
   function removeUsersDB() {
-    var filePath = path.join(process.env.HOME,
-      '/.local/share/foxbox/users_db.sqlite');
-    fs.unlinkSync(filePath);
+    try {
```
JohanLorenzo: Nit: Duplication could be factorized in an internal function like `_removeFileIfItExists()`

test/integration/lib/webpush_server.js
```diff
@@ -0,0 +1,93 @@
...
+  var serverPublicKey = serverECDH.generateKeys('base64');
+  var _server = express();
+  
+  var port = config.get('webpush.port');
+  var endpoint = config.get('webpush.endpoint');
+  
+  var incomingPushMsg;
+  var serverInstance;
+
+  function getEndpointURI() {
```
JohanLorenzo, line 23: Nit: Javascript's `get` would be perfect for this getter. Like:

``` js
get endpointUri() {
  return config.get('webpush.ip') + ':' + port + endpoint;
}
```

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+const Config = require('config-js');
+
+var config = new Config('./test/integration/lib/config/foxbox.js');
+var Prepper = require('../lib/testPrepper.js');
+
+Prepper.makeSuite('Test Push Service locally',function(){
+  var header = new Config('./test/integration/lib/config/header.js');
+  var credential = config.get('credential'); 
+
+  var setupURL = config.get('foxbox.url') + '/users/setup';
```
JohanLorenzo: As the logic already exist in 3 other files, I'd factorize the URLs in a class.

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+  var loginURL = config.get('foxbox.url') + '/users/login';
+  var serviceURL = config.get('foxbox.url') + '/api/v1';
+  var serviceListURL = serviceURL + '/services'; 
+  var getterURL = serviceURL + '/channels/get';
+  var setterURL = serviceURL + '/channels/set';
+    
+  it('Create a new credential',function(){
```
JohanLorenzo: I don't recall exactly if I asked a similar question already. Why do we have to check the new credentials are created? Couldn't this be done in a `before()`?

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+    return chakram.post(setupURL,credential)
+    .then(function(setupResp){
+      expect(setupResp).to.have.status(201);
+    });
+  });
+
+  describe ('Once logged in', function () {
+    
+    before(function() {
+      var encoded_cred = new Buffer(credential.username+
```
JohanLorenzo: I believe we have to put the credential logic in one single place. If the logic changes, we'll end up with 4 different places to update.

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+      var pushService = false;
+
+      return chakram.get(serviceListURL)
+      .then(function(listResponse) {
+        for (var entry of listResponse.body) {
+          if (entry.adapter === 'webpush@link.mozilla.org') {
+              pushService = true;                
+           }
+        }
+        expect(listResponse).to.have.status(200);
```
JohanLorenzo: Nit: Move the 200 check before dealing with the body

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+        expect(pushService).equals(true);
+      });
+    });
+
+    it('Add push subscription',function(){     
+      var pushURI = Prepper.webPush_server.getEndpointURI();
+      var pushkey = Prepper.webPush_server.getPublicKey();
+      var setter = 'setter:subscribe.webpush@link.mozilla.org';
+      var payload = {
```
JohanLorenzo: Nit: Beautify this JSON

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+        'select': {'id':setter},
```
JohanLorenzo: Nit: Beautify this JSON

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+        'value': 
+          {'Json': 
+            {'subscriptions':[{
+              'push_uri':pushURI,
+              'public_key':pushkey}]}}};
+
+        return chakram.put(setterURL,payload)
+        .then(function(cmdResp){
+         expect(cmdResp).to.have.status(200);
+         expect(cmdResp.body[setter]).equals(null);
```
JohanLorenzo: Noob question: What does it tell us if `setter` is not null?

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+    });
+    
+    it('Verify push subscription',function(){   
+      var getter =  'getter:subscription.webpush@link.mozilla.org';
+      var payload = {'id': getter};    
+      
+      return chakram.put(getterURL,payload)
+      .then(function(cmdResp){
+        expect(cmdResp).to.have.status(200);
+        expect(cmdResp.body[getter].Json.subscriptions[0].public_key).equals(
```
JohanLorenzo: :+1: Nice check!

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+              'public_key':pushkey}]}}};
+
+        return chakram.put(setterURL,payload)
+        .then(function(cmdResp){
+         expect(cmdResp).to.have.status(200);
+         expect(cmdResp.body[setter]).equals(null);
+       });
+    });
+    
+    it('Verify push subscription',function(){
```
JohanLorenzo: I would add more details in the title, 'Add push subscription' also does some verifications

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+    });
+ 
+    it('Verify notification resources',function(){        
+      var getter =  'getter:resource.webpush@link.mozilla.org';
+      var payload = {'id': getter};    
+      
+      return chakram.put(getterURL,payload)
+      .then(function(cmdResp){
+        expect(cmdResp).to.have.status(200);
+        expect(cmdResp.body[getter].Json.resources[0]).equals('livingroom');
```
JohanLorenzo: This test depends on the previous one, doesn't it?

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+        return chakram.put(setterURL, payload)
+        .then(function(cmdResponse) {
+          // collect the response from the webpush simulator 
+          // after 100ms of (network)latency
+          setTimeout(done,100);
+        });
+      });
+
+      it('Check notification is received by Push server',function(){
+        var output = JSON.parse(Prepper.webPush_server.getDecodedPushMsg());
```
JohanLorenzo: :+1:

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,165 @@
...
+      
+      return chakram.put(getterURL,payload)
+      .then(function(cmdResp){
+        expect(cmdResp).to.have.status(200);
+        expect(cmdResp.body[getter].Json.resources[0]).equals('livingroom');
+        expect(cmdResp.body[getter].Json.resources[1]).equals('washroom');
+     });
+    });
+
+    it('Trigger notification',function(){
```
JohanLorenzo: I'm not sure I understood the test correctly but maybe we could make the expectation clearer. How about something like `accepts notification trigger requests`?

test/integration/lib/webpush_server.js
```diff
@@ -0,0 +1,93 @@
...
+      var sharedSecret = serverECDH.computeSecret(pkcut,'base64');
+      ece.saveKey('webpushKey', sharedSecret);
+        
+      // Decrypt the message using the shared secret and provided salt
+      var decrypted = ece.decrypt(encryptedBuffer, {
+              keyid: 'webpushKey',
+              salt: saltcut,
+              padSize: 1,
+            });
+      incomingPushMsg = decrypted.toString();
```
JohanLorenzo: This looks okay for now, but we might face race conditions if many responses are received almost at the same time. Maybe `incomingPushMsg` could be an array?

test/integration/lib/webpush_server.js
```diff
@@ -0,0 +1,93 @@
...
+        });
+    });
+  
+    // Handles the incoming ECDH encrypted webpush message from foxbox
+    _server.post(endpoint, function (req, res) {
+      assert.equal(req.headers['content-encoding'], 
+        'aesgcm128', 'Content-Encoding header correct');
+
+      // Collect salt, publickey, and the message
+      var saltcut = req.headers.encryption.replace(/keyid=p256dh;salt=/g,'');
```
JohanLorenzo: What does `cut` mean in this context?

package.json
JohanLorenzo: The effect will happen locally if you run `npm run clean_db` or `npm run test-integration`

package.json
```diff
@@ -38,5 +38,8 @@
     "config-js": "^1.1.9",
     "express": "^4.13.4",
     "iso-date": "^1.0.0"
```
JohanLorenzo: Missing dependency: raw-body https://travis-ci.org/fxbox/foxbox/jobs/124315439#L929

This job used to be marked as passed, but it was actually due to an error on Jenkins side: they rebooted the VM

JohanLorenzo:
Nice tests! I added some comments, but I think we're not too far from merging them in tree :smiley:

package.json
nojunpark: yes, the config will be removed for this particular instance of the travis run upon calling above mentioned npm run commands.

test/integration/lib/webpush_server.js
nojunpark, line 23: but I don't think it works this case, because I am in an anonymous function, and not in a var.

test/integration/test/webpush_test.js
nojunpark: if it's not null, it will come with an error (most likely a syntax error).  If everything is correct, null is returned.

test/integration/test/ip_camera_test.js
```diff
@@ -8,103 +8,66 @@ var config = new Config('./test/integration/lib/config/foxbox.js');
...
-  var setupURL = config.get('foxbox.url') + '/users/setup';
-  var loginURL = config.get('foxbox.url') + '/users/login';
-  var serviceURL = config.get('foxbox.url') + '/api/v1';
-  var serviceListURL = serviceURL + '/services'; 
-  var getterURL = serviceURL + '/channels/get';
-  var setterURL = serviceURL + '/channels/set';
   var cameraService;
 
-  it('create a new credential',function(){
+  before(Prepper.foxboxManager.foxboxLogin);
```
JohanLorenzo, line 14: Nice!

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,130 @@
...
+      'select': {
+        'id':setter
+      }, 
+      'value': {
+        'Json': {
+          'subscriptions':[{
+            'push_uri':pushURI,
+            'public_key':pushkey
+          }]}}};
+          var getterPayload = {'id': getter};
```
JohanLorenzo: Nit: Confusing indentation

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,130 @@
...
+  it('Set resources to receive notification',function(){     
+    var getter =  'getter:resource.webpush@link.mozilla.org';   
+    var setter = 'setter:resource.webpush@link.mozilla.org';
+    var setterPayload = {
+      'select': {
+        'id': setter}, 
+        'value': {
+          'Json': {
+            'resources':['livingroom', 'washroom']}}};
+            var getterPayload = {'id': getter};
```
JohanLorenzo: Nit: Confusing indentation

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,130 @@
...
+    var resource = 'washroom';
+    var notificationText = 'lights on!';
+    var payload = {
+      'select': {
+        'id': setter
+      }, 
+      'value': {
+        'WebPushNotify': {
+          'resource':resource,'message':notificationText}}} ;
+          return chakram.put(Prepper.foxboxManager.setterURL, payload)
```
JohanLorenzo: Nit: Confusing indentation

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,130 @@
...
+    before(function(done) {
+      var setter = 'setter:notify.webpush@link.mozilla.org';
+      var payload = {
+        'select': {
+          'id': setter
+        }, 
+        'value': {
+          'WebPushNotify': {
+            'resource':resource,'message':notificationText}}} ;
+            return chakram.put(Prepper.foxboxManager.setterURL, payload)
```
JohanLorenzo: Nit: Confusing indentation

test/integration/lib/webpush_server.js
JohanLorenzo, line 23: Oh, that's a drawback of anonymous functions, then. I also see the function reexport that is a type of duplication (each time someones requires to add a function, he/she has to it at 2 different places).

Let's keep it as is for now, but that sounds like a good refactor to do. What do you think?

test/integration/test/webpush_test.js
```diff
@@ -0,0 +1,130 @@
...
+          'subscriptions':[{
+            'push_uri':pushURI,
+            'public_key':pushkey
+          }]}}};
+          var getterPayload = {'id': getter};    
+
+          return chakram.put(Prepper.foxboxManager.setterURL,setterPayload)
+          .then(function(cmdResp){
+           expect(cmdResp).to.have.status(200);
+           expect(cmdResp.body[setter]).equals(null);
```
JohanLorenzo: Following up on the [null discussion](https://github.com/fxbox/foxbox/pull/395#discussion_r60411373): I think it would be good to specify why we check the null state in a comment

JohanLorenzo:
Looks good to me modulo a few nits!

r+

test/integration/lib/webpush_server.js
nojunpark, line 23: sure, I'm all open to proper refactoring

test/integration/test/webpush_test.js
nojunpark: ugh, I shouldn't trust sublime text's reindent feature anymore.

PR #376, opened 2016-04-14
[WIP] Standalone executable for the LetsEncrypt DNS-01 challenge.

samgiles:
Additionally updates `letsencrypt.sh`

PR #399, opened 2016-04-18
Separate dnschallenge executable

samgiles:
Same as #376 

Testin' Travis.


samgiles:
r? @fabricedesre

It seems the OpenSSL link failures on Mac OSX are intermittent - although it doesn't affect this PR :( - maybe because of cached binaries?

fabricedesre:
Any reason for the dnschallenge program to not be in its own crate (even in the same repo) ? It looks self contained enough.

samgiles:
That would be nice, especially if it was in the same repo - the reason it's not here is because the `tls` stuff isn't in a separate crate which it depends on.  It would be cool to move that out too, maybe to a `net` crate.  I could give this a go for this PR.

samgiles:
@fabricedesre how does this look?

@JohanLorenzo  Here I've made a components directory with separate crates, and updated the test scripts to run the tests in each components directory.  I've also added to build.rs so that the dns challenge binary is built when running `cargo build`.  How does this look?

README.md
```diff
@@ -91,6 +91,19 @@ default = []
...
+Some are third party, others are built as components found in the `components`
+directory.
+
+| Dependency     | Optional?                                      | Where to find it                                                       |
+| -------------- | ---------------------------------------------- | ---------------------------------------------------------------------- |
+| `dnschallenge` | No (required for LetsEncrypt DNS-01 challenge) | Built as a binary with `cargo build`, see `components/dns_challenge/target/<profile>` directory |
+| `bash`         | No (required for LetsEncrypt client)           | System package manager                                                 |
+| `python`       | Yes (required to run `pagekite.py`)            | System package manager                                                 |
+| `pagekite.py`  | Yes (required to enable tunneling)             | https://pagekite.net/wiki/OpenSource/                                  |
+
```
fabricedesre, line 106: nit: not sure if this is because of the diff, but we may have a alignment issues in this table. My eyes are bleeding.

components/dns_challenge/main.rs
```diff
@@ -0,0 +1,72 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#![allow(dead_code)]
+
+// Needed to derive `Serialize` on ServiceProperties
```
fabricedesre: nit: invalid comment. (we need to update the main src/main.rs too)

components/dns_challenge/main.rs
```diff
@@ -0,0 +1,72 @@
...
+#![feature(custom_derive, plugin)]
+#![plugin(serde_macros)]
+
+// Make linter fail for every warning
+#![plugin(clippy)]
+#![deny(clippy)]
+// Needed for many #[derive(...)] macros
+#![allow(used_underscore_binding)]
+
+#![feature(associated_consts)]
```
fabricedesre: can you double check what you need exactly? It looks like `#![allow(used_underscore_binding)]` and `#![feature(associated_consts)]` may not be needed

components/dns_challenge/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[dependencies]
+clippy = "0.0.63"
+hyper = "0.7.2"
+log = "0.3"
+mktemp = "0.1.2"
+openssl = "0.7.6"
+openssl-sys = "0.7.6"
+serde = "0.7.0"
+serde_json = "0.7.0"
+serde_macros = "0.7.2"
```
fabricedesre: why do we need serde?

components/tls/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+[dependencies]
+clippy = "0.0.63"
+hyper = "0.7.2"
+log = "0.3"
+mktemp = "0.1.2"
+openssl = "0.7.6"
+openssl-sys = "0.7.6"
+serde = "0.7.0"
+serde_json = "0.7.0"
+serde_macros = "0.7.2"
```
fabricedesre: same here: do we need serde?

fabricedesre:
I haven't tested, but overall I like it! @JohanLorenzo that would mean that thinkerbell would move to components/thinkerbell too.

tools/travis-linux.sh
```diff
@@ -43,7 +43,16 @@ set_up_tests() {
 run_tests() {
     npm run test-selenium
     npm run test-integration-travis
+
     # Note: Cargo recompiles every dependency with dead code. That's why
     # this step is currently the last
+
+    echo "run_tests: Running crate tests..."
```
JohanLorenzo: FYI: I [centralized that logic](https://github.com/fxbox/foxbox/pull/406/files#diff-f4983c80cbd9b1e076492899a26bd4d8R11) in #406. 

As rebasing #406 might be tricky (due to commits coming from another repo), would you mind waiting on that PR to land?

build.rs
```diff
@@ -44,7 +46,21 @@ fn copy_shared_static_files() {
...
+            .current_dir(run_in_dir)
+            .spawn()
+            .unwrap()
+            .wait()
+}
+
 fn main() {
     update_local_git_hook();
     copy_shared_static_files();
+    cargo_build_in_directory("./components/dns_challenge").unwrap();
```
JohanLorenzo, line 65: LGTM

tools/travis-linux.sh
```diff
@@ -43,7 +43,16 @@ set_up_tests() {
...
+    for DIR in $(ls -d components/*); do
```
JohanLorenzo: It's usually a [bad practice](http://mywiki.wooledge.org/BashGuide/Practices#Don.27t_Ever_Do_These) to use the output of `ls` mainly because of spaces in file names. `find` would be more reliable

samgiles:
@fabricedesre Turns out most of the stuff isn't needed anymore in the `dnschallenge` executable - only the `tls` crate. 
The formatting was a little off on one line in the README, a `|` wasn't extended to the edge of the table.  But, yeh, the diff looks horrible because of wrapping, it looks alright in an editor and renders the right thing.

Ok to merge?

fabricedesre:
Thanks, go ahead!

PR #400, opened 2016-04-18
Call taxonomy's AdapterManager::stop method, so that adapters' stop m…

### LINKED ISSUE #387: Call all adapters' stop method when foxbox stops
Currently we only call `AdapterManager.stop()` which does nothing. Because the Adapter trait provides an default empty `stop` method, we could automatically call all adapters' stop methods.

Note: it's not easy to call the adapters' stop methods from AdapterManager because we don't have an easy access to the adapter objects.

julienw:
…ethods are called as well (fixes #387)

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -100,6 +100,7 @@ impl Controller for FoxBox {
 
         debug!("Stopping controller");
         adapter_manager.stop();
+        taxo_manager.stop();
     }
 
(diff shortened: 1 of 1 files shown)


julienw:
I checked this works for me: there is a new log when we call openzwave-adapter's stop method :)

Yoric:
Code is too complicated.

Nah, just kidding, r=me :)

PR #371, opened 2016-04-14
README: Document "Failed set host name" error

JohanLorenzo:
Fixes #363 

Follow up #281 cc @samgiles 

r? @azasypkin @dhylands

Issue #363, opened 2016-04-13
Sometimes, foxbox needs to run as root.


dhylands:
Today, when I tried to run foxbox, (I tried with and without --disable-tls and also tried removing the certs directory, all with the same effect):

```
2080 >RUST_BACKTRACE=1 cargo run -- --disable-tls
     Running `target/debug/foxbox --disable-tls`
thread '<main>' panicked at 'Failed set host name: Access denied (code -20)', /home/dhylands/.multirust/toolchains/nightly-2016-04-10/cargo/git/checkouts/multicast-dns-d51af2fa146824d6/a6e4bcc/src/adapters/avahi/adapter.rs:346
stack backtrace:
   1:     0x7ff3d349fe80 - std::sys::backtrace::tracing::imp::write::h714760a4c8c0cdd8
   2:     0x7ff3d34a5fab - std::panicking::default_hook::_$u7b$$u7b$closure$u7d$$u7d$::hff309ab1d83ffd90
   3:     0x7ff3d34a5c33 - std::panicking::default_hook::h08ad3bb09872855b
   4:     0x7ff3d348e17f - std::sys_common::unwind::begin_unwind_inner::h406d5f1a330b854b
   5:     0x7ff3d348f1c8 - std::sys_common::unwind::begin_unwind_fmt::h57ea3fbee1a40196
   6:     0x7ff3d2d49f38 - _<adapters..avahi..adapter..AvahiAdapter as adapters..adapter..HostAdapter>::set_name::h2bed60aa1ab07b54
                        at /home/work/fxbox/foxbox/<std macros>:8
   7:     0x7ff3d2d4ec0f - multicast_dns::host::host_manager::HostManager::set_name::h8b6b157855e59ff3
                        at /home/dhylands/.multirust/toolchains/nightly-2016-04-10/cargo/git/checkouts/multicast-dns-d51af2fa146824d6/a6e4bcc/src/host/host_manager.rs:21
   8:     0x7ff3d2772fa6 - foxbox::update_hostname::hc21239e54baa1f82
                        at src/main.rs:167
   9:     0x7ff3d27737ce - foxbox::main::h000728ee4216cb83
                        at src/main.rs:241
  10:     0x7ff3d34a5874 - std::sys_common::unwind::try::try_fn::hfe4b14ce00a3f42f
  11:     0x7ff3d349f05b - __rust_try
  12:     0x7ff3d34a52bb - std::rt::lang_start::h12d9b3310b48a726
[...]

PR #402, opened 2016-04-19
Add Arch Linux groups in membership checks

JohanLorenzo:
Follow up #389 

r? @dhylands

### DIFF
--- a/src/permissions.rs
+++ b/src/permissions.rs
@@ -1,0 +1,63 @@
+use users::{ Groups, Users };
+use users::os::unix::GroupExt;
+
+pub fn check_group_membership<U: Users+ Groups>(cache: &U, expected_group_names: &[&str]) {
+    let mut is_one_group_found = false;
+    let mut i = 0;
+    let expected_group_size = expected_group_names.len();
+
+    while !is_one_group_found && i < expected_group_size {
+        let group_name = expected_group_names[i];
+
+        if let Some(group) = cache.get_group_by_name(group_name) {
+            is_one_group_found = true;
+            let current_uid = cache.get_current_uid();
+            let user = cache.get_user_by_uid(current_uid).unwrap();
+
+            if !group.members().contains(&user.name().to_owned()) {
+                panic!("Not a member of the {} group", group_name);
+            }
+        }
+
+        i += 1;
+    }
+
+    if !is_one_group_found {
+        panic!("None of the groups {:?} is defined in /etc/group. We likely don't support your \
+        Linux distribution. Please file a bug.", expected_group_names);
+    }
+}
+
+
+#[cfg(test)]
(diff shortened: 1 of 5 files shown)


src/permissions.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        if let Some(group) = cache.get_group_by_name(group_name) {
+            is_one_group_found = true;
+            let current_uid = cache.get_current_uid();
+            let user = cache.get_user_by_uid(current_uid).unwrap();
+
+            if !group.members().contains(&user.name().to_owned()) {
+                panic!("Not a member of the {} group", group_name);
+            }
+        }
```
dhylands, line 22: I don't like this. This is checking whether the current user is in the group, not whether the current process is in the group. So this test can pass (user is a member of the group) when the process is not. For example, if you edited the /etc/group file to add the user, then this would pass even though the shell which did the editing isn't yet a member.

libc::getgroups returns the groups that this process is currently a member of. So I think we should go back to using that and then check each of those groups to see if its name matches one of the ones passed in.

uid shouldn't enter into the equation anywhere (and if it did we should be using geteuid and not getuid).

This is a subtle distinction, but one that I think is quite important.

dhylands:
In general, Ilike the changes. But r- for removing the call to get_groups.

src/permissions.rs
```diff
@@ -0,0 +1,63 @@
...
+    while !is_one_group_found && i < expected_group_size {
+        let group_name = expected_group_names[i];
+
+        if let Some(group) = cache.get_group_by_name(group_name) {
+            is_one_group_found = true;
+            let current_uid = cache.get_current_uid();
+            let user = cache.get_user_by_uid(current_uid).unwrap();
+
+            if !group.members().contains(&user.name().to_owned()) {
+                panic!("Not a member of the {} group", group_name);
```
dhylands, line 20: This should probably show all of the group names passed in (i.e. all of expected_group_names)

dhylands:
@fabricedesre and I talked about this today, and we think that we should scrap the permission check and instead put in better error reporting.

The original problem started out from https://github.com/fxbox/foxbox/issues/363

So if we instead catch the permission failure and report an error message that suggests making the user a member of netdev (or network) that that would be the preferable way to go.

We can do something similar in the zwave adapter and suggest dialout/uucp if we get a permission error there.

src/permissions.rs
JohanLorenzo, line 22: Thanks for the background. I agree with you this would have been a major issue if we wanted to keep the PR.

JohanLorenzo:
@dhylands I agree with you, it's better to suggest than to hard-code the enforcement. On a short term perspective, we could document that in the README. I made a proposal at #371. What do you think?

PR #406, opened 2016-04-20
Add thinkerbell

JohanLorenzo:


JohanLorenzo:
Here are the files with history preserved. This will help people when `git log` or `git blame` are needed.

[The procedure used](https://gist.github.com/bxt/9746062) is based on `git filter-branch`. That command was necessary to rewrite the history of Thinkerbell as if all the files were committed in `$FOXBOX_HOME/components/thinkerbell`. That path is similar to the one [defined in servo](https://github.com/servo/servo/tree/master/components).

Following up on the discussion with @julienw this morning, we can't pull directly commits from another repo. Otherwise, git tries to merge common files (like `Cargo.toml`, `.gitignore`).

Then another `filter-branch` was done to prefix every commit message with the original repo name.

One side effect of `filter-branch`: commits don't have the same sha1 as they had in fxbox/thinkerbell. These could be an issue if people mentioned commits in Github issues. This seems like an edge case compared to the benefit of having the whole log and the correct authors.

As a consequence of different sha1s, commits authors will see their Github contributions growing "for free".

Reviews suggestions:
- Look at the 2 last commits. They contain the changes that made the local files work with each others.
- Observe the history

Incidently, I would also suggest to perform **a regular merge** of the PR, instead of squashing the commits :smiley:

r? @fabricedesre @Yoric

julienw:
> Following up on the discussion with @julienw this morning, we can't pull directly commits from another repo. Otherwise, git tries to merge common files (like Cargo.toml, .gitignore).

Here is a suggestion (I don't know if this works):

```
git remote add thinkerbell <thinkerbell_url>
git checkout -b thinkerbell_to_merge thinkerbell/master
mkdir thinkerbell && git mv * thinkerbell # likely you'll need to specify all files instead of using *
git commit
git push <your_remote> HEAD
### create pull request in github 
```

julienw:
So we tried the suggestion from https://github.com/fxbox/foxbox/pull/406#issuecomment-212499752 and it technically works. That said, after discussing about it I now think the first approach probably makes more sense.

JohanLorenzo:
> I now think the first approach probably makes more sense.

The main argument was in favor of prefixing commit messages. It solves these 2 problems:
- What if somebody checks out in a commit from thinkerbell and can't build foxbox anymore?
- `git bisect`: What if you end up on the thinkerbell repo? How would a developer new to the project would know that was not supposed to happen?

julienw:
both commit messages and moving everything to a subdirectory in the whole history

fabricedesre:
lgtm. I'm gonna re-enable merge commits just for this one time ;)

Yoric:
r=me

PR #411, opened 2016-04-22
Bump iron-fork to 0.3.0 (and remove .cargo)

### LINKED ISSUE #367: It seems that .cargo is checked into git and it shouldn't be.
It means that we can't use it to point other stuff locally without running into problems with git.

Shouldn't we just specify in the Cargo.toml file that it's using a local fork?

JohanLorenzo:
Fixes #367


JohanLorenzo:
Upgrading iron-fork to 0.3.0 turned out to be easier than I thought. The only changes made since we forked were located in Cargo.toml (and the ones made for SNI in [iron-fork/src/iron.rs](https://github.com/fxbox/foxbox/commit/b35592c2b13511bec8c071852bf6c9cab3bdbc71#diff-8ee9cd9df532fc0b2e1b54cabee75128)).

r? @samgiles

fabricedesre:
@samgiles any hope to upstream your changes?

samgiles:
@fabricedesre Not as it is right now - there are 4 different implementations to achieve the same thing, I have an API suggestion completely different from my implementtion on my PR from a maintainer.  I just need to do that - and then _hopefully_ we'll get a merge.

samgiles:
@JohanLorenzo lgtm! Thanks.

JohanLorenzo:
Rebased on top of master. I had to also specify the "replace" in subcrates. 

I also had to fork ws-rs, in order to have the mio revisions matching https://github.com/JohanLorenzo/ws-rs/blob/0.4.6/Cargo.toml#L15

Cargo.toml
```diff
@@ -46,25 +52,24 @@ rusqlite = "0.6.0"
...
+staticfile = "0.2.0"
 transformable_channels = "^0.1"
 unicase = "1.3.0"
-tls = { path = "./components/tls" }
 time = "0.1"
 timer = "0.1.6"
 uuid = "0.1.18"
 url = "0.5.7"
-ws = { git = "https://github.com/housleyjk/ws-rs.git", rev = "d154fc5" }
+ws = { git = "https://github.com/JohanLorenzo/ws-rs.git" , branch = "0.4.6" }
```
JohanLorenzo, line 62: Needed, otherwise mio version mismatch. 
- https://crates.io/crates/ws/0.4.6 uses mio 0.5.0 
- whereas we use 0.5.0, but with some revisions forward https://github.com/fxbox/foxbox/pull/411/files#diff-80398c5faae3c069e4e6aa2ed11b28c0R42

PR #412, opened 2016-04-22
Fix `cargo update` error when compiling ws

JohanLorenzo:
Like @isabelrios discovered this morning, `cargo update` breaks today's build.

```
/home/jlorenzo/.multirust/toolchains/nightly-2016-04-10/cargo/git/checkouts/ws-rs-012dd83a96dba3eb/d154fc5/src/handshake.rs:339:21: 339:35 error: no method named `serialize_path` found for type `&url::Url` in the current scope
/home/jlorenzo/.multirust/toolchains/nightly-2016-04-10/cargo/git/checkouts/ws-rs-012dd83a96dba3eb/d154fc5/src/handshake.rs:339                 url.serialize_path().unwrap_or("/".to_owned()),
```

After an investigation with @ferjm, we found out that `url` is now compiled in 3 different versions:  0.2.38, 0.5.9 and 1.0.0. The first 2 versions have been here so far, so the culprit is 1.0.0.

`ws` is not meant to be built against a [version > 0.5](https://github.com/housleyjk/ws-rs/blob/8827a6da17f9ec8d25eb7b31e3640d27667d8919/Cargo.toml#L18). But it turns out only a few commits pinpoint version. [d154fc5](https://github.com/housleyjk/ws-rs/blob/d154fc56e8dd06d9172efd08ef0db9c244668cbf/Cargo.toml#L20) (the revision we picked) was one of them. 

r? @isabelrios

**Edit:** The build failed on Travis because it requires #411 to land first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 name = "foxbox"
 version = "0.1.0"
 dependencies = [
- "chrono 0.2.21 (registry+https://github.com/rust-lang/crates.io-index)",
+ "chrono 0.2.22 (registry+https://github.com/rust-lang/crates.io-index)",
  "clippy 0.0.63 (registry+https://github.com/rust-lang/crates.io-index)",
  "docopt 0.6.80 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
Actually done in #411

PR #413, opened 2016-04-22
Travis: Add optional job to forecast `cargo update` failures

### LINKED ISSUE #409: Forecast `cargo update` failures
Like @isabelrios discovered this morning, `cargo update` breaks today's build. **Edit:** This is due to one deep dependency that was upgraded not to a backward-compatible version (see #412 for more details)

We might face some similar breakages in the future. @samgiles proposed to make a new travis job that will be in charge of compiling foxbox against the most recent `cargo update`. Like for the most recent rustc nightly job, failures will be allowed.

JohanLorenzo:
For reference:
- Example of [failed job](https://travis-ci.org/fxbox/foxbox/jobs/125009235)

Fixes #409. Depends on #411

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,12 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
-    - sudo: required
-      os: linux
+    - os: linux
       rust: nightly-2016-04-10
       env: BUILD_ENV=-arm_cross_compile
+    - os: linux
+      rust: nightly
+      env: BUILD_ENV=-cargo_update
 
 addons:
   firefox: latest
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -21,10 +21,12 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
-    - sudo: required
-      os: linux
+    - os: linux
       rust: nightly-2016-04-10
       env: BUILD_ENV=-arm_cross_compile
+    - os: linux
+      rust: nightly
```
JohanLorenzo, line 28: I wish we could build against the pinned version of nightly. But we can allow failures only on [groups of jobs](https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail).

fabricedesre:
r=me

PR #415, opened 2016-04-23
Build the dns_challenge crate properly

fabricedesre:
There are a couple of issues with the current build.rs code:
- we always build a default, `debug` version of the dns_challenge crate.
- we never cross-compile it, so we fail totally on the target hardware.

This patch addresses these issues.

Further work is needed to get `cargo clean` to also cleanup components/*

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -4,9 +4,8 @@
 
 use std::env;
 use std::fs;
-use std::io::Result as IoResult;
 use std::path::{ Path, PathBuf };
-use std::process::{ Command, ExitStatus };
+use std::process::Command;
 
 fn update_local_git_hook() {
     let p = env::current_dir().unwrap();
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -46,21 +49,59 @@ fn copy_shared_static_files() {
...
+    if get_env("PROFILE") == "release" {
+        let tmp = command;
+        command = tmp.arg("--release");
+    }
+
+    let output = command.current_dir(run_in_dir)
+                        .output()
+                        .unwrap();
+
+    if !output.status.success() {
```
samgiles: Why display the output only if the command was successful?
It might be useful to see the output of this command regardless of success state, by default, stdin, stdout and stderr are inherited from the parent with `spawn` on a `Command`.

samgiles:
Nice, other than the question this LGTM.

PR #416, opened 2016-04-25
[WIP] Split travis jobs into faster ones

### LINKED ISSUE #410: Make travis jobs faster
With the dependencies being in the same repo, Travis results might take an hour overall. Time is split into 2 parts:
- Time for travis to find a free VM - during [peak hours](https://www.traviscistatus.com/#week) (from 7am to 10am Pacific Time), it can take up to 30 minutes.
- Time to actually do things - about 15 minutes for the best job (OS X) and 30 minutes for the worst one (Linux with all steps).

I see 2 ways of tackling these issues.
#1. Prevent multiple builds in the same job

In the Linux job particularly, we build foxbox or one of its dependencies many times:
1. when the first `cargo build` is run
2. when `cargo test` is executed in foxbox
   1. each time `cargo test` runs in a subcrate => We build every dependency one more time. This a missing feature in cargo. See https://github.com/rust-lang/rfcs/pull/1525
3. when we run `cargo test` with dead code => every single dependency gets recompiled with no dead code optimization.

In the cross-compile job, we also build once in release mode the other time in debug.

I would recommend to put each of these different builds in self-contained jobs. Moreover, if subcrates are compiled an tested in their own job, it would allow us to deal with publishing on crates.io in that same job. 
#2. Run on the container based infrastructure whenever we can

That infrastructure is faster to spawn and has more available machine than the VM infra. Using them would be ideal.

However, most of our cases requires us to run in a VM:
- When we [run on Mac](https://docs.travis-ci.com/user/workers/os-x-infrastructure/).
- When we can't install a package via the travis apt addon:
  - either because we need to [edit sources.list ourselves](https://github.com/fxbox/foxbox/blob/afbd8726f9f736cf52feff2827446e70230c5c0c/tools/travis-linux-arm_cross_compile.sh#L6)
  - or because these packaged are [not supported by the addon](https://github.com/fxbox/foxbox/blob/bc172a37a89eae9b8b279d1d636bbe75863c9b21/tools/travis-linux.sh#L11) (yet?)

Hence, the only cases I see we could use the container would be the `cargo test` in subcrates.

JohanLorenzo:
Fixes #410


JohanLorenzo:
The time to get machine spawned hasn't improved. Closing this PR per the comments left in #410.

PR #421, opened 2016-04-25
Extract the Watcher code into its own file

julienw:


components/openzwave-adapter/src/watchers.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use transformable_channels::mpsc::ExtSender;
+
+use std::collections::HashMap;
+use std::sync::{ Arc, Mutex, Weak };
+
+pub type SyncSender = Mutex<Box<ExtSender<WatchEvent>>>;
+type WatchersMap = HashMap<usize, Arc<SyncSender>>;
+type RangedWeakSender = (Option<Range>, Weak<SyncSender>);
+pub type RangedSyncSender = (Option<Range>, Arc<SyncSender>);
```
julienw, line 14: I'm sure it's possible to make all this more generic.

julienw:
@dhylands what do you think ? It's mostly moving code around and adding `pub` in some places. I'm sure we could make this more generic and avoid importing transformable_channels etc in watchers.rs, but I'd like to keep this for the future. Right now I'd like to de-densify the main lib.rs.

julienw:
@dhylands r?

components/openzwave-adapter/src/watchers.rs
```diff
@@ -0,0 +1,76 @@
...
+            map: self.map.clone()
+        }
+    }
+
+    fn get(&self, index: usize) -> Option<Arc<SyncSender>> {
+        let map = self.map.lock().unwrap();
+        map.get(&index).cloned()
+    }
+
+    pub fn get_from_tax_id(&self, tax_id: &TaxId<Getter>) -> Option<Vec<RangedSyncSender>> {
```
dhylands: nit: I noticed that here Tax is used an an abbreviation for taxonomy and other places use Taxo. We should pick one or the other and use it consistently through the code.

dhylands:
r+

components/openzwave-adapter/src/watchers.rs
julienw: Yeah, you're right, I can do the change here as well :) I used to use `tax` but I find this really confusing and that's why I started changing it. I can do a separate PR to change this in the main lib.rs as well.

julienw:
@dhylands I added a second commit to rename all Tax to Taxo, I'd like a quick second look. This builds so I think I didn't forget anything :)

dhylands:
lgtm - r+

PR #426, opened 2016-04-25
Switch the /ping endpoint to return 204 instead of 200

### LINKED ISSUE #422: [Polish] Return a HTTP status code of 204 on /ping entry point
The `/ping` entry point doesn't need a body response.
The most appropriate HTTP status code in this case is `204 No Content`.

Here's a quick description of this code:

```
204 No Content

The server has fulfilled the request but does not need to return an entity-body, 
and might want to return updated metainformation. The response MAY include new 
or updated metainformation in the form of entity-headers, which if present 
SHOULD be associated with the requested variant.
```

fabricedesre:
Fixes #422 

r? @ferjm

### DIFF
--- a/src/http_server.rs
+++ b/src/http_server.rs
@@ -51,7 +51,7 @@ struct Ping;
 
 impl Handler for Ping {
     fn handle (&self, _: &mut Request) -> IronResult<Response> {
-        Ok(Response::with(Status::Ok))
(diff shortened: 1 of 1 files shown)


fabricedesre:
I got tired re-triggering the Mac OS build "intermittent" failure.

ferjm:
LGTM

PR #427, opened 2016-04-25
first try for test to open the webapp

isabelrios:


isabelrios:
Hi @JohanLorenzo, travis is green but the coverage has decreased, not sure what should I do. r?

test/selenium/lib/setup_webapp.js
```diff
@@ -14,6 +15,10 @@ SetUpWebapp.prototype = {
 
     getSetUpView : function() {
         return new SetUpView(this.driver);
+    },
+
+    getAppMainView : function() {
+    	return new MainView(this.driver);
```
JohanLorenzo: Nit: misleading indentation

test/selenium/web_app_test.js
```diff
@@ -0,0 +1,48 @@
+
+/* Any copyright is dedicated to the Public Domain.
```
JohanLorenzo: Nit: We follow the MPL licence

test/selenium/web_app_test.js
```diff
@@ -0,0 +1,48 @@
...
+
+var SetUpWebapp = require('./lib/setup_webapp.js');
+var webAppMainPage;
+var setUpWebapp;
+
+describe('UI app', function() {
+  var driver;
+  this.timeout(8000);
+
+  //const PASS = '12345678';
```
JohanLorenzo: Nit: Remove commented code

test/selenium/web_app_test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  after(function() {
+    driver.quit();
+  });
+     
+  describe('open the web app', function() {
+
+    beforeEach(function() {
+          setUpWebapp = new SetUpWebapp(driver);
```
JohanLorenzo: Nit: misleading indentation over the next 3 lines

test/selenium/web_app_test.js
```diff
@@ -0,0 +1,48 @@
...
+          webAppMainPage = setUpWebapp.getAppMainView();
+          return webAppMainPage;
+        });
+
+    it('should log in from web app', function() {
+      return webAppMainPage.connectToFoxBox().then((logInView) => 
+        logInView.loginSuccess(12345678));
```
JohanLorenzo: Where is the password first set? I would explicitely set it in a `before()`

isabelrios: It is set in the sessions_ui_test.js.  Both tests are dependant, which is not the best, first the set up and then the log in in the app. Thought about other options but not sure which is the best to proceed without modifing all the tests we already have

JohanLorenzo:
The coverage decrease is probably due to master. It happens when your branch is not up to date compared to master.

The patch looks good to me overall. There is one issue that seems blocking: the password looks set somewhere else. Apart from that, everything is good

test/selenium/web_app_test.js
JohanLorenzo: For reference @npark-mozilla  made some similar setup steps for the integration tests. One way to do is to [delete the sqlite files](https://github.com/fxbox/foxbox/blob/master/test/integration/lib/foxboxHelper.js#L33) and [start the foxbox process](https://github.com/fxbox/foxbox/blob/master/test/integration/lib/foxboxHelper.js#L40) one more time.

test/selenium/lib/setup_webapp.js
```diff
@@ -2,19 +2,68 @@
...
+        console.log('start new foxbox');
+        foxboxInstance = spawn('./target/debug/foxbox',
+        ['-c',
+        '--disable-tls']
+        );
+      setTimeout(callback, FOXBOX_STARTUP_WAIT_TIME_IN_MS);
+      return;
+    },
+
+    killFoxBox: function () {
```
nojunpark: Hmm, here, you are also spawning foxbox, the spawning instance should be called separately, it should be in this method

test/selenium/sessions_ui_test.js
```diff
@@ -7,13 +7,19 @@ var webdriver = require('selenium-webdriver'),
...
 var HOST_URL = util.format(
     'http://%s:%s/',
     process.env.BOX_LOCAL_NAME && process.env.BOX_LOCAL_NAME != 'localhost' ?
         (process.env.BOX_LOCAL_NAME + '.local.') : 'localhost',
     process.env.BOX_PORT || '3000'
 );
 
+Prepper.makeSuite('Test web app locally',function(){
+
+  before(Prepper.foxboxManager.fullOptionStart);
```
nojunpark: just realized, you don't need this if you're using makesuite.

test/selenium/sessions_ui_test.js
```diff
@@ -44,6 +50,8 @@ describe('sessions ui', function() {
 
   after(function() {
```
nojunpark, line 48: and you don't need this after, it's all part of makesuite already

test/selenium/lib/foxboxHelperSelenium.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
```
isabelrios, line 1: This is the same No-Jun created for integration with a few changes. Wanted to try if changing the [path ](https://github.com/isabelrios/foxbox/blob/ui_app_test/test/selenium/lib/foxboxHelperSelenium.js#L24)would fix the test not passing and yeah, that is the case.
Then I thought, although there is duplicated code, that having this for UI tests would be helpful, we need kill and start the foxbox(but not with all the services as integration tests do) 
This could be a first approach and then refactor to have one Helper for each if possible?

test/selenium/lib/testPrepperSelenium.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
```
isabelrios, line 1: Same than previous comment, using what No-Jun implemented for integration but with less options not needed so far for the UI tests

test/selenium/lib/view/log_in/accessors.js
```diff
@@ -0,0 +1,24 @@
...
+
+function LoginAccessors() {
+  Accessors.apply(this, arguments);
+}
+
+LoginAccessors.prototype = Object.assign({
+
+  get passwordField() {
+    return this.waitForElement(By.id('signup-pwd1'));
+  },
```
isabelrios: I know this and the other two accessors are defined for [here](https://github.com/fxbox/foxbox/blob/master/test/selenium/lib/view/sign_up/accessors.js#L6) but that way, they are not accessible from the app, need to do this.

isabelrios:
In the meantime the two red jobs will be fixed, and since the job for the test is green, could you @JohanLorenzo please take a look at how the test is implemented?
I left a few comments about the approach to get the test passing along with the existing sessioun_ui_test

test/selenium/lib/foxboxHelperSelenium.js
```diff
@@ -0,0 +1,46 @@
...
+      if (e.code === 'ENOENT') {
+        console.log(errMsg);
+      }
+    }
+  }
+
+  function removeUsersDB() {
+    _removeFileIfItExists('/.local/share/foxbox-debug/users_db.sqlite',
+      'User DB not found!');
+    _removeFileIfItExists('/.local/share/foxbox-debug/webpush.sqlite',
```
isabelrios: Oh, will remove these two lines

test/selenium/lib/view/log_in/view.js
```diff
@@ -0,0 +1,28 @@
...
+function LoginView() {
+  [].push.call(arguments, LoginAccessors);
+  View.apply(this, arguments);
+  
+  this.accessors.submitButton;
+}
+
+LoginView.prototype = Object.assign({
+
+  loginSuccess: function(password) {
```
JohanLorenzo: As we're entering both password and the password confirmation, I assume we're actually signing up. I would rename this function and its class. So we're clear on the fact that it's not just a regular log in.

test/selenium/lib/view/log_in/accessors.js
JohanLorenzo: I'm not too sure to understand why. Would it be easier if we used the SignUpView instead?

test/selenium/lib/foxboxHelperSelenium.js
JohanLorenzo, line 1: Yeah, that could be a hassle to modify the helper coming from the integration tests. Let's keep it as is, and factorize them in a next PR

JohanLorenzo:
Okay to copy the foxbox stater/stopper in this PR. I think we can make the intent clearer if we showed that we're landing on the first sign page. The rest looks good to me.

Regarding one of your red failures, #442 should fix it.

test/selenium/lib/view/log_in/accessors.js
isabelrios: The problem I have is that using those selectors, and the way they are defined, they are not located. Remember what was necessary for the app, like wait for the element to be located and displayed...other way, seems like they are not found. Will try again to modify this part and see it can use those locators

test/selenium/lib/view/log_in/view.js
isabelrios: Yeah, that makes sense

test/selenium/lib/view/log_in/accessors.js
isabelrios: Wait, you are right, for those locators that is not needed...sorry about that!

test/selenium/sessions_ui_test.js
```diff
@@ -7,13 +7,17 @@ var webdriver = require('selenium-webdriver'),
...
     process.env.BOX_LOCAL_NAME && process.env.BOX_LOCAL_NAME != 'localhost' ?
         (process.env.BOX_LOCAL_NAME + '.local.') : 'localhost',
     process.env.BOX_PORT || '3000'
 );
 
+Prepper.makeSuite('Test set up UI',function(){
+
 describe('sessions ui', function() {
   var driver;
   this.timeout(8000);
```
nojunpark: I don't think you'll need this timeout here, because the Prepper has the long timeout value before the describe() call

isabelrios:
Hi @JohanLorenzo, finally, after many tries, travis is totally green! r?

test/selenium/lib/view/sign_up/view.js
```diff
@@ -49,6 +49,18 @@ SetUpView.prototype = {
         });
     },
 
+    successSignUpFromApp: function(password, confirmPassword) {
```
JohanLorenzo: Nit: If we know sign up should success, there is no need to let the possibility to enter wrong data. I would remove the `confirmPassword` variable

JohanLorenzo:
r+ modulo [one small nit](https://github.com/fxbox/foxbox/pull/427#discussion_r61547260). Thanks!

isabelrios:
Reviewer's comments fixed and travis green. Merging

Issue #418, opened 2016-04-25
[Polish] Autofocus the login field


gmarty:
Let's set the focus on the password field when the authentication page opens.
@ferjm

ferjm:
Fixed by #429

Issue #417, opened 2016-04-25
[Polish] Allow submitting the login page with Enter


gmarty:
That'd be nice to submit the login page when pressing the Enter key.
@ferjm maybe.

ferjm:
Fixed by #429

PR #429, opened 2016-04-26
Issues #417 and #418. Submit on enter and autofocus for signin and si…

ferjm:
…gnup. r=gmarty

r? @gmarty


gmarty:
LGTM.
Do you have login and user creation tests? If so, we should add a test to make sure that Enter submits the form.
Otherwise r+!

PR #435, opened 2016-04-26
Fix most build warnings in openzwave-adapter

julienw:
- convert results wherever it makes sense
- extract IdMap into its own file
- use unwrap() for IdMap's locks which allows to simplify its API
- Write error logs for each result that we cannot handle (in the notification callback)

### DIFF
--- a/components/openzwave-adapter/src/id_map.rs
+++ b/components/openzwave-adapter/src/id_map.rs
@@ -1,0 +1,33 @@
+use taxonomy::util::Id as TaxoId;
+
+use std::sync::{ Arc, RwLock };
+
+#[derive(Debug, Clone)]
+pub struct IdMap<Kind, Type> {
+    map: Arc<RwLock<Vec<(TaxoId<Kind>, Type)>>>
+}
+
+impl<Kind, Type> IdMap<Kind, Type> where Type: Eq + Clone, Kind: Clone {
+    pub fn new() -> Self {
+        IdMap {
+            map: Arc::new(RwLock::new(Vec::new()))
+        }
+    }
+
+    pub fn push(&mut self, id: TaxoId<Kind>, ozw_object: Type) {
+        let mut guard = self.map.write().unwrap(); // we have bigger problems if we're poisoned
+        guard.push((id, ozw_object));
+    }
+
+    pub fn find_tax_id_from_ozw(&self, needle: &Type) -> Option<TaxoId<Kind>> {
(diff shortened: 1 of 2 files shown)


components/openzwave-adapter/src/id_map.rs
```diff
@@ -0,0 +1,33 @@
...
+        let find_result = guard.iter().find(|&&(_, ref item)| item == needle);
+        find_result.map(|&(ref id, _)| id.clone())
+    }
+
+    pub fn find_ozw_from_tax_id(&self, needle: &TaxoId<Kind>) -> Option<Type> {
+        let guard = self.map.read().unwrap(); // we have bigger problems if we're poisoned
+        let find_result = guard.iter().find(|&&(ref id, _)| id == needle);
+        find_result.map(|&(_, ref ozw_object)| ozw_object.clone())
+    }
+}
```
julienw, line 33: This was mostly moved from `src/lib.rs` with the following changes:
- we call "unwrap" directly here, so we don't need to return Results. I discussed with some local Rust experts, and we decided we should rather use https://doc.rust-lang.org/std/panic/fn.recover.html to avoid panics from reaching the C code than having Results for things that are very very unlikely like here. I'll file a separate bug for this.
- TaxId is renamed to TaxoId; I think it's better as "Tax" always refered to me to other types of taxes ;)

components/openzwave-adapter/src/lib.rs
```diff
@@ -234,18 +208,19 @@ fn set_ozw_vid_from_tax_value(vid: &ValueID, value: Value) -> Result<(), TaxErro
...
+                Value::DoorLocked(locked_unlocked) => { vid.set_bool(locked_unlocked == DoorLocked::Locked) }
+                _ => { return Err(TaxError::InvalidValue(value)) } // TODO InvalidType would be better but we'll need to fix specific types for specific TaxIds
             }
         }
-        _ => { return Err(TaxError::InternalError(InternalError::GenericError(format!("Unsupported OZW type: {:?}", vid.get_type())))); }
+        _ => { return Err(TaxError::InternalError(InternalError::GenericError(format!("Unsupported OZW type: {:?}", vid.get_type())))) }
     };
-    Ok(())
+
+    result.map_err(|e| TaxError::InternalError(InternalError::GenericError(format!("Error while setting a value: {}", e))))
```
julienw, line 223: I'm not really satisfied by these GenericErrors but we'll do better later.

julienw:
@dhylands r?

dhylands:
lgtm - r+

PR #436, opened 2016-04-26
Properly remove a Thinkerbell rule at runtime.

mcav:
Fix for https://github.com/fxbox/app/issues/86


azasypkin:
Looks good to me and works well with the application, thanks for fixing this!

r+

PR #437, opened 2016-04-26
Call "preventDefault()" for the submit buttons at login and signup pages. r=fabrice

azasypkin:
Default type for the buttons is `submit` but actual submit is performed in JS via XHR, so the easiest fix is to prevent default behavior for submit button.


ferjm:
Thanks for fixing this!

PR #432, opened 2016-04-26
Link libupnp with pkg_config

JohanLorenzo:
Fixes #430

PR #438, opened 2016-04-27
Readme: Add pkg-config dependency

JohanLorenzo:
Follow up #432
Also documented addition depenencies for OSX


JohanLorenzo:
Thanks for spotting it @isabelrios. r?

isabelrios:
r+, thanks for fixing all these issues and document it!

PR #440, opened 2016-04-27
Move dns challenge as an additional binary for foxbox (Closes #423, #…

### LINKED ISSUE #423: `cargo clean` doesn't clean components/ crates
ping @JohanLorenzo @samgiles

julienw:
…424)


julienw:
@fabricedesre @JohanLorenzo @samgiles what do you think ?

samgiles:
It works for me!

README.md
```diff
@@ -143,17 +143,20 @@ There are several command line options to start the daemon:
 Currently you would likely want to start the daemon like this:
 
 ```bash
-cargo run -- -r http://knilxof.org:4242 --disable-tls
+./run.sh -- -r http://knilxof.org:4242 --disable-tls
```
fabricedesre, line 146: We should not need the extra `--`

README.md
```diff
@@ -108,17 +108,17 @@ Foxbox expects certain executables to be available in the `PATH` during its exec
...
-| `dnschallenge` | No (required for LetsEncrypt DNS-01 challenge) | Built as a binary with `cargo build`, see `components/dns_challenge/target/<profile>` directory |
-| `bash`         | No (required for LetsEncrypt client)           | System package manager                                                                          |
-| `python`       | Yes (required to run `pagekite.py`)            | System package manager                                                                          |
-| `pagekite.py`  | Yes (required to enable tunneling)             | https://pagekite.net/wiki/OpenSource/                                                           |
+| Dependency     | Optional?                                      | Where to find it                                                                                         |
+| -------------- | ---------------------------------------------- |--------------------------------------------------------------------------------------------------------- |
+| `dnschallenge` | No (required for LetsEncrypt DNS-01 challenge) | Built as a binary with `cargo build` in the same target file as foxbox, see `target/<profile>` directory |
```
fabricedesre: What does "in the same target file" means? Is that "in the same target profile" ?

README.md
julienw, line 146: I discussed with @JohanLorenzo and we decided to keep it especially to be able to use cargo's `--verbose`.

We could special-case this but this is more work.

run.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+PATH=target/debug:"$PATH" cargo run --bin foxbox "$@"
```
fabricedesre, line 3: That's not good enough. We need to be able to run both debug (default) and release builds. We can enforce --release to be only recognized as the first parameter if that makes things easier though.

README.md
julienw: I guess I wanted to say "same target directory" :) I'll change this.

fabricedesre:
I'm fine with the approach, but there are a few things to address.

README.md
JohanLorenzo, line 146: We discussed that with @julienw, we thought `--` still has some value, when you want to `--verbose` or `--freatures` on cargo, for instance.

run.sh
julienw, line 3: I designed this to be used only as a development aid. As you said before, the PATH handling should be part of the deployment setup, and so I thought that we don't need to handle `--release`.

I'm totally fine working on this a little more (but I'm in PTO tonight :p) but I want to keep this run.sh very simple initially. I don't want a full-featured runner for now.

README.md
fabricedesre, line 146: But then your run.sh script adds very little over the cargo command...

julienw, line 146: yes, it's merely an alias to make things easier for the developer now that we have several binaries.

Totally open to rework this though :)

run.sh
fabricedesre, line 3: Ok, let's do a followup if other people feel as strongly as I do.

JohanLorenzo:
Looks good to me

julienw:
I pushed the small typo change.

julienw:
landed in ecc92fae61141185046ceaa7131a53a44258d4b4

PR #443, opened 2016-04-28
Add a Setter for pairing nodes

julienw:


components/openzwave-adapter/src/lib.rs
```diff
@@ -448,6 +500,8 @@ impl taxonomy::adapter::Adapter for OpenzwaveAdapter {
         values.drain().map(|(id, value)| {
             if let Some(ozw_vid) = self.setter_map.find_ozw_from_taxo_id(&id) {
                 (id, set_ozw_vid_from_taxo_value(&ozw_vid, value))
+            } else if let Some(ozw_controller) = self.pairing_map.find_ozw_from_taxo_id(&id) {
+                (id, start_pairing(&self.ozw, ozw_controller.get_home_id(), &value))
```
julienw: After all it would be easier if add_node was on the controller :)

components/openzwave-adapter/src/lib.rs
```diff
@@ -439,7 +491,7 @@ impl taxonomy::adapter::Adapter for OpenzwaveAdapter {
 
                 ozw_vid_as_taxo_value(&ozw_vid)
             });
-            let value_result: Result<Option<Value>, TaxoError> = taxo_value.ok_or(TaxoError::InternalError(InternalError::NoSuchGetter(id.clone())));
+            let value_result: Result<Option<Value>, TaxoError> = taxo_value.ok_or(TaxoError::GetterDoesNotSupportPolling(id.clone()));
```
julienw, line 482: this will likely be an id that is not a value_id... Likely we'll have to do better later on.

components/openzwave-adapter/src/lib.rs
```diff
@@ -241,10 +258,12 @@ impl OpenzwaveAdapter {
             setter_map: IdMap::new(),
             watchers: Arc::new(Mutex::new(Watchers::new())),
             value_cache: Arc::new(Mutex::new(HashMap::new())),
+            controller_map: IdMap::new(),
+            pairing_map: IdMap::new(),
         });
 
+        try!(box_manager.add_adapter(adapter.clone()));
         adapter.spawn_notification_thread(rx, box_manager);
-        try!(box_manager.add_adapter(adapter));
```
julienw, line 247: I realized we need to add the adapter before adding any other things. Even if it's really unlikely it would be possible with the previous code.

components/openzwave-adapter/src/lib.rs
```diff
@@ -241,10 +258,12 @@ impl OpenzwaveAdapter {
             setter_map: IdMap::new(),
             watchers: Arc::new(Mutex::new(Watchers::new())),
             value_cache: Arc::new(Mutex::new(HashMap::new())),
+            controller_map: IdMap::new(),
```
julienw, line 261: I'm not using this one yet but we'll need it when we'll implement messages to remove a Controller.

julienw:
3 commits: the 2 first are for taxonomy, the last one is for openzwave-adapter (with a slight addition to taxonomy as well).

I tested using these curl commands:

```
curl -X PUT -d '{"select":{"id": "OpenZWave-controller-01852cf6-pair"}, "value": {"Secure":"Insecure"}}' http://localhost:3000/api/v1/channels/set
curl -X PUT -d '{"select":{"id": "OpenZWave-controller-01852cf6-pair"}, "value": {"Secure":"Secure"}}' http://localhost:3000/api/v1/channels/set
```

(without authentication)
Of course you'll need to change the ID.

You should see a log in the foxbox log. I didn't try to actually pair a device either so I'd appreciate if you can have a look that this actually works.

julienw:
@dhylands please have a look and possibly change or land while I'm in PTO.

Enjoy !

julienw:
@azasypkin you might want to look at the new Taxonomy type I'm adding here.

components/taxonomy/src/services.rs
```diff
@@ -209,6 +209,16 @@ pub enum ChannelKind {
...
+    /// This kind is used with values that need to specify whether they're in secure mode.
+    /// ```
+    /// use foxbox_taxonomy::services::*;
+    /// use foxbox_taxonomy::parse::*;
+    ///
+    /// let parsed = ChannelKind::from_str("\"Secure\"").unwrap();
+    /// assert_eq!(parsed, ChannelKind::Secure);
+    /// ```
+    Secure,
+
```
julienw, line 226: I realize we likely want something like Pair or Pairing here. Secure is the Type, not the Kind...

I think you'll have to do the change and ask a review from yoric on monday...

components/openzwave-adapter/src/lib.rs
```diff
@@ -173,6 +173,21 @@ fn set_ozw_vid_from_taxo_value(vid: &ValueID, value: Value) -> Result<(), TaxoEr
     result.map_err(|e| TaxoError::InternalError(InternalError::GenericError(format!("Error while setting a value: {}", e))))
 }
 
+fn start_pairing(ozw: &ZWaveManager, home_id: u32, value: &Value) -> Result<(), TaxoError> {
+    match *value {
+        Value::Secure(ref is_secure) => {
+            let is_secure_bool = match *is_secure { Secure::Insecure => false, Secure::Secure => true };
```
fabricedesre: could that be just: let is_secure = *is_secure == Secure::Secure ?

components/openzwave-adapter/src/lib.rs
```diff
@@ -173,6 +173,21 @@ fn set_ozw_vid_from_taxo_value(vid: &ValueID, value: Value) -> Result<(), TaxoEr
...
+            try!(
+                ozw.add_node(home_id, is_secure_bool)
```
dhylands, line 181: Yeah - I tested this and instead of passing is_secure_bool, you can just put: *is_secure == Secure::Secure

julienw, line 181: ok thanks ! Actually this implements PartiaLEq so that's why :)

dhylands:
I'm also in the process of adding unpairing ...

julienw:
Should we wait for the decentralized taxonomy ?

dhylands:
r+

I also created #474 to add unpairing.

julienw:
Hey @dhylands, will you have a look at the last 2 commits ? I fixed the comment fabrice left, but I also changed the ChannelKind.

julienw:
@yoric can you please have a look at the taxonomy part of this PR ?

components/taxonomy/src/values.rs
```diff
@@ -70,6 +70,9 @@ pub enum Type {
     /// A boolean locked/unlocked states. Used for door locks.
     DoorLocked,
 
+    /// A boolean secure/insecure states. Used for zwave pairing and zwave secure devices.
+    Secure,
```
julienw: not sure if the type should be Secure or something else (IsSecure ?)

dhylands:
r+, although I still think Pair should be called ZWavePair

components/taxonomy/src/values.rs
Yoric: `IsSecure` sounds better.

components/taxonomy/src/values.rs
```diff
@@ -572,6 +577,143 @@ impl Deserialize for DoorLocked {
...
+    }
+}
+
+impl Ord for Secure {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.as_bool().cmp(&other.as_bool())
+    }
+}
+
+///
```
Yoric, line 669: I'm a bit surprised. Is this deserialization code used?

components/taxonomy/src/values.rs
```diff
@@ -1601,53 +1783,23 @@ impl PartialOrd for Value {
...
             (&WebPushNotify(ref a), &WebPushNotify(ref b)) => a.resource.partial_cmp(&b.resource),
-            (&WebPushNotify(_), _) => None,
-
             (&Binary(self::Binary {mimetype: ref a_mimetype, data: ref a_data}),
              &Binary(self::Binary {mimetype: ref b_mimetype, data: ref b_data})) if a_mimetype == b_mimetype => a_data.partial_cmp(b_data),
-            (&Binary(_), _) => None,
-
-            (&Range(_), &Range(_)) => None,
-            (&Range(_), _) => None,
+            _ => None
```
Yoric: Please don't do that. I realize that the existing code is more verbose, but it lets the compiler spot forgotten cases. Adding a catch-all clause breaks this nice property.

Yoric:
> r+, although I still think Pair should be called ZWavePair

Are there plans to make this kind work with non-ZWave adapters? If not, I agree with @dhylands, let's make it `ZWavePair`.

julienw:
> Are there plans to make this kind work with non-ZWave adapters? If not, I agree with @dhylands, let's make it ZWavePair.

Overtime I thought it could work for Philips Hue for example. But I have no idea whether we'll need the same parameter.

components/taxonomy/src/values.rs
julienw: OK, that's why I kept it in a separate commit, I thought you might not want it.

components/taxonomy/src/values.rs
julienw, line 669: Once we have some UX to trigger the pair through this setter, I guess we will use it ?

Yoric, line 669: Normally, the Serde (De)serialization is not used.

julienw, line 669: as said on IRC, Value itself derive Serialize and Deserialize, so I need them for now.

PR #444, opened 2016-04-29
Add the `cache` directive to Travis

fabricedesre:
Making use of https://github.com/travis-ci/travis-build/pull/710

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ rust:
   - nightly-2016-04-10
   - nightly
 
+cache: cargo
+
 matrix:
   fast_finish: true
   allow_failures:
--- a/tools/travis-linux-arm_cross_compile.sh
+++ b/tools/travis-linux-arm_cross_compile.sh
@@ -30,6 +30,7 @@ EOF
 
 _set_up_cargo_config() {
     mkdir -p "$HOME/.cargo"
+    rm -f "$HOME/.cargo/config"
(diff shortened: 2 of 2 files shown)


fabricedesre:
r? @JohanLorenzo 
Looks like the Mac failure is the intermittent?

JohanLorenzo:
Nice shortcut! Nonetheless, I don't think we should add it, for now. 

Both Mac jobs failed in your PR, which doesn't seem good. We've got 3 jobs since the cache disappeared. No intemittent failure happened there ([job 1](https://travis-ci.org/fxbox/foxbox/jobs/126448476), [job 2](https://travis-ci.org/fxbox/foxbox/jobs/126450014), [job 3](https://travis-ci.org/fxbox/foxbox/jobs/126621269)). Just to make sure,  I created [another dummy PR](445) and ran the mac jobs. I didn't see any intermittent failure after 2 times ([try 1](https://travis-ci.org/fxbox/foxbox/jobs/126597739), [try 2](https://travis-ci.org/fxbox/foxbox/jobs/126604785)) .

Then I cherry-picked your commits, and the failure immediately came at the first try. On that try, I enables `set -x`, which shows us [this last command](https://travis-ci.org/fxbox/foxbox/jobs/126604991#L1470). I don't know what it's supposed to do. I can investigate, but in the meantime I'd prefer not re-enabling cache.

fabricedesre:
ok, let's close that for now.

PR #446, opened 2016-04-29
Cargo sets DYLD_LIBRARY_PATH on Mac OS. Unset it for this process

samgiles:
The dylib loading of commands inside the letsencrypt client was failing: resulting in `Trace/BPT trap: 5`, this would only happen when using `cargo run`.

It's weird that this has started to happen now, but this solves the problem for this case.

@fabricedesre  r? 
cc @isabelrios


fabricedesre:
I'd rather have someone using a Mac verifying that.

PR #448, opened 2016-04-30
Recipe add/delete integration test

nojunpark:


test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,188 @@
...
+    source = JSON.stringify(source);
+
+  before(Prepper.foxboxManager.foxboxLogin);
+
+  it('Check service list for recipe',function(){        
+    return chakram.get(Prepper.foxboxManager.serviceListURL)
+    .then(function(listResponse) {
+      expect(listResponse).to.have.status(200);
+      var found = false;
+      for (var entry of listResponse.body) {
```
JohanLorenzo: I remember this pattern is used some places else in the test suite, but I remembered a simpler way to search within an array:

``` js
var isFound = listResponse.body.some(entry => entry.adapter === 'thinkerbell@link.mozilla.org');
expect(isFound).to.be.true; // .equals(true) passes if ['1'] is returned for instance, IIRC
```

We could update the PR at several places to be more concise

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,188 @@
...
+          if (entry.id === 'thinkerbell/First Recipe') {
+            found = true;
+          }
+        }
+      }
+      expect(found).equals(true);     
+    });
+  }); 
+
+  it('Add two recipes at the same time',function(){
```
JohanLorenzo: I think we can improve the wording => Does "Add two recipes one by one" sound better?

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,188 @@
...
+          }
+        }
+      }
+      expect(found).equals(true);     
+    });
+  }); 
+
+  it('Add two recipes at the same time',function(){        
+    
+    var payload = {
```
JohanLorenzo: Let's extract the payload generation in a function:

``` js

generateThinkerbellNewRecipePayload(recipeName) {
  return {
    'select':{
       'kind':'AddThinkerbellRule'
    },
    'value':{
       'ThinkerbellRule':{
          'name': recipeName,
          'source': source
        }
    }
  }
}
```

and then

``` js
var secondRecipePayload = generateThinkerbellNewRecipePayload('Second recipe')
return chakram.put(Prepper.foxboxManager.setterURL, payload) // By the way, this call can also be factorized
```

**Edit:** Following up [that comment](https://github.com/fxbox/foxbox/pull/448/files#r61723505), we could go like:

``` js
Foxbox.prototype = {
  addRecipe: function(recipeName) {
    var payload = this._generateNewRecipePayload('Second recipe')
    return chakram.put(Prepper.foxboxManager.setterURL, payload)
  },

  _generateNewRecipePayload: function(recipeName) {
    // See above
  } 
}
```

and then in the test:

``` js
  return foxbox.addRecipe('first recipe').then(response => {
    // and so on
  })
```

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,188 @@
...
+      'select':{
+         'kind':'AddThinkerbellRule'
+      },
+      'value':{
+         'ThinkerbellRule':{
+            'name':'Second Recipe',
```
JohanLorenzo: The fact that we have to start with 'second' shows that we don't clean up after the first tests. Let's add an `afterEach()` step that calls thinkerbell to drop everything

``` js
afterEach(() => {
  return foxbox.deleteAllRecipes();
});
```

then `foxbox` would look like:

``` js
Foxbox.prototype = {
  get allRecipes() {
    return chakram.get(this.setterURL);
  },

  deleteRecipe: function(recipe) {
    var payload = {'select': {'id': recipe.id },'value':null};
    return chakram.put(this.setterURL, payload);
  },

  deleteAllRecipes: function() {
    return this.allRecipes.then(recipes => {
       var deletionPromises = recipes.map(recipe => this.deleteRecipe(recipe));
       return Promise.all(deletionPromises);
    });
  } 
}
```

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,188 @@
...
+          if (entry.id === 'thinkerbell/First Recipe') {
+            found = true;
+          }
+        }
+      }
+      expect(found).equals(false);
+    });
+  });
+ 
+  it('Remove all remaining recipes',function(){
```
JohanLorenzo: Like [said here](https://github.com/fxbox/foxbox/pull/448/files#r61720681), "one by one" would sound better, I believe.

JohanLorenzo:
~~r? @JohanLorenzo~~

Nice to have recipe tests written! I believe we can make them easier to read by factorizing. I left some details along the files.

test/integration/test/recipe_test.js
nojunpark: aha, good to know, thanks!

test/integration/test/recipe_test.js
nojunpark: I thought about it, but I wanted to test a scenario where I add another recipe when a recipe is already present, and if I load a recipe in before(), the test would get too long.   Would above be allowed in such case?

Yoric:
This will be extremely useful, thanks for doing this. I am currently working on the next-generation Taxonomy, and this will cause several API changes, so it's good that we have integration tests.

nojunpark:
thanks, I anticipate the taxonomy changes, and will update the test accordingly as well, I'm also hoping to extend my tests, once I get the hang of it.

test/integration/test/recipe_test.js
nojunpark: I think I should do this refactoring once the decentralized taxonomy is in, I'm not sure whether I should put generateNewRecipePayload() inside foxbox.  For now, I have put them inside the recipe_test.js

nojunpark:
@JohanLorenzo  r?

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,177 @@
...
+
+  before(Prepper.foxboxManager.foxboxLogin);
+
+  it('Check service list for recipe',function(){        
+    return chakram.get(Prepper.foxboxManager.serviceListURL)
+    .then(function(listResponse) {
+      expect(listResponse).to.have.status(200);
+      var isFound = listResponse.body.some(
+        entry => entry.adapter === 'thinkerbell@link.mozilla.org');
+      // .equals(true) passes if ['1'] is returned for instance, IIRC
```
JohanLorenzo: Nit: I don't think this comment is needed in the code

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,177 @@
...
+            entry => entry.id === 'thinkerbell/Second Recipe');
+          expect(isFound).to.be.true; 
+          isFound = listResponse.body.some(
+            entry => entry.id === 'thinkerbell/Third Recipe');
+          expect(isFound).to.be.true; 
+        });
+      });
+  
+    after(function(){
+      var payload = {'select':
```
JohanLorenzo: Nit: This can be reduced to something smaller, like:

``` js
var promises = ['First', 'Second', 'Third'].map(number => chakram.put(Prepper.foxboxManager.setterURL, {'select':
  {'id': 'thinkerbell/' + number + ' Recipe/remove'},'value':null});
return Promise.all(promises);
```

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,177 @@
...
+      {'id': 'thinkerbell/First Recipe/remove'},'value':null};
+      return chakram.put(Prepper.foxboxManager.setterURL, payload)
+      .then(function(cmdResponse) {
+        payload = {'select':
+        {'id': 'thinkerbell/Second Recipe/remove'},'value':null};
+        return chakram.put(Prepper.foxboxManager.setterURL, payload);
+      })
+      .then(function(cmdResponse) {
+        payload = {'select':
+        {'id': 'thinkerbell/Second Recipe/remove'},'value':null};
```
JohanLorenzo: Nit: Bad copy and paste

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,177 @@
...
+    });
+  });
+
+  describe('remove recipes', function(){
+
+    before(function(){
+      return chakram.put(Prepper.foxboxManager.setterURL, 
+        generateThinkerbellNewRecipePayload('First Recipe'))
+        .then(function(cmdResponse) {
+          expect(cmdResponse).to.have.status(200);
```
JohanLorenzo: I would put assertions only in the tests themselves.

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,177 @@
...
+        {'id': 'thinkerbell/Second Recipe/remove'},'value':null};
+        return chakram.put(Prepper.foxboxManager.setterURL, payload);
+      });
+    });
+  });
+
+  describe('remove recipes', function(){
+
+    before(function(){
+      return chakram.put(Prepper.foxboxManager.setterURL,
```
JohanLorenzo: Nit: [Like there](https://github.com/fxbox/foxbox/pull/448/files#r61840116), we can reduce this before to something smaller.

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,177 @@
...
+        .then(function(cmdResponse) {
+          expect(cmdResponse).to.have.status(200);
+          return chakram.put(Prepper.foxboxManager.setterURL, 
+            generateThinkerbellNewRecipePayload('Third Recipe'));
+        });
+    });
+
+    it('Remove single recipe',function(){
+      var payload = {'select':
+      {'id': 'thinkerbell/First Recipe/remove'},'value':null};
```
JohanLorenzo: As we define the deletion payload many times, I would define a function `generateDeleteRecipePayload(recipeName)`

test/integration/test/recipe_test.js
```diff
@@ -0,0 +1,177 @@
...
+            'kind':'OpenClosed',
+            'range':{'Eq':{'OpenClosed':'Open'}}
+          }],
+          'execute':
+            [{'destination':
+              [{'kind':'Log'}],
+              'kind':'Log',
+              'value':{'String':'Closed'}}]}
+            ]};
+source = JSON.stringify(source);
```
JohanLorenzo, line 19: I'm not sure, why is it necessary for source to be a plain string?

JohanLorenzo:
Looks good to me, modulo the few nits raised. r+

test/integration/test/recipe_test.js
nojunpark: oh great, thanks!

test/integration/test/recipe_test.js
nojunpark, line 19: that's the accepted format for thinkerbell at the moment

PR #452, opened 2016-05-02
Fix selenium spawning 2 processes

### LINKED ISSUE #451: Selenium tests: Foxbox is not properly killed
# STR
1. `npm run test-selenium` => Notice the tests passing
2. `npm run test-selenium` one more time => Tests are not passing anymore
3. `ps aux | grep foxbox` => 3 instances of foxbox are listed

We likely spawn 2 instances each time we run the tests (one with nodejs and one with the tests). I think I see how to prevent that.

cc FYI @isabelrios @npark-mozilla

JohanLorenzo:
Fixes #451


nojunpark:
actually, I think there is a caching issue with foxbox.io.  It seems to me that it caches the old instance of foxbox, so it presents a list of available foxbox instances , thus the first test fails in some instances.  We might want to move the foxbox.io accessing test outside of travis.

test/selenium/lib/foxboxHelperSelenium.js
```diff
@@ -4,15 +4,16 @@ const path = require('path');
 
 const spawn = require('child_process').spawn;
 
+const PROFILE_PATH = path.join(process.env.HOME, '.local/share/foxbox-tests');
```
nojunpark: you're missing a / at the end, so it won't delete the sqlite file.

isabelrios: That would explain the result travis shows now, passing the first test but not the other since it fails in getting the sign up page as the user has not been removed

JohanLorenzo: Added the trailing slash (the path wasn't correct, otherwise). It seems either the sign up database is not removed or we don't contact the right foxbox.

JohanLorenzo:
I haven't seen any intermittent so far, when I re-enabled the test deactivated in #463. We have now one instance of foxbox running. When we restart foxbox, we keep the same public domain name (logs are now displayed in travis). As a consequence, if the github.io test fail, I think that's because we have 2 machines behind the same private IP. If that's the case, we would be able to see that in the logs.

I'd also like to refactor the test prepper to share it with the integration tests. But I'd prefer to do it in a follow up.

r? @isabelrios @npark-mozilla

isabelrios:
Hi @JohanLorenzo I set your branch and tried locally. I was able to reproduce the test failing. I run the tests on your branch first and both passes. Then I run foxbox and the tests on master. And the second time I tried to run the test here in your branch, the first test failed with the known error, it sees another foxbox...I checked that I do not have any other instance of foxbox running while launching the tests and that the ~/.local/share directory is empty...
Not sure why I still see another foxbox when opening the app, it is weird :(
Not sure where/how it is stored that foxbox :(

JohanLorenzo:
@isabelrios I got something similar on my end too. Just to double-check: Did you run the tests within less than 3 minutes? If that's the case, #471 should fix that.

isabelrios:
@JohanLorenzo yeah, I did not wait more than 3 minutes...and doing the same, running the tests without waiting (first on your branch, then on master, then again on your branch), I am still able to see the same behavior in #471

JohanLorenzo:
Was r+ in #471 (confirmed over IRC 10 minutes ago)

PR #453, opened 2016-05-02
Moving to new Philips Hue API authentication

cr:
This switches the Hue adapter to the new authentication mechanism that was optional until it became mandatory in firmware version 01032318 released on 2016-04-18. You will need to re-pair your hubs.

### DIFF
--- a/src/adapters/philips_hue/hub.rs
+++ b/src/adapters/philips_hue/hub.rs
@@ -11,19 +11,18 @@
 //! The module spawns a management thread for every hub.
 
 use serde_json;
-use std::sync::Arc;
+use std::sync::{ Arc, Mutex };
 use std::thread;
 use std::time::Duration;
 use super::hub_api::HubApi;
 use super::{ HueAction, PhilipsHueAdapter, structs };
 use traits::Controller;
-use uuid::Uuid;
 
 pub struct Hub<C> {
     pub adapter: PhilipsHueAdapter<C>,
     pub id: String,
     pub ip: String,
-    pub api: Arc<HubApi>,
+    pub api: Arc<Mutex<HubApi>>,
 }
 
 impl<C: Controller> Hub<C> {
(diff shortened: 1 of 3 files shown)


src/adapters/philips_hue/hub_api.rs
```diff
@@ -88,13 +96,63 @@ impl HubApi {
...
+            return Err(())
+        }
+        let response = match response.pop() {
+            Some(response) => { response },
+            None => { return Err(()) }
+        };
+        match response.success {
+            Some(success) => {
+                Ok(Some(success.username))
+            },
```
fabricedesre: We can get slightly more readable code using `if let`:
if let Some(success) = response.success {
    Ok(Some(success.username))
} else {
    ...
}

src/adapters/philips_hue/hub.rs
```diff
@@ -70,13 +69,28 @@ impl<C: Controller> Hub<C> {
...
+                                // Save the new token
+                                adapter.controller.get_config().set(
+                                    "philips_hue",
+                                    &format!("token_{}", id),
+                                    &new_token);
+                                api.lock().unwrap().update_token(&new_token);
+                                break;
+                            },
+                            Ok(None) => {
+                                info!("Push pairing button on Philips Hue Bridge {}", id);
```
fabricedesre: We first display a warn!("Philips Hue detected but not paired...") and here an info!(). warn!() are a bit easier to spot as a call to action I think.

fabricedesre:
Worked well with my bridge! Can you update your patch to use more `if let` when we only match something or None?

src/adapters/philips_hue/hub.rs
cr: Changed.

src/adapters/philips_hue/hub_api.rs
cr: Changed for maintaining the style. I actually think it's less readable, though.

cr:
The `if let Some(foo) = bar` pattern doesn't really work with the variable rebinding in L126.

fabricedesre:
Ok, please rebase without the dnschallenge/Cargo.lock changes and merge!

PR #456, opened 2016-05-02
Reuse the same TagStorage object for the lifetime of the State struct

fabricedesre:


fabricedesre:
@Yoric : Even with the lazy SQLite opening, we were creating a lot of TagStorage objects and then opening the db (one for each add_getter / add_setter etc.) : at least 10 times during startup, and many more when thinkerbell is used.

This let us reuse the same TagStorage instance per State struct lifetime, and go down to opening the underlying db only once.

PR #459, opened 2016-05-03
Logging tags database name to debug log

cr:
I'm seeing four INFO log lines regarding the `taxonomy tags database` per light service on my system, that looks like one log line per setter or getter. I propose to take this irrelevant info out of user view by diverting it to the debug log.


fabricedesre:
r=me

PR #461, opened 2016-05-03
Use https url for the Hue nupnp server

fabricedesre:
That's saving us a http redirect.

r? @cr


cr:
Interesting. They must have added HTTPS more or less recently. We should take note that this does not check the SSL certificate and thus provides no MITM protection. I agree it's better than nothing. Looks good to me.

cr:
Please also note that an existing `"nupnp_url"` key in `foxbox.conf` will need a manual edit for the change to become effective.

PR #463, opened 2016-05-03
deactivate selenium tests

nojunpark:


nojunpark:
Deactivate selenium tests - need to move the external network accessing test out of travis, or make sure the UI app does not cache older instance of foxbox

PR #464, opened 2016-05-04
Travis: Deactivate only the test using github.io

JohanLorenzo:
Follow up #463

### DIFF
--- a/test/selenium/app_test.js
+++ b/test/selenium/app_test.js
@@ -31,7 +31,7 @@ Prepper.makeSuite('Test to open web app', function() {
     driver.quit();
   });
 
-  describe('open the web app', function() {
+  describe.skip('open the web app', function() {
 
     beforeEach(function() {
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
Based on [this Travis job](https://travis-ci.org/fxbox/foxbox/jobs/127838483), I believe we can just take out github.io test, which should bring things back to what they used to be (before Friday). What do think @isabelrios @npark-mozilla ? r?

nojunpark:
yup agreed

isabelrios:
Agreed

JohanLorenzo:
Tests in components were failing because of https://github.com/fxbox/foxbox/issues/468 . Like said in [this comment](https://github.com/fxbox/foxbox/pull/464#issuecomment-216921688), we can see the effect of pinpointing the tests to deactivate in the travis job. 

Landing.

PR #375, opened 2016-04-14
Add cross-compilation

JohanLorenzo:
Fixes #368

PR #465, opened 2016-05-04
Updating to rust 2016-04-29

fabricedesre:


fabricedesre:
@JohanLorenzo Do you have any idea why the cross compilation fails? It seems to not pick up the new rustc version.

JohanLorenzo:
@fabricedesre You're right, it does keep [downloading and using 2016-04-10](https://travis-ci.org/fxbox/foxbox/jobs/127902913#L1170). I came realized the rustc nightly is also defined at [a second place](https://github.com/fxbox/foxbox/pull/465/files#diff-354f30a63fb0907d4ad57269548329e3R25), especially for the cross-compile build.

It's been there since we the [cross-compile build got in](https://github.com/fxbox/foxbox/pull/375/files#diff-354f30a63fb0907d4ad57269548329e3R26). At that time, I recall I couldn't define a variable that can be used in `.travis.yml`. Removing the duplicated line will introduce another build for nightly. As we're already facing machine availability issues, I'd suggest to keep the duplicated line and [add a comment here](https://github.com/fxbox/foxbox/pull/465/files#diff-354f30a63fb0907d4ad57269548329e3R16) saying something like "Please check the build matrix to see if this value needs to be updated there too".

PR #467, opened 2016-05-05
Add hue adapter color change test

nojunpark:


test/integration/lib/philipsHue_server.js
```diff
@@ -99,16 +118,23 @@ var philipsHue_server = (function() {
 
     _server.put('/api/:foxboxId/lights/:light_id/state', 
       bodyParser.json(),function (req, res) {
+        if (req.body.hue !== undefined && 
+          req.body.sat !== undefined && 
+          req.body.bri !== undefined) {
+          setHSB(req.params.light_id,req.body.hue,req.body.sat,req.body.bri);
+        }
+        
         if (req.body.on) {
```
JohanLorenzo: For consistency, I'd also change this line:

``` js
if (req.body.on === true)
```

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
+'use strict';
+
+const chakram = require('chakram'), expect = chakram.expect;
+
+var Prepper = require('../lib/testPrepper.js');
+
+function translateHSB(h,s,b) {
```
JohanLorenzo, line 7: Nice function! Out of curiosity, why is it out of the Philips class?

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+  before(Prepper.foxboxManager.foxboxLogin);
+
+  it('check 3 bulbs are registered',function(){
+    
+    // collect all getters for the lightbulbs
+    return chakram.put(Prepper.foxboxManager.getterURL,getterPayload)
+    .then(function(listResponse) {
+      lights = Object.keys(listResponse.body);
+      expect(lights.length).equals(3);
+      expect(listResponse).to.have.status(200);
```
JohanLorenzo, line 26: Nit: status must be the fist check

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+  var lights;
+  var getterPayload = [{'kind': 'LightOn'}];
+
+  before(Prepper.foxboxManager.foxboxLogin);
+
+  it('check 3 bulbs are registered',function(){
+    
+    // collect all getters for the lightbulbs
+    return chakram.put(Prepper.foxboxManager.getterURL,getterPayload)
+    .then(function(listResponse) {
```
JohanLorenzo, line 25: Nit: I don't remember, do we not use array functions for consistency purpose?

test/integration/test/philips_color_test.js
nojunpark, line 7: Oh, because i thought only this test would use it - technically, philips server doesn't do translation - the foxbox does, to the formate that philips server accepts

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+      lights = Object.keys(listResponse.body);
+      expect(lights.length).equals(3);
+      expect(listResponse).to.have.status(200);
+    });
+  });
+
+  describe('manipulate lights', function () {
+    before(function() {
+      // turn off all lights in the simulator
+      Prepper.philipshue_server.turnOffLight(1);
```
JohanLorenzo: Nit: There are 2 other places where we do these 3 calls. I'd factorize then in a function called `turnOffAllLights()`

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+      Prepper.philipshue_server.turnOffLight(2);
+      Prepper.philipshue_server.turnOffLight(3);
+
+      return chakram.put(Prepper.foxboxManager.getterURL,getterPayload)
+      .then(function(listResponse) {
+        lights = Object.keys(listResponse.body);
+        expect(lights.length).equals(3);
+        expect(listResponse).to.have.status(200);
```
JohanLorenzo: Nit: status check at the beginning of the then

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+  describe('manipulate lights', function () {
+    before(function() {
+      // turn off all lights in the simulator
+      Prepper.philipshue_server.turnOffLight(1);
+      Prepper.philipshue_server.turnOffLight(2);
+      Prepper.philipshue_server.turnOffLight(3);
+
+      return chakram.put(Prepper.foxboxManager.getterURL,getterPayload)
+      .then(function(listResponse) {
+        lights = Object.keys(listResponse.body);
```
JohanLorenzo: Nit: No need to perform assertions in `before()`

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+        expect(lights.length).equals(3);
+        expect(listResponse).to.have.status(200);
+      });
+    });
+
+    // Currently, there is no mapping between the foxbox
+    // id and the philips hue id until the tag feature is implemented
+    it('Turn on all lights', function(){
```
JohanLorenzo: Nit: I don't think we need to have this test since it doesn't check the color state

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+}
+
+Prepper.makeSuite('Control lights locally',function(){
+
+  var lights;
+  var getterPayload = [{'kind': 'LightOn'}];
+
+  before(Prepper.foxboxManager.foxboxLogin);
+
+  it('check 3 bulbs are registered',function(){
```
JohanLorenzo: Nit: I don't think we need to have this test since it doesn't check the color state

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+        expect(lights.length).equals(3);
+
+        lights.forEach(light => {
+          expect(listResponse.body[light].OnOff).equals('On'); 
+        });
+      });
+    });
+    
+    var testParams = [{testName:'upper boundary',
+    lightID:0, hue: 359, sat:0.9, bri:0.9},
```
JohanLorenzo: Nit: I'd put more simple values, for debugging purpose

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+        var bri = testParam.bri;
+        
+        // send various values within valid range
+        var payload = {'select': {'id': lights[testParam.lightID]}, 
+        'value': { 'Color': {'h':hue,'s':sat,'v':bri} } };
+
+        return chakram.put(Prepper.foxboxManager.setterURL,payload)
+        .then(function(cmdResponse) {
+          expect(cmdResponse).to.have.status(200);
+          var HSB = Prepper.philipshue_server.
```
JohanLorenzo: Nit: lower case, otherwise we might think it's a class

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+          getHSB(Prepper.philipshue_server.lastLight());
+          var trans = translateHSB(hue,sat,bri);
+          expect(HSB.h).equals(trans.h);
+          expect(HSB.s).equals(trans.s);
+          expect(HSB.b).equals(trans.b);             
+        });
+      });
+    });
+
+    it('Send invalid sat value', function() {
```
JohanLorenzo: Nice test!

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+          expect(HSB.s).equals(trans.s);
+          expect(HSB.b).equals(trans.b);             
+        });
+      });
+    });
+
+    it('Send invalid sat value', function() {   
+      lights[0] = lights[0].replace('getter:power','setter:color');
+      var hue = 200;
+      var sat = 1.5;  // Value should be less than or equal to 1
```
JohanLorenzo: Nit: I'd explicitly name the variable that holds the incorrect value. The others should be shared with the other tests, I believe

test/integration/test/philips_color_test.js
```diff
@@ -0,0 +1,144 @@
...
+        expect(cmdResponse.body).equals(
+          'TypeError { name: "s", at: "body.value.Color.Color",' +
+           ' expected: "a number in [0, 1]" }');
+      });
+    }); 
+
+    it('Send invalid brightness value', function() {   
+      lights[1] = lights[1].replace('getter:power','setter:color');
+      var hue = 200;
+      var sat = 0.5;  // Value should be less than or equal to 1
```
JohanLorenzo: Nit: Misleading comment. I'd remove it and use a clearer variable name

JohanLorenzo:
Looks good to me overall. Please remove the duplicate tests, and fix the small nits here and there. r+

test/integration/test/philips_color_test.js
JohanLorenzo, line 7: I see. Thanks for the extra-explanation!

test/integration/test/philips_color_test.js
nojunpark, line 25: I just remembered, for this kind of method we don't, above is the standard chakram convention for handling http response.

test/integration/test/philips_color_test.js
nojunpark: true, I don't think I need before() and this test

test/integration/test/philips_color_test.js
nojunpark: oh, actually, those are the upper bound values, since hue has a range of 0..359, sat: 0...1, and bri: 0...1

Issue #354, opened 2016-04-13
[MacOS][Travis] Unit test failure: 'test_managed_process_restart' panicked at 'Unknown waitpid error: No child processes (os error 10)'


JohanLorenzo:
Discovered while working on #333. There is a permanent failure that occurs on MacOS only. @samgiles and I agreed on ignoring this test on that platform for now. This bug is meant to put it back on.

Here's an example of the stacktrace: https://travis-ci.org/fxbox/foxbox/jobs/122540025#L523

JohanLorenzo:
Closed by #469

PR #469, opened 2016-05-06
Refactor so Backoff is separate from the ManagedProcess

samgiles:
- Add an API to retrieve number of times the process has restarted (useful for testing, will possibly follow up with #99)
- Small change to avoid fatals in weird conditions from libc (reactivated test: see #354)
- Added `checklock!` macro, to check and force poisoned locks.

This is in preparation for a managed process that isn't spawned with a command, but instead spawned from a closure (using `libc::fork`).

r? @fabricedesre

### DIFF
--- a/src/managed_process.rs
+++ b/src/managed_process.rs
@@ -7,7 +7,7 @@ use libc::{self,  c_int};
 
 use std::thread;
 use std::thread::JoinHandle;
-use std::sync::{ Arc, Mutex };
+use std::sync::{ Arc, Mutex, RwLock };
 use std::process::Child;
 use std::io::{ Error, ErrorKind, Result };
 use std::time::{ SystemTime, Duration, UNIX_EPOCH };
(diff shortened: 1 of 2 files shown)


src/managed_process.rs
```diff
@@ -26,7 +26,53 @@ fn seconds_since_epoch() -> f64 {
...
-    thread:      JoinHandle<()>
+    thread:      JoinHandle<()>,
+    backoff:     Arc<RwLock<Backoff>>,
+}
+
+struct Backoff {
+    restart_count: u64,
+    restart_threshold: f64,
+    start_time: f64,
+    end_time: f64,
```
fabricedesre: Can you use Duration/SystemTime instead of f64 for the various times? That would make the code clearer.

src/managed_process.rs
```diff
@@ -26,7 +26,53 @@ fn seconds_since_epoch() -> f64 {
...
+
+    fn next_backoff(&mut self) -> Duration {
+        self.end_time = seconds_since_epoch();
+
+        let seconds_to_backoff =
+            if (self.end_time - self.start_time) < self.restart_threshold {
+                self.backoff += 1;
+
+                // non-linear back off
+                (self.backoff * self.backoff) >> 1
```
fabricedesre: That will be 0 at the first call of `next_backoff()`, when `self.backoff == 1` and thus we restart the process immediately. I feel we may want to initialize `self.backoff` to 1 instead.

samgiles: Yeh, that's a slight nuance, wasn't sure if trying to restart immediately on the first failure was a bad thing :) - will initialise to 1 as it's possibly unexpected and likely not that useful in real life.

samgiles:
@fabricedesre I've addressed your comments, ok to merge?

src/managed_process.rs
```diff
@@ -16,17 +16,67 @@ use std::time::{ SystemTime, Duration, UNIX_EPOCH };
...
-    thread:      JoinHandle<()>
+    thread:      JoinHandle<()>,
+    backoff:     Arc<RwLock<Backoff>>,
+}
+
+struct Backoff {
+    restart_count: u64,
+    restart_threshold: Duration,
+    start_time: Duration,
+    end_time: Duration,
```
fabricedesre: I think `start_time` and `end_time` should be `std::time::Instant`, not `Duration`.

PR #470, opened 2016-05-06
[WIP] Decentralized Taxonomy, part 1.

Yoric:
r? @mcav 

Note that we can't land this yet, because we need to port Thinkerbell, the Taxonomy Router and Adapters.


Yoric:
I'm not entirely happy with the code, so I'd be interested in a second opinion.

My main problem at the moment is that we have two types of values (`Request` and `Value`) and two APIs (`native` and `json`).

A `Request` is essentially a JSON + Binary, while a `Value` is something that has already been parsed from a `Request`. I suppose we could get rid of `Value` entirely, and only use `Request` everywhere, but I have the intuition that we'd be losing an important service. Perhaps we should start with this, and only later re-introduce `Value` as needed.

Similarly, the `native` API manipulates `Value`s, while the `json` API manipulate `Request`s. This makes the `json` API harder to test, but perhaps that's not too much of a problem. This also makes it more clumsy for manipulation from within the Foxbox (e.g. Thinkerbell). Again, that may not be a real issue.

Cc @julienw, in case you have ideas.

components/taxonomy/src/adapters/adapter.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// if temperature rises above a certain value.
+    ///
+    /// The signature's `accepts` represents the condition on values.
+    pub watch: Option<Signature>,
+}
+
+
+impl<'a> Feature<'a> {
+    /// Define an empty feature
+    pub fn empty(id: &Id<FeatureId>, service: &Id<ServiceId>) -> Self {
```
julienw, line 167: should this be a default instead ?

components/taxonomy/src/adapters/adapter.rs
```diff
@@ -0,0 +1,364 @@
...
+///   send: Some(Signature {
+///     accepts: Expects::Requires(Arc::new(IsOnFormat)),
+///     ..Signature::default()
+///   }),
+///
+///   .. Feature::empty(&Id::new("light #1"), &Id::new("lights manager #1"))
+/// };
+/// ```
+#[derive(Clone)]
+pub struct Feature<'a> {
```
julienw, line 90: I think we need something to indicate a feature is user-facing or configuration stuff (eg: shoud the temperature be in C or F, or should the polling happen once every minute, or once every 10 minutes, etc)

components/taxonomy/src/adapters/adapter.rs
```diff
@@ -0,0 +1,364 @@
...
+        value: Value
+    },
+
+    /// Fired when we exit the range specified when we started watching. If no range was
+    /// specified, never fired.
+    Exit {
+        id: Id<FeatureId>,
+        value: Value
+    }
+}
```
julienw, line 257: I don't properly understand why we need Enter/Exit if we have no Range anymore now.

mcav:
Thanks @Yoric, I know you've put a lot of thought into these changes. I know that this is a huge undertaking. There's a lot of code and concepts to digest here; it's somewhat of a challenge to understand in isolation without seeing the (planned) changes to the adapters.

I'd like to apologize ahead of time: I still haven't had time to go thorough a completely thorough review of the code, but I wanted to offer a few general thoughts:

Overall, this approach still seems more complex than it ought to be. (I understand that the 'decentralized' effort isn't intended to necessarily simplify the architecture, but instead to remove some centralized dependencies.) Your intuition seems to agree that there's still some duplication here w/r/t `Request`/`Value` and `native`/`json`.

Generally, I'm inclined to want to remove anything that is not absolutely, positively necessary, even if it reduces type-safety. I'm more concerned about the ability of developers to understand the codebase than I am about ensuring we get everything exactly right.

From discussions with Fabrice, it sounds like we'll want to move adapters out-of-process at some point, effectively isolating them from the rest of the system. I've heard discussion about a rust-serialize-based IPC mechanism for communication.

My intuition there is that we should open ourselves to the possibility of creating non-Rust-based adapters. In that case, we'd want the communication bridge to be lightweight and language-agnostic. Non-Rust-based adapters might influence decisions here, regarding decentralized taxonomy.

Personally, I would rather see us adopt a language-agnostic approach to taxonomy — for instance, defining simple, JSON-based conventions for data exchanged by IoT devices, inspired by existing schemas like HomeKit. (Protocol Buffers might be interesting, perhaps they have rust bindings?) I get the sense that we're reinventing the wheel here. I'm not sure of the details, but broadly, it'd be cool if it didn't feel like we're just arbitrarily defining new data formats.

I understand that these comments aren't necessarily helpful for this specific pull request, and that some of my suggestions amount to a completely different "taxonomy" than we have today. It's probably worth having a team conversation about the broader picture here.

Again, I'd like to acknowledge that you've done a lot of work here, and I know there are a lot of details about these changes that I haven't yet had the time to understand and appreciate. I hope to continue looking at this over the next couple of days, such that I might have more detailed/applicable feedback.

cc @fabricedesre and @cr

components/taxonomy/src/adapters/adapter.rs
Yoric, line 90: Agreed, but later :)

components/taxonomy/src/adapters/adapter.rs
Yoric, line 167: Well, I can rename it `default`, but I still need to provide a unique `Id` for `id` and `service`.

Yoric:
Well, I'm stuck until we have further conversation.

> From discussions with Fabrice, it sounds like we'll want to move adapters out-of-process at some point, effectively isolating them from the rest of the system. I've heard discussion about a rust-serialize-based IPC mechanism for communication.
> 
> My intuition there is that we should open ourselves to the possibility of creating non-Rust-based adapters. In that case, we'd want the communication bridge to be lightweight and language-agnostic. Non-Rust-based adapters might influence decisions here, regarding decentralized taxonomy.

Well, part of the reasoning behind Decentralized Taxonomy is to make such a transition easier. I'm not planning to solve all the issues in the world in one mega-commit, though, so there will undoubtedly be more to do even after we have landed Decentralized Taxonomy.

> Personally, I would rather see us adopt a language-agnostic approach to taxonomy — for instance, defining simple, JSON-based conventions for data exchanged by IoT devices, inspired by existing schemas like HomeKit. (Protocol Buffers might be interesting, perhaps they have rust bindings?) I get the sense that we're reinventing the wheel here. I'm not sure of the details, but broadly, it'd be cool if it didn't feel like we're just arbitrarily defining new data formats.

Fwiw, introducing Protocol Buffers at this stage would only complicate things further and wouldn't simplify anything that I can think of.

components/taxonomy/src/adapters/adapter.rs
julienw, line 167: ok got it

Yoric:
Side-note: after much hesitation, I'll try and land Decentralized Taxonomy piece-wise, starting with the simple stuff. When we reach changes that require deeper changes, changes should be easier to review.

Yoric:
Closing PR is now obsolete.

PR #471, opened 2016-05-09
Selenium tests: Prevent two foxboxes from appearing in UI

JohanLorenzo:
Follow up #452


JohanLorenzo:
I just find why #452 didn't change the domain name of foxbox during tests. If you delete the certificates (which [currently happens](https://github.com/fxbox/foxbox/pull/471/files#diff-8cb3f13102317a849181d6f984f3438cL14)), foxbox will try to get a new domain name, otherwise it'll keep the one already assigned. Here's a small patch (see the 2 last commits) to make sure we don't carry anything from `~/.local/share/foxbox-tests` except the certificates.

Thus, we can now run `npm run test-selenium` many times in row, and only 1 foxbox will appear on the UI.

r? @isabelrios @npark-mozilla 

cc @samgiles to let him know we rely on certificates from now on.

isabelrios:
Tried many times and running the tests in a row works prefectly, thanks for fixing this!The patch looks good to me r+

test/selenium/app_test.js
```diff
@@ -3,18 +3,15 @@
 var Prepper = require('./lib/testPrepperSelenium.js');
 var webdriver = require('selenium-webdriver');
 
-var HOST_URL = 'https://fxbox.github.io/app';
+var HOST_URL = 'http://fxbox.github.io/app';
```
nojunpark, line 6: ah, so for our test it'll be set to http as well?

JohanLorenzo, line 6: We run foxbox with `--disable-tls`, so it's preferable to not mix http with https. We also cannot go with full https because we have to manually add one certificate exception

test/selenium/lib/testPrepperSelenium.js
```diff
@@ -5,19 +5,18 @@ var foxboxManager = require('./foxboxHelperSelenium.js');
...
-
     console.log('test started');
     foxboxManager.fullOptionStart(done);
   }
 
   function makeSuite(desc, test) {
     describe(desc, function () {
-      this.timeout(5000000);
+      this.timeout(30000);
       before(beforeTest);
```
nojunpark, line 15: I am currently in process of separating the sim startup routine so we can customize which sim to start up (needed for the hue authentication test), will ask for review once it's done

test/selenium/app_test.js
nojunpark, line 6: makes sense, I thought you'll enable full https for the UI test. cool!

nojunpark:
looks good to me, r+

test/selenium/lib/testPrepperSelenium.js
JohanLorenzo, line 15: Cool! Thanks for letting me know

PR #472, opened 2016-05-09
Selenium tests: Make tests less aware of how to set up the environment

JohanLorenzo:
Follow up #471


JohanLorenzo:
This PR changes these things:
- The helper is only in change of just managing the foxbox process => It has been renamed.
- For consistency, we don't use inline functions to define it, but prototypes. 
- The main App object is in charge of instantiating the webdriver. This allows the tests to be not aware of the details on how to set it up. Also, if one day, we have to change an option given to webdriver, we'd only have to change the app class.
- The prepper only interacts with the `app` and `foxboxProcessManager`. There is no need to expose other things. Hence it has been renamed to `makeSuite()`. 
- Small clean ups (duplicated success password, wrong use of waitForElement)

What do you think? r? @isabelrios @npark-mozilla 

I think we can go further with the refactor of the tests (by not depending from each other for instance). But the review would get too big IMO. I'd like to do it in a follow up.

nojunpark:
r+ looks good to me, perhaps we should also make cleanup changes to the integration test library files as well.

isabelrios:
This looks really good! tests much clearer and seems easy to use the libs to keep adding tests and functionality. Great work @JohanLorenzo.
There is only one thing I have seen when launching the tests, at the beginning I see two browsers open, one in blank and the other were the tests are running. For the second test, session_ui_test, only one is open and closed at the end. When tests finish all browsers are close, so I guess it would not be a problem but would like to comment just in case.
Thanks!

JohanLorenzo:
Thanks for pointing this out, @isabelrios. It turns out they are 2 windows [because of this line](https://github.com/fxbox/foxbox/pull/472/files#diff-b0b0b8eb64bf62de4eed6999d65b1e27R14). The driver is now created when JS files are interpreted. This is why we see 2 windows. This shouldn't be an issue. We keep running 1 test at a time.

isabelrios:
I see, thanks for explaining @JohanLorenzo. Great, r+ then :)

PR #474, opened 2016-05-10
Add Setter for unpairing a node

dhylands:
This modifies julien's PR #443 to add the complementary unpairing command.

### DIFF
--- a/components/openzwave-adapter/src/id_map.rs
+++ b/components/openzwave-adapter/src/id_map.rs
@@ -30,4 +30,9 @@ impl<Kind, Type> IdMap<Kind, Type> where Type: Eq + Clone, Kind: Clone {
         let find_result = guard.iter().find(|&&(ref id, _)| id == needle);
         find_result.map(|&(_, ref ozw_object)| ozw_object.clone())
     }
+
+    pub fn remove_by_ozw(&mut self, needle: &Type) -> Option<TaxoId<Kind>> {
+        let mut guard = self.map.write().unwrap(); // we have bigger problems if we're poisoned
+        guard.iter().position(|&(_, ref item)| item == needle).map(|index| guard.remove(index).0)
+    }
 }
--- a/components/openzwave-adapter/src/lib.rs
+++ b/components/openzwave-adapter/src/lib.rs
@@ -196,6 +196,14 @@ fn start_including(ozw: &ZWaveManager, home_id: u32, value: &Value) -> Result<()
     }
 }
 
+fn start_excluding(ozw: &ZWaveManager, home_id: u32) -> Result<(), TaxoError> {
+    try!(
+         ozw.remove_node(home_id)
+            .map_err(|e| TaxoError::InternalError(InternalError::GenericError(format!("Error while excluding node on network {}: {}", home_id, e)))));
+    info!("[OpenZWaveAdapter] Controller on network {} is awaiting an exclude, please do the appropriate steps to exclude a device.", home_id);
+    Ok(())
+}
+
 type ValueCache = HashMap<TaxoId<Channel>, Value>;
 
 pub struct OpenzwaveAdapter {
@@ -211,6 +219,7 @@ pub struct OpenzwaveAdapter {
     value_cache: Arc<Mutex<ValueCache>>,
     controller_map: IdMap<ServiceId, Controller>,
     include_map: IdMap<Channel, Controller>,
+    exclude_map: IdMap<Channel, Controller>,
 }
 
 fn ensure_directory<T: AsRef<Path> + ?Sized>(directory: &T) -> Result<(), Error> {
@@ -269,6 +278,7 @@ impl OpenzwaveAdapter {
             value_cache: Arc::new(Mutex::new(HashMap::new())),
             controller_map: IdMap::new(),
             include_map: IdMap::new(),
+            exclude_map: IdMap::new(),
         });
 
         try!(box_manager.add_adapter(adapter.clone()));
@@ -287,6 +297,7 @@ impl OpenzwaveAdapter {
         let mut setter_map = self.setter_map.clone();
         let mut controller_map = self.controller_map.clone();
         let mut include_map = self.include_map.clone();
+        let mut exclude_map = self.exclude_map.clone();
 
         let watchers = self.watchers.clone();
         let value_cache = self.value_cache.clone();
@@ -324,6 +335,21 @@ impl OpenzwaveAdapter {
                         }).unwrap_or_else(|e| {
                             error!("Couldn't add the setter {}: {}", include_setter_id, e);
                         });
+
+                        let exclude_setter_name = format!("OpenZWave-controller-{:08x}-exclude", home_id);
+                        let exclude_setter_id = TaxoId::new(&exclude_setter_name);
+                        exclude_map.push(exclude_setter_id.clone(), controller);
+
+                        box_manager.add_channel(Channel {
+                            feature: TaxoId::new("zwave/exclude"),
+                            supports_send: Some(Signature::nothing()),
+                            id: exclude_setter_id.clone(),
+                            service: service_id.clone(),
+                            adapter: adapter_id.clone(),
+                            .. Channel::default()
+                        }).unwrap_or_else(|e| {
+                            error!("Couldn't add the setter {}: {}", exclude_setter_id, e);
+                        });
                     }
                     ZWaveNotification::NodeNew(_node)               => {}
                     ZWaveNotification::NodeAdded(node)              => {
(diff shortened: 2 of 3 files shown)


dhylands:
r+ on Julien's portion of the patch
r? @julienw on the new portion

components/taxonomy/src/services.rs
```diff
@@ -158,6 +158,21 @@ pub enum ChannelKind {
...
+    /// This kind is represented by string "NoKind".
+    ///
+    /// ```
+    /// use foxbox_taxonomy::services::*;
+    /// use foxbox_taxonomy::parse::*;
+    ///
+    /// let parsed = ChannelKind::from_str("\"NoKind\"").unwrap();
+    /// assert_eq!(parsed, ChannelKind::NoKind);
+    /// ```
+    NoKind,
```
julienw: I'd like @Yoric advice on this.

components/openzwave-adapter/src/lib.rs
```diff
@@ -352,6 +385,12 @@ impl OpenzwaveAdapter {
                         if kind.is_none() { continue }
                         let kind = kind.unwrap();
 
+                        let mut tags = HashSet::new();
+                        let label = vid.get_label();
+                        if !label.is_empty() {
+                            tags.insert(TaxoId::<TagId>::new(&vid.get_label()));
```
julienw: nit: you don't need the `<TagId>` because it's properly infered through the called method's prototype.

And so you can remove it from the import line above.

julienw: also, wondering if this should be a tag or a property with a `label` key (or something else that we agree with our frontend friends). @azasypkin @gmarty ?

components/openzwave-adapter/src/lib.rs
```diff
@@ -451,7 +490,14 @@ impl OpenzwaveAdapter {
...
                     }
-                    ZWaveNotification::ValueRemoved(_value)         => {}
+                    ZWaveNotification::ValueRemoved(vid)            => {
+                        if let Some(getter_id) = getter_map.find_taxo_id_from_ozw(&vid) {
+                            let _ = box_manager.remove_getter(&getter_id);
+                        }
+                        if let Some(setter_id) = setter_map.find_taxo_id_from_ozw(&vid) {
+                            let _ = box_manager.remove_setter(&setter_id);
+                        }
+                    }
```
julienw, line 500: :+1:

julienw, line 500: instead of using `let _` maybe you can actually use the result and display a log if there is an error ?

julienw, line 500: also I think that we should remove the values from the maps getter_map and setter_map.

components/openzwave-adapter/src/lib.rs
```diff
@@ -337,11 +366,15 @@ impl OpenzwaveAdapter {
                         // https://github.com/fxbox/taxonomy/issues/97
                         // When it's done we can move the properties change from above to here.
                     }
-                    ZWaveNotification::NodeRemoved(_node)           => {}
+                    ZWaveNotification::NodeRemoved(node)           => {
+                        if let Some(service_id) = node_map.find_taxo_id_from_ozw(&node) {
+                            let _ = box_manager.remove_service(&service_id);
```
julienw: also use the result to display a log here; and we should remove the node from node_map

tools/scripts/svc.py
```diff
@@ -12,7 +12,8 @@
 
 kind_map = {
     'Password': 'String',
-    'Username': 'String'
+    'Username': 'String',
+    'NoKind': 'Unit'
```
julienw: do we need `Pair` (the new `Secure`) there as well ?

julienw:
I left a few comments on the commits; also please rebase on top of my PR that has a few more commits (including the one you did here for the equality condition, so we can keep one or the other, I don't mind)

components/taxonomy/src/services.rs
Yoric: That doesn't really make sense. A `ChannelKind` is supposed to describe a feature, not the type of the feature.

julienw: ok, so I think we want "Unpair" here (as a mirror to "Pair" I add in https://github.com/fxbox/foxbox/pull/443)

components/openzwave-adapter/src/lib.rs
azasypkin: Could you please clarify what this label is about exactly?

I'm currently working on the UI where user can add/remove tags (for the services for now), is this `label` the thing that can be safely removed by the user without losing something very important?

julienw: This is a human-readable string that defines the service. I see this as a way for the user to know which device this is.

components/taxonomy/src/services.rs
dhylands: > That doesn't really make sense. A ChannelKind is supposed to describe a feature, not the type of the feature.

@Yoric, I thought channelKind was describing the data being sent over the channel? It feels redundant to have to create artificial channel kinds for each operation that has no data. It doesn't seem to add anything other than complexity.

Right now, it feels like I have to add 20-30 lines of taxonomy to call 1 line of adapter code.

dhylands: @Yoric - Never Mind - I think I get it.

Yoric: With Decentralized Taxonomy, the process should be much lighter. I have a few difficulties landing Decentralized Taxonomy, though, so this may be a few weeks before we have this lighter process.

components/openzwave-adapter/src/lib.rs
dhylands: > also, wondering if this should be a tag or a property with a label key (or something else that we agree with our frontend friends). @azasypkin @gmarty ?

It should probably be a property with a label key. So for now I'll remover the tag stuff and add comment so we can add a property once changable properties are supported, except getters & setters don't seem to have properties.

dhylands:
Rebased on b2a7bad0e0fa1a0a3fb889ce720f859ff841d02e and addressed review comments.

julienw:
I landed the getter PR, will you please rebase on top of master? Then I'll have a second look :)
Thx !

dhylands:
Rebased on master, and tested.

julienw:
very sorry about this, seems like it slipped under my radar... and now with the taxonomy changes it bitrotted :/

dhylands:
@julienw rebased to the latest master and un-bitrotted, and tested.

components/openzwave-adapter/src/lib.rs
```diff
@@ -324,6 +335,21 @@ impl OpenzwaveAdapter {
...
                             error!("Couldn't add the setter {}: {}", include_setter_id, e);
                         });
+
+                        let exclude_setter_name = format!("OpenZWave-controller-{:08x}-exclude", home_id);
+                        let exclude_setter_id = TaxoId::new(&exclude_setter_name);
+                        exclude_map.push(exclude_setter_id.clone(), controller);
+
+                        box_manager.add_channel(Channel {
+                            feature: TaxoId::new("zwave/exclude"),
+                            supports_send: Some(Signature::returns(Maybe::Nothing)),
```
julienw: nit: use `Signature::nothing()`

components/openzwave-adapter/src/id_map.rs
```diff
@@ -30,4 +30,11 @@ impl<Kind, Type> IdMap<Kind, Type> where Type: Eq + Clone, Kind: Clone {
         let find_result = guard.iter().find(|&&(ref id, _)| id == needle);
         find_result.map(|&(_, ref ozw_object)| ozw_object.clone())
     }
+
+    pub fn remove_by_ozw(&mut self, needle: &Type) {
+        let mut guard = self.map.write().unwrap(); // we have bigger problems if we're poisoned
+        if let Some(index) = guard.iter().position(|&(_, ref item)| item == needle) {
+            guard.remove(index);
+        }
+    }
```
julienw, line 37: nit: if you return `Option<(TaxoId<Kind>, Type)>`, we can use it later on. Maybe this can work:

``` rust
guard.iter().position(|&(_, ref item)| item == needle).map(|index| guard.remove(index))
```

(as `remove` returns the removed element).

components/openzwave-adapter/src/lib.rs
```diff
@@ -346,11 +372,17 @@ impl OpenzwaveAdapter {
                         // https://github.com/fxbox/taxonomy/issues/97
                         // When it's done we can move the properties change from above to here.
                     }
-                    ZWaveNotification::NodeRemoved(_node)           => {}
+                    ZWaveNotification::NodeRemoved(node)           => {
+                        if let Some(service_id) = node_map.find_taxo_id_from_ozw(&node) {
```
julienw: here you can use `remove_by_ozw` if it returns the removed item as advised above (is there a reason you don't remove the item from node_map here BTW?)

components/openzwave-adapter/src/lib.rs
```diff
@@ -454,7 +486,20 @@ impl OpenzwaveAdapter {
...
+                                error!("Unable to remove getter_id {}: {}", getter_id, e);
+                            });
+                            getter_map.remove_by_ozw(&vid);
+                        }
+                        if let Some(setter_id) = setter_map.find_taxo_id_from_ozw(&vid) {
+                            box_manager.remove_channel(&setter_id).unwrap_or_else(|e| {
+                                error!("Unable to remove setter_id {}: {}", setter_id, e);
+                            });
+                            setter_map.remove_by_ozw(&vid);
+                        }
```
julienw, line 499: same here, you can use `remove_by_ozw` directly instead of `find_taxo_id_from_ozw` if it returns the removed item. And I think we'd also be threadsafe thanks to the guard in remove_by_ozw.

components/openzwave-adapter/src/id_map.rs
dhylands, line 37: I think it makes more sense to just return the id (rather than the entire tuple).

dhylands:
@julienw - applied review comments

julienw:
r=me but please wait for a green build and do a sanity run with this patch in. Thanks for the patch !

PR #475, opened 2016-05-10
Travis: Make jshint rule broader

JohanLorenzo:
Blocks #472 

While working on #472, I realized some files are not scanned by jshint. I changed the pattern in Travis and fixed the pointed out files.


isabelrios:
Oh thanks for fixing this, my bad I missed that line in those files. r+

PR #478, opened 2016-05-10
Decentralizing: Merging Getter and Setter

Yoric:


components/openzwave-adapter/src/lib.rs
```diff
@@ -302,36 +302,28 @@ impl OpenzwaveAdapter {
                         let kind = kind.unwrap();
 
                         if has_getter {
-                            let getter_id = TaxoId::new(&value_id);
+                            let getter_id = TaxoId::<Channel>::new(&value_id);
```
julienw, line 305: are you sure this is needed ? Isn't it infered by the subsequent push ?

components/openzwave-adapter/src/lib.rs
```diff
@@ -302,36 +302,28 @@ impl OpenzwaveAdapter {
...
-                                }
+                                kind: kind.clone(),
                             }).unwrap_or_else(|e| {
                                 error!("Couldn't add the getter {}: {}", value_id, e);
                             });
                         }
 
                         if has_setter {
-                            let setter_id = TaxoId::new(&value_id);
+                            let setter_id = TaxoId::<Channel>::new(&value_id);
```
julienw, line 319: same question

components/openzwave-adapter/src/lib.rs
```diff
@@ -182,8 +182,8 @@ pub struct OpenzwaveAdapter {
     version: [u32; 4],
     ozw: ZWaveManager,
     node_map: IdMap<ServiceId, Node>,
-    getter_map: IdMap<Getter, ValueID>,
-    setter_map: IdMap<Setter, ValueID>,
+    getter_map: IdMap<Channel, ValueID>,
+    setter_map: IdMap<Channel, ValueID>,
```
julienw, line 186: I think we'll want to merge these maps into the same map; probably more than you're willing to do in this patch but please file a bug.

julienw:
A few comments on the zwave part; you'll need a review for the other parts though.

components/openzwave-adapter/src/lib.rs
Yoric, line 186: I'm currently writing the patch to do this in Taxonomy, so yeah, that's planned.

components/openzwave-adapter/src/lib.rs
Yoric, line 305: I'll double-check.

components/openzwave-adapter/src/lib.rs
julienw, line 186: But here we're not in taxonomy ? Will you do this here as well ?

Yoric, line 186: Not 100% sure. I'll file an issue, as you suggested.

dhylands:
r+ for ip_camera related changes

Yoric:
Also, f? @mcav if you wish to see in which direction Decentralized Taxonomy is heading.

fabricedesre:
r=me for the adapter/tts/ changes

PR #480, opened 2016-05-10
Selenium tests: Make tests in session_ui not dependent one from each other

JohanLorenzo:
Depends on #472

### DIFF
--- a/test/selenium/lib/make_suite.js
+++ b/test/selenium/lib/make_suite.js
@@ -20,33 +20,33 @@ Suite.prototype = {
 
       before(() => {
         return self.foxboxProcessManager.start()
-          .then(() => self.app.init());
+        .then(() => self.app.init());
       });
 
       subSuite(self.app);
 
       after(() => {
         return self.app.stop()
-          .then(() => self.foxboxProcessManager.kill())
-          .then(() => self.foxboxProcessManager.cleanData())
+        .then(() => self.foxboxProcessManager.kill())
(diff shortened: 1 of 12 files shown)


JohanLorenzo:
Set ups and tear downs are now explicitly detailed, allowing us to run one test without needing the result from another. I also seized the occasion to enforce the POP in the remaining tests. We should not loose functional coverage between the old tests and the new ones, but another pair of eyes would help me checking that :smiley: .

r? @npark-mozilla

test/selenium/lib/make_suite.js
```diff
@@ -3,33 +3,50 @@
...
     });
-  });
-}
+  },
+
+  browserCleanUp() {
+    return this.app.clear()
+    .then(() => this.app.init(),
+          () => this.app.init());
+  },
```
nojunpark, line 42: just curious, why the init() called twice here? i'm not so familiar with the arrow notation.

JohanLorenzo, line 42: `then()` can take 2 parameters: the first one is a callback when things are going fine, and the second one when it failed. Here I wanted to run the init() even if `clear()` failed (for instance, when you run a test in the middle, and there is nothing to clear). You're right on one point, though: things are odd written this way. I'll add a comment explaining why things are written twice in this context.

nojunpark:
other than the need for a comment, everything looks good to me, r+

PR #481, opened 2016-05-10
Hueauthtest- Part 1: Clean up the sim startup

nojunpark:
Make the bootup sequence more manageable, and remove obsolete tests


nojunpark:
This PR basically has no new features, I made it so it can handle separate enable of sims, and did some cleanup

test/integration/lib/foxbox_process_manager.js
```diff
@@ -10,7 +10,7 @@ var config = new Config('./test/integration/lib/config/foxbox.js');
 var FOXBOX_STARTUP_WAIT_TIME_IN_MS = 5000;
 var foxboxInstance;
 
-var helper = (function() {
+var foxbox_process_manager = (function() {
```
JohanLorenzo, line 13: Nit: we usually camelCase instead of snake_case in JS.

**Edit:** /me realized snake case is also used some places else

test/integration/test/ip_camera_test.js
```diff
@@ -5,12 +5,14 @@ const Config = require('config-js');
...
 
 var config = new Config('./test/integration/lib/config/foxbox.js');
-var Prepper = require('../lib/testPrepper.js');
+var Prepper = require('../lib/make_suite.js');
 
 Prepper.makeSuite('Control camera locally',function(){
 
   var cameraService;
 
+  before(Prepper.turnAllSims);
```
JohanLorenzo: Having many `before()`s is undocumented. Do you have a link to demonstrate that mocha is supposed to run all the befores in the correct order?

Otherwise, I'd stick with something like:

``` js
before(() => {
  return Prepper.turnOnAllSimulators()
    .then(() => Prepper.turnOnFoxbox())
    .then(() => Prepper.foxboxManager.foxboxLogin())
});
```

That makes me realize `turnOnAllSimulators` is suppossedly sync, `turnOnFoxbox()` expects a callback and `foxboxLogin()` is a Promise. I don't know how many `before()`s deal with that particular case. I'd make what needs to be async returns promises

test/integration/lib/make_suite.js
```diff
@@ -0,0 +1,97 @@
...
+var foxboxManager = require('../lib/foxbox_process_manager.js');
+var config = new Config('./test/integration/lib/config/foxbox.js');
+
+
+var make_suite = (function() {
+  var hueSimOn = false;
+  var cameraSimOn = false;
+  var webPushSimOn = false;
+  
+  function turnAllSims() {
```
JohanLorenzo: Nit: `on` is missing. I'd also explicitly name `simulators` instead of `sims` => `turnOnAllSimulators`

test/integration/lib/make_suite.js
```diff
@@ -0,0 +1,97 @@
...
+  function turnAllSims() { 
+      
+    //Start the foxbox and associated simulators
+    turnOnHue();
+    turnOnCamera(); 
+    turnOnWebPush();
+ }
+
+  function turnOnHue(){
+    nupnp_server.start(config.get('nupnp_server.id'),
```
JohanLorenzo: I realized `app.listen()` [is async](http://expressjs.com/en/4x/api.html#app.listen). I'd make `setup()` for each server return a Promise

test/integration/lib/make_suite.js
```diff
@@ -0,0 +1,97 @@
...
+
+
+var make_suite = (function() {
+  var hueSimOn = false;
+  var cameraSimOn = false;
+  var webPushSimOn = false;
+  
+  function turnAllSims() { 
+      
+    //Start the foxbox and associated simulators
```
JohanLorenzo: Nit: Lying comment. Foxbox is not started here. Let's get rid of it :smiley:

test/integration/lib/make_suite.js
```diff
@@ -0,0 +1,97 @@
...
+    cameraSimOn = false;
+    return ipcamera_server.stop();
+  }
+
+  function webPushOff() {
+    webPushSimOn = false;
+    return webPush_server.stop();
+  }
+
+  function makeSuite(desc, test) {
```
JohanLorenzo: Nit: test can also be a suite. Let's rename it `subSuite`?

test/integration/lib/make_suite.js
```diff
@@ -0,0 +1,97 @@
...
+    describe(desc, function () {
+      this.timeout(500000);
```
JohanLorenzo: ~~This timeout won't pass on Travis. There's a 10-minute-limit there.~~ This timeout is nearly at the 10-minute-limit allowed by Travis. I'd suggest to downgrade the time to 30 or 60 seconds, so people will have a faster feedback.

test/integration/test/no_travis/pagekite_access_test.js
```diff
@@ -1,96 +0,0 @@
-'use strict';
```
JohanLorenzo, line 1: It seems these tests are not present somewhere else in the PR. Should the file be renamed?

JohanLorenzo:
Nice refactor. There are a few async things to clean up before merging the PR, IMO

test/integration/test/ip_camera_test.js
nojunpark: I think I can't use arrow functions within the hook, as the doc says: "Passing arrow functions to Mocha is discouraged. Their lexical binding of the this value makes them unable to access the Mocha context, and statements like this.timeout(1000); will not work inside an arrow function."

Also both the non-arrow and arrow format didn't work with mocha before() hook.

I did read that it supports multiple before() in order, but that wasn't from the formal documentation. foxboxLogin() is a chain, but it is sync, so I don't think we need to worry about asnycs on that.

test/integration/lib/make_suite.js
nojunpark: huh, how did it end up  there, sorry :<

test/integration/test/no_travis/pagekite_access_test.js
nojunpark, line 1: No, I had to remove this test, because with the TLS implementation, the ping server will not provide the tunnel url, making this test obsolete.

test/integration/lib/foxbox_process_manager.js
nojunpark, line 13: Okay, so... I'll leave it as it is.

test/integration/lib/make_suite.js
nojunpark: I wrapped each listen() with promise.

nojunpark:
@JohanLorenzo r?

test/integration/lib/ipcamera_server.js
```diff
@@ -49,32 +49,37 @@ var ipCamera_server = (function() {
...
+
     _server.get('/image/jpeg.cgi', function (req, res) {
-        //send a jpeg
-        console.log('Camera: snapshot request received');
-        res.status(200).sendFile(
-          path.join(__dirname,'colville.jpg'));
+      //send a jpeg
+      console.log('Camera: snapshot request received');
+      res.status(200).sendFile(
+        path.join(__dirname,'colville.jpg'));
```
JohanLorenzo, line 59: Nice piece of :art:, by the way!

test/integration/lib/ipcamera_server.js
```diff
@@ -49,32 +49,37 @@ var ipCamera_server = (function() {
...
     });
 
-    instance = _server.listen(parseInt(port), function () {
+    return new Promise(resolve => {
+      // Stopping cameraUPnPServer causes socket error on next script. disabled.
+      // Also, Wireshark does not show additional traffic even when not stopped.
```
JohanLorenzo: I'm not sure what this comment is about when starting the server. Is it a copy and paste issue [from here](https://github.com/fxbox/foxbox/pull/481/files#diff-14e2cb55c60a4b51fa665e1ed344c25cR75)

test/integration/lib/make_suite.js
JohanLorenzo: Nice. Promises must be returned in the turnOnFunctions, otherwise mocha will likely not wait on them

test/integration/lib/ipcamera_server.js
nojunpark, line 59: it's a Canadian painting!

test/integration/lib/ipcamera_server.js
nojunpark: ugh, yes it was a copy and paste error.

nojunpark:
@JohanLorenzo r?

test/integration/lib/foxbox_process_manager.js
JohanLorenzo, line 13: Yeah, let's update all the case issues in a follow up that will enforce that with jshint

test/integration/lib/make_suite.js
```diff
@@ -0,0 +1,109 @@
...
+    return ipcamera_server.setup();
+  }
+
+  function turnOnWebPush() {
+    webPushSimulatorOn = true;
+    return webPush_server.setup();
+  }
+
+  function turnOnFoxbox(done) {
+    foxboxManager.fullOptionStart(done);
```
JohanLorenzo: Nit: For a matter consistency, it would be nice to return a promise here too

test/integration/lib/make_suite.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  function nupnpOff() {
+    nupnp_server.stop();
```
JohanLorenzo: Nit: Promise not bubbled up

test/integration/lib/make_suite.js
```diff
@@ -0,0 +1,109 @@
...
+    cameraSimulatorOn = false;
+    return ipcamera_server.stop();
+  }
+
+  function webPushOff() {
+    webPushSimulatorOn = false;
+    return webPush_server.stop();
+  }
+
+  function subSuite(desc, test) {
```
JohanLorenzo: I apologize for [that unclear comment](https://github.com/fxbox/foxbox/pull/481#discussion_r62815479). I meant: keep the function name `makeSuite`, but change the parameter `test` into `subSuite`.

JohanLorenzo:
Looks good to me, minus a few small nits. r+

test/integration/lib/make_suite.js
nojunpark: thought about it, but I since it doesn't have any async, i thought it'd be a bit misleading, plus I couldn't find a easy way to add promise that resolves on callbacks.

JohanLorenzo: Sorry, `fullOptionStart()` is async because of the presence of `setTimeout()`. Here's [an example](https://github.com/fxbox/foxbox/blob/master/test/selenium/lib/foxbox_process_manager.js#L20-L28) of wrapping `setTimeout()` in a promise.

nojunpark: ah I see my mistake now, I was trying to include settimeout in an anon. function. thanks!

PR #484, opened 2016-05-13
Decentralizing: Introducing supports_*

Yoric:
This set of patches continues towards Decentralized Taxonomy:
- A channel watch will now be informed upon topology changes, regardless of whether these changes occur in getter or setter channels;
- Getter and setter channels are now fully merged. Instead of several different data types, we now have a single structure `Channel`, with capabilities `supports_send`, `supports_fetch`, `supports_watch`.

Note that this set of patches is built on top of the (yet unlanded) https://github.com/fxbox/foxbox/pull/478.

### DIFF
--- a/components/taxonomy/src/selector.rs
+++ b/components/taxonomy/src/selector.rs
@@ -30,8 +30,19 @@ pub trait ServiceLike {
     fn id(&self) -> &Id<ServiceId>;
     fn adapter(&self) -> &Id<AdapterId>;
     fn with_tags<F>(&self, f: F) -> bool where F: Fn(&HashSet<Id<TagId>>) -> bool;
-    fn has_getters<F>(&self, f: F) -> bool where F: Fn(&Channel) -> bool;
-    fn has_setters<F>(&self, f: F) -> bool where F: Fn(&Channel) -> bool;
+    #[deprecated]
+    fn has_setters<F>(&self, f: F) -> bool where F: Fn(&Channel) -> bool {
+        self.has_channels(|chan| {
+            chan.supports_send && f(chan)
+        })
+    }
+    #[deprecated]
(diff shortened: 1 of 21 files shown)


mcav:
thinkerbell changes lgtm

julienw:
looks good for openzwave, but I'll assume you tested the code with a zwave adapter. Tell me if you need me to do it.

Also the checks are failing below :)

julienw:
I think you'll need a review for the other pieces of code too.

Yoric:
r? @dhylands for IP Camera
r? @cr for Philipps Hue
f? @fabricedesre fror TTS

components/taxonomy/src/selector.rs
```diff
@@ -30,8 +30,19 @@ pub trait ServiceLike {
...
+        self.has_channels(|chan| {
+            chan.supports_send && f(chan)
+        })
+    }
+    #[deprecated]
+    fn has_getters<F>(&self, f: F) -> bool where F: Fn(&Channel) -> bool {
+        self.has_channels(|chan| {
+            chan.supports_fetch && f(chan)
+        })
+    }
```
fabricedesre, line 44: Seems like we can just get rid of these 2 methods no?

Yoric, line 44: Currently, they are still used by a selector. As this would require breaking the REST API, I preferred keeping that change for another patch.

cr:
r+ @Yoric

The Hue changes look good to me.

fabricedesre:
r=me for the tts changes

components/taxonomy/src/selector.rs
fabricedesre, line 44: Ha yes... damn this github search is useless...

dhylands:
r+ for the ip_camera stuff as well

PR #485, opened 2016-05-15
Trying rustfmt

fabricedesre:
Let's see if https://github.com/rust-lang-nursery/rustfmt works for us.

### DIFF
--- a/components/openzwave-adapter/src/lib.rs
+++ b/components/openzwave-adapter/src/lib.rs
@@ -9,24 +9,24 @@ mod watchers;
 
 
 use taxonomy::channel::*;
-use taxonomy::util::{ Id as TaxoId, Maybe, ref_eq };
-use taxonomy::services::{ AdapterId, ServiceId, Service };
+use taxonomy::util::{Id as TaxoId, Maybe, ref_eq};
+use taxonomy::services::{AdapterId, ServiceId, Service};
 use taxonomy::values::*;
-use taxonomy::api::{ Operation, ResultMap, Error as TaxoError, InternalError, User };
-use taxonomy::adapter::{ AdapterManagerHandle, AdapterWatchGuard, WatchEvent };
+use taxonomy::api::{Operation, ResultMap, Error as TaxoError, InternalError, User};
+use taxonomy::adapter::{AdapterManagerHandle, AdapterWatchGuard, WatchEvent};
 use transformable_channels::mpsc::ExtSender;
 
-use openzwave::{ ConfigPath, InitOptions, ZWaveManager, ZWaveNotification };
-use openzwave::{ CommandClass, ValueGenre, ValueType, ValueID };
-use openzwave::{ Controller, Node };
+use openzwave::{ConfigPath, InitOptions, ZWaveManager, ZWaveNotification};
+use openzwave::{CommandClass, ValueGenre, ValueType, ValueID};
+use openzwave::{Controller, Node};
 
 use std::error;
 use std::fmt;
-use std::{ fs, io };
+use std::{fs, io};
 use std::path::Path;
 use std::thread;
 use std::sync::mpsc;
-use std::sync::{ Arc, Mutex };
+use std::sync::{Arc, Mutex};
 use std::collections::HashMap;
 
 use id_map::IdMap;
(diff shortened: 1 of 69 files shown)


fabricedesre:
The changes look huge but it's been all done by running `cargo fmt` in the main directory and in each component/ subdirectory.

I like that this gives us consistent style and formatting without effort, even if this doesn't always match my personal preferences.

If we go this route, we need to enforce that consistently, eg.  by running rustfmt like clippy in the pre-commit hook.

components/openzwave-adapter/src/lib.rs
```diff
@@ -123,11 +127,15 @@ impl RangeChecker for Range {
...
 fn taxo_kind_from_ozw_vid(vid: &ValueID) -> Option<ChannelKind> {
     match (vid.get_type(), vid.get_command_class(), vid.get_index()) {
-        (ValueType::ValueType_Bool, Some(CommandClass::DoorLock),     0) => Some(ChannelKind::DoorLocked),
-        (ValueType::ValueType_Bool, Some(CommandClass::SensorBinary), _) => Some(ChannelKind::OpenClosed),
+        (ValueType::ValueType_Bool, Some(CommandClass::DoorLock), 0) => {
+            Some(ChannelKind::DoorLocked)
+        }
+        (ValueType::ValueType_Bool, Some(CommandClass::SensorBinary), _) => {
+            Some(ChannelKind::OpenClosed)
+        }
```
julienw: I could argue that it's less readable after the change.

components/openzwave-adapter/src/lib.rs
```diff
@@ -161,16 +181,22 @@ fn set_ozw_vid_from_taxo_value(vid: &ValueID, value: Value) -> Result<(), TaxoEr
...
             }
         }
         _ => { return Err(TaxoError::InternalError(InternalError::GenericError(format!("Unsupported OZW type: {:?}", vid.get_type())))) }
     };
 
-    result.map_err(|e| TaxoError::InternalError(InternalError::GenericError(format!("Error while setting a value: {}", e))))
+    result.map_err(|e| {
+        TaxoError::InternalError(InternalError::GenericError(format!("Error while setting a \
+                                                                      value: {}",
+                                                                     e)))
```
julienw: here as well

components/openzwave-adapter/src/lib.rs
```diff
@@ -262,94 +294,106 @@ impl OpenzwaveAdapter {
...
                         node_map.push(service_id.clone(), node);
 
                         let mut service = Service::empty(service_id.clone(), adapter_id.clone());
                         service.properties.insert(String::from("name"), node.get_name());
-                        service.properties.insert(String::from("product_name"), node.get_product_name());
-                        service.properties.insert(String::from("manufacturer_name"), node.get_manufacturer_name());
+                        service.properties
+                            .insert(String::from("product_name"), node.get_product_name());
+                        service.properties.insert(String::from("manufacturer_name"),
+                                                  node.get_manufacturer_name());
```
julienw, line 404: why doesn't the tool do the same thing in the 2 previous lines ?

components/openzwave-adapter/src/lib.rs
```diff
@@ -262,94 +294,106 @@ impl OpenzwaveAdapter {
...
                         });
                     }
-                    ZWaveNotification::NodeNaming(_node)             => {
+                    ZWaveNotification::NodeNaming(_node) => {
                         // unfortunately we can't change a service' properties :(
                         // https://github.com/fxbox/taxonomy/issues/97
                         // When it's done we can move the properties change from above to here.
                     }
-                    ZWaveNotification::NodeRemoved(_node)           => {}
-                    ZWaveNotification::ValueAdded(vid)              => {
```
julienw, line 379: we tried to have everything aligned here :(

components/openzwave-adapter/src/lib.rs
```diff
@@ -360,44 +404,68 @@ impl OpenzwaveAdapter {
                         };
 
                         for &(ref range, ref sender) in &watchers {
-                            debug!("[OpenzwaveAdapter::ValueChanged] Iterating over watcher {:?} {:?}", taxo_id, range);
+                            debug!("[OpenzwaveAdapter::ValueChanged] Iterating over watcher {:?} \
+                                    {:?}",
```
julienw, line 502: not sure this is good for readability :(

components/openzwave-adapter/src/lib.rs
fabricedesre, line 404: Not sure, but maybe because of the line length.

PR #486, opened 2016-05-16
Add support for multiple zwave-dongles.

dhylands:
This also needs: https://github.com/fxbox/openzwave-stateful-rust/pull/19

r? @julienw


dhylands:
Travis fails since https://github.com/fxbox/openzwave-stateful-rust/pull/19 is missing.

components/openzwave-adapter/src/lib.rs
```diff
@@ -204,12 +204,13 @@ fn ensure_directory<T: AsRef<Path> + ?Sized>(directory: &T) -> Result<(), Error>
...
 impl OpenzwaveAdapter {
-    pub fn init<T: AdapterManagerHandle + Send + Sync + 'static> (box_manager: &Arc<T>, user_path: &str, device: Option<String>) -> Result<(), Error> {
+    pub fn init<T: AdapterManagerHandle + Send + Sync + 'static> (box_manager: &Arc<T>, user_path: &str, devices: Option<String>) -> Result<(), Error> {
 
         try!(ensure_directory(user_path));
 
         let options = InitOptions {
-            device: device,
+            // We treat devices as a whitespace delimited list of device names.
+            devices: devices.map(|s| s.split_whitespace().map(|s| s.to_owned()).collect()),
```
julienw: I just wonder why whitespace and not eg comma + space trim ? That seems more natural to me (but maybe that's just me).

also, I wonder if `.cloned()` would work instead of your `.map(...)` ?

src/adapters/mod.rs
```diff
@@ -66,8 +66,10 @@ impl<T: Controller> AdapterManager<T> {
         let scripts_path = &self.controller.get_profile().path_for("thinkerbell_scripts.sqlite");
         ThinkerbellAdapter::init(manager, scripts_path).unwrap(); // FIXME: no unwrap!
         let profile_openzwave = &self.controller.get_profile().path_for("openzwave");
-        let openzwave_device = self.controller.clone().get_config().get("openzwave", "device");
-        OpenzwaveAdapter::init(manager, profile_openzwave, openzwave_device).unwrap();
+
+        // device is for backwards compatability
```
julienw: wondering if we need to bother about backward compatibility (BTW: it's `backward compatibility` :) ) now or if a mail on the ML would be enough :)

julienw:
one nit and one question about the backward compatibility handling. r=me with this

components/openzwave-adapter/src/lib.rs
dhylands: I couldn't get anything to compile. .cloned isn't implemented for str
split_whitespace essentially returns an Option<&str> which is why I need to use map. .cloned seems to be associated with .iter, so I'm not quite sure how it can be worked into this.

julienw:
r=me if you tested :)

dhylands:
> r=me if you tested :)

I've tested with my one dongle. And I've tested with an FTDI serial attached to  /dev/ttyUSB0 and my dongle showing as /dev/ttyACM0 (previously it would pick the FTDI dongle which doesn't do zwave).

I won't be able to test with multiple dongles until we get to London.

julienw:
OK I'll test this today and merge if it works properly.

dhylands:
Don't merge it yet. I've been working on getting a successful test run and ran into a variety of issues. I was going to merge once I get a clean test run.

PR #487, opened 2016-05-17
Fix tag management API routes. r=yoric

azasypkin:
API endpoints that are used for `payload_api2!` macros don't correspond to authentication (in the same file) and cors-related endpoints (`http_server.rs`).

### DIFF
--- a/src/http_server.rs
+++ b/src/http_server.rs
@@ -91,8 +91,8 @@ impl<T: Controller> HttpServer<T> {
             (vec![Method::Get, Method::Post], "api/v1/channels/setters".to_owned()),
             (vec![Method::Put], "api/v1/channels/get".to_owned()),
             (vec![Method::Put], "api/v1/channels/set".to_owned()),
-            (vec![Method::Post, Method::Delete], "api/v1/channel/getters/tags".to_owned()),
-            (vec![Method::Post, Method::Delete], "api/v1/channel/setters/tags".to_owned())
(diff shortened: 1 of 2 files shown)


azasypkin:
@Yoric r? whenever you have time.

Yoric:
I'm afraid that this leaves the file as inconsistent as it was. We still have instances of `"channel"` and instances of `"channels"`. Could you fix all and synchronize with @gmarty to adapt the front-end?

azasypkin:
> I'm afraid that this leaves the file as inconsistent as it was. We still have instances of "channel" and instances of "channels". 

Definitely, do you want me to update all places to use either `channel` or `channels` everywhere or ...?

Like this:
- `channels/getters`;
- `channels/get`;
- `channels/getters/tags`.

> Could you fix all and synchronize with @gmarty to adapt the front-end?

No worries, I'm fixing UI at the same time in https://github.com/fxbox/app/pull/150 :) That's how I've stumbled upon this issue.

Yoric:
> Definitely, do you want me to update all places to use either channel or channels everywhere or ...?

That would be great. Let's make everything plural.

azasypkin:
> That would be great. Let's make everything plural.

Okay, now everything is in plural form :) r? @Yoric 

If you want you can test it using the app [1], just open any service view and tap on "tag" icon in the header.

[1] https://azasypkin.github.io/app

Yoric:
r=me

azasypkin:
Thanks!

PR #491, opened 2016-05-20
Decentralizing: Making it possible for adapters to manipulate Payload instead of Value

Yoric:
This has two aims:
1. this will help us progressively remove the big centralized enums (`ChannelKind`, `Type`, `Value`);
2. this will help adapter developers write off-process adapters.

### DIFF
--- a/components/openzwave-adapter/src/lib.rs
+++ b/components/openzwave-adapter/src/lib.rs
@@ -11,7 +11,7 @@ mod watchers;
 use taxonomy::util::Id as TaxoId;
 use taxonomy::services::{ AdapterId, ServiceId, Service, Channel, ChannelKind };
 use taxonomy::values::*;
-use taxonomy::api::{ ResultMap, Error as TaxoError, InternalError, User };
+use taxonomy::api::{ Operation, ResultMap, Error as TaxoError, InternalError, User };
 use taxonomy::adapter::{ AdapterManagerHandle, AdapterWatchGuard, WatchEvent };
 use transformable_channels::mpsc::ExtSender;
 
@@ -473,7 +473,7 @@ impl taxonomy::adapter::Adapter for OpenzwaveAdapter {
 
                 ozw_vid_as_taxo_value(&ozw_vid)
             });
-            let value_result: Result<Option<Value>, TaxoError> = taxo_value.ok_or(TaxoError::GetterDoesNotSupportPolling(id.clone()));
+            let value_result: Result<Option<Value>, TaxoError> = taxo_value.ok_or(TaxoError::OperationNotSupported(Operation::Fetch, id.clone()));
             (id, value_result)
         }).collect()
     }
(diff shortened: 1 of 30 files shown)


cr:
The Hue integration tests show some breakage: https://travis-ci.org/fxbox/foxbox/jobs/131698779#L1749

cr:
The Hue changes are trivial. r+ from my side for that.

I can't see how this is breaking integration tests, so the issue referred to in https://github.com/fxbox/foxbox/pull/491#issuecomment-220622657 is probably unrelated.

components/taxonomy/src/adapter.rs
```diff
@@ -86,22 +87,59 @@ pub trait AdapterManagerHandle: Send {
...
+pub trait RawAdapter: Send + Sync {
+    /// An id unique to this adapter. This id must persist between
+    /// reboots/reconnections.
+    fn id(&self) -> Id<AdapterId>;
+
+    fn fetch_values(&self, mut target: Vec<(Id<Channel>, Type)>, _: User) -> ResultMap<Id<Channel>, Option<(Payload, Type)>, Error> {
+        target.drain(..).map(|(id, _)| {
+            (id.clone(), Err(Error::OperationNotSupported(Operation::Watch, id)))
+        }).collect()
+    }
```
fabricedesre, line 121: nit: blank like between methods. Also, this file has license header.

fabricedesre:
r=me on the taxonomy_router changes.

components/openzwave-adapter/src/lib.rs
```diff
@@ -490,11 +490,11 @@ impl taxonomy::adapter::Adapter for OpenzwaveAdapter {
         }).collect()
     }
 
-    fn register_watch(&self, mut values: Vec<(TaxoId<Channel>, Option<Value>, Box<ExtSender<WatchEvent>>)>) -> Vec<(TaxoId<Channel>, Result<Box<AdapterWatchGuard>, TaxoError>)> {
+    fn register_watch(&self, mut values: Vec<(TaxoId<Channel>, Option<Value>, Box<ExtSender<WatchEvent<Value>>>)>) -> Vec<(TaxoId<Channel>, Result<Box<AdapterWatchGuard>, TaxoError>)> {
```
julienw, line 493: Because we use `WatchEvent<Value>` everywhere, maybe it would make sense to either not change this now (maybe you'll need it later, then change it later ? or even in a separate patch if you really want it), or add a type alias.

julienw:
r=me for openzwave, I only left a nit comment.

dhylands:
r=me for ip_camera changes

components/openzwave-adapter/src/lib.rs
Yoric, line 493: Unfortunately, an alias wouldn't work. If you do `type Foo = WatchEvent<Value>`, you cannot use `Foo::Enter` and `Foo:Exit`, so this would make things more complicated.

Also, there are already a few uses of `WatchEvent<T>` where `T` is not `Value` (well, only one at the moment). Defining two different types with the same structure didn't seem like a good use of Rust's ADT.

PR #495, opened 2016-05-23
Move reusable pieces to their own 'foxbox_core' crate

fabricedesre:


fabricedesre:
This PR moves in their own crate a few files that provide reusable features. That will let them be accessible from other components/ crates.

I don't know why we fail to cross-compile... @JohanLorenzo does that ring a bell? if not I'll check what happens with a simpler project that uses serde_macros as this looks like the pain point.

JohanLorenzo:
Sorry, I've never seen this failure before. One thing that caught my attention: serde_macros is correctly downloaded and compiled. foxbox_taxonomy seems correctly compiled too (and they both use serde_macros 0.7.5). Maybe we could change the [cargo build rule](https://github.com/fxbox/foxbox/blob/master/tools/travis-linux-arm_cross_compile.sh#L74) into `cargo build -j1 --verbose --target="$RUST_TARGET"`? That would enable us to see if foxbox_taxonomy is indeed compiled and what's the compiling target.

fabricedesre:
Ok, so I actually didn't need serde_macros in the core/ crate.

PR #499, opened 2016-05-24
Reactivate and fix Thinkerbell intermittent

JohanLorenzo:
After some investigation done with @Yoric, we've managed to hunt down the intermittent. Root cause: [register_watch()](https://github.com/fxbox/foxbox/blob/1e093b03c3cda929abdf5c256baa0d1d3343093c/components/thinkerbell/src/fake_env.rs#L86) was sometimes called after the [exec.start()](https://github.com/fxbox/foxbox/blob/1e093b03c3cda929abdf5c256baa0d1d3343093c/components/thinkerbell/tests/test_rules.rs#L139) was done.

This occurs because exec.start() claims to be done right before [exec.run() is executed](https://github.com/fxbox/foxbox/blob/1e093b03c3cda929abdf5c256baa0d1d3343093c/components/thinkerbell/src/run.rs#L84-L85). But watchers are actually registered within exec.run(). So, like @Yoric suggested, we can just emit the init done event right after watchers are added.

I was able to reproduce the bug locally around 1/10 times. As it used to hang, the first failure was easily "spottable" with

``` sh
cd components/thinkerbell
for i in {0..100}; do echo $i && cargo test -- --nocapture; done
```

Once the patch applied, I ran twice the command above, and I didn't get any single failure. I feel confident in this patch. 

One last note: I left some commits that helped me narrowing down the problem (travis configurations, logs everywhere). They've been reverted so the final diff just contains the fix and the reactivation. I left those commits for archive-purpose, if something similar comes again, one day.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: required
 
 os:
+  - osx
   - linux
 
 dist: trusty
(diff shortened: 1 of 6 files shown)


JohanLorenzo:
@Yoric was sitting next to me and confirmed [his last comment](https://github.com/fxbox/foxbox/pull/499#commitcomment-17613093) was equivalent to an r+. Landing.

PR #505, opened 2016-05-26
Basic multi-user UI. r=fabrice

ferjm:


components/thinkerbell/src/manager.rs
```diff
@@ -76,7 +76,7 @@ impl<Env, T> ScriptManager<Env, T>
             id          TEXT NOT NULL PRIMARY KEY,
             source      TEXT NOT NULL,
             is_enabled  BOOL NOT NULL DEFAULT 1,
-            owner       INTEGER NOT NULL DEFAULT -1
+            owner       TEXT
```
fabricedesre, line 79: We don't care yet, but we really need to plan for db updates...

src/adapters/webpush/mod.rs
```diff
@@ -42,7 +42,7 @@ static ADAPTER_NAME: &'static str = "WebPush adapter (built-in)";
 static ADAPTER_VENDOR: &'static str = "team@link.mozilla.org";
 static ADAPTER_VERSION: [u32;4] = [0, 0, 0, 0];
 // This user identifier will be used when authentication is disabled.
-static NO_AUTH_USER_ID: i32 = -1;
+static NO_AUTH_USER_ID: &'static str = "-1";
```
fabricedesre: this looks pretty hackish... We should stick with None instead.

fabricedesre:
There are a few things I'd like to see improve:
- webpush using None instead of "-1"
- more comments in the js frontend code. I glanced over, and there are parts where tbh I'm not sure what's happening.

Follow ups to keep in mind:
- moving the User type in foxbox_users and use it elsewhere.
- clean db upgrades (it's not limited to webpush though).
- l10n for the frontend.

ferjm:
> moving the User type in foxbox_users and use it elsewhere.

Actually, I'd prefer to make this simply an Option<String> so we don't create unnecessary dependencies (i.e. taxonomy with users).

ferjm:
@fabricedesre could you take another look at this, please?

src/adapters/webpush/db.rs
```diff
@@ -159,64 +168,65 @@ pub fn remove_test_db() {
...
     before_each {
+        use foxbox_taxonomy::api::User;
         let db = WebPushDb::new(&get_db_environment());
     }
 
     it "should manage subscription correctly" {
         use super::super::Subscription;
 
-        let subs0 = db.get_subscriptions(1).unwrap();
+        let subs0 = db.get_subscriptions(&User::Id("1".to_owned())).unwrap();
```
fabricedesre: nit: you're using String::from("1") everywhere else.

fabricedesre:
r=me thanks!

PR #508, opened 2016-05-30
Include the authentication checking feature for the hue simulator

nojunpark:


nojunpark:
@JohanLorenzo r?

test/integration/lib/foxbox_process_manager.js
```diff
@@ -35,6 +35,12 @@ var foxbox_process_manager = (function() {
       'User DB not found!');
     _removeFileIfItExists('/.local/share/foxbox/webpush.sqlite',
       'webpush DB not found!');
+    _removeFileIfItExists('/.local/share/foxbox/foxbox.conf',
+      'Foxbox config file not found!');
+    _removeFileIfItExists('/.local/share/foxbox/taxonomy_tags.sqlite',
```
JohanLorenzo: Nit: Rename the function => We don't only delete databases anymore. I wonder if we shouldn't delete every file under `.local/share` except the certificates. Here's a way to do it https://github.com/fxbox/foxbox/blob/master/test/selenium/lib/foxbox_process_manager.js#L44

test/integration/lib/make_suite.js
```diff
@@ -38,9 +38,13 @@ var make_suite = (function() {
       config.get('nupnp_server.port'));
   }
 
-  function turnOnHue(){
+  function turnOnHue(authentication){
+    if (typeof authentication === 'undefined') {
```
JohanLorenzo: Nit: Can be simplified with `authentication = authentication || false`.

test/integration/lib/make_suite.js
```diff
@@ -104,7 +108,7 @@ var make_suite = (function() {
 
   return { makeSuite, turnOnAllSimulators,
     philipshue_server,ipcamera_server, webPush_server,
-    foxboxManager, turnOnFoxbox};
+    foxboxManager, turnOnFoxbox,turnOnHue,turnOnHueNupnp};
```
JohanLorenzo: Nit: Spaces between objects' arguments.

test/integration/lib/philipsHue_server.js
```diff
@@ -73,80 +75,20 @@ var philipsHue_server = (function() {
...
-    _server.get('/api/:foxboxId/lights', function (req, res) {
-      if (req.params.foxboxId === foxboxId){ 
-        res.status(200).json(light_status); 
-      }
-      else {
-        res.status(404); 
-        throw 'foxboxId mismatch: got ' + req.params.foxboxId; 
-      }
-    });
+  function pressButton() {
```
JohanLorenzo, line 80: I'm not sure what's the intent behind this function. Maybe a comment would help understanding it

test/integration/lib/philipsHue_server.js
```diff
@@ -73,80 +75,20 @@ var philipsHue_server = (function() {
...
-          light_status[req.params.light_id]);
-        last_received_cmd += ':' + JSON.stringify(req.body);
-        last_commanded_light = req.params.light_id;
-        res.status(200).json(response);     
-      });
+  function setup(port,authentication) {
+    // Reset variables, since the instance gets 'reused' between tests
+    light_status = api_resp.lights; 
+    foxboxId = undefined;
+    server_port = port;
```
JohanLorenzo, line 91: I don't think this line and the next one reset anything, do they?

nojunpark, line 91: actually it does, the server_port and auth_status are passed to the next test.

test/integration/lib/philipsHue_server.js
nojunpark, line 80: sure, I'll put the comment.  it mimics the button on the philips hub, after the button press, it assigns the username to the requester.

test/integration/test/ip_camera_test.js
```diff
@@ -74,7 +74,8 @@ Prepper.makeSuite('Control camera locally',function(){
   it('take a picture (using feature)',function(){
     var setter = 'channel:' +
       cameraService.id.replace('service:','snapshot.');
-    var payload = {'select': {'feature': "camera/store-snapshot"}, 'value': {'Unit': {}}};
+    var payload = {'select': {'feature': 'camera/store-snapshot'},
```
JohanLorenzo, line 77: Nit: Confusing indentation

test/integration/test/ip_camera_test.js
```diff
@@ -102,9 +103,8 @@ Prepper.makeSuite('Control camera locally',function(){
 
   // Checks the correct size of bytes is received
   it('download the picture (using feature)',function(){
-    var getter = 'channel:' +
```
JohanLorenzo: What don't we need this getter anymore?

test/integration/test/philips_light_test.js
```diff
@@ -18,6 +19,7 @@ Prepper.makeSuite('Control lights locally',function(){
     // collect all getters for the lightbulbs
     return chakram.put(Prepper.foxboxManager.getterURL,getterPayload)
     .then(function(listResponse) {
+      console.log(listResponse.body);
```
JohanLorenzo: Nit: Do we need this debug trace, here?

nojunpark: oh no, I thought I removed this, thanks!

test/integration/test/philips_authentication_test.js
```diff
@@ -0,0 +1,46 @@
...
+    setTimeout(done,5000);
+  });
+
+  it('Send light query without authentication',function(){
+  
+    // collect all getters for the lightbulbs
+    return chakram.put(Prepper.foxboxManager.getterURL,getterPayload)
+    .then(function(listResponse) {
+      lights = Object.keys(listResponse.body);
+      expect(lights.length).equals(0);
```
JohanLorenzo, line 26: Do you think it's worth to check the HTTP response code?

test/integration/test/philips_authentication_test.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+  });
+
+  describe('Authenticate with Philips Hue', function () {
+    before('Press the button', function(done) {
+      Prepper.philipshue_server.pressButton();
+      setTimeout(done,5000); //give time to collect light info
```
JohanLorenzo: I'd put this wait in `pressButton()` and wrap it up in a promise. Otherwise, we're gonna need to duplicate that setTimeout in every suite

test/integration/test/philips_authentication_test.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+
+    it('Send light query after authentication',function(){
+  
+      // collect all getters for the lightbulbs
+      return chakram.put(Prepper.foxboxManager.getterURL,getterPayload)
+      .then(function(listResponse) {
+        lights = Object.keys(listResponse.body);
+        expect(lights.length).equals(3);
```
JohanLorenzo, line 42: Should we check the HTTP response code, here too?

test/integration/test/philips_light_test.js
```diff
@@ -79,7 +81,8 @@ Prepper.makeSuite('Control lights locally',function(){
 
     it('Turn off all lights at once', function() {
 
-      var payload = {'select': {'feature': 'light/is-on'}, 'value': {'OnOff':'Off'}};
+      var payload = {'select': {'feature': 'light/is-on'}, 
+      'value': {'OnOff':'Off'}};
```
JohanLorenzo: Nit: Confusing indentation

test/integration/test/philips_authentication_test.js
nojunpark, line 26: ah, right, maybe we could check for 200

test/integration/lib/philipsHue_server.js
```diff
@@ -160,9 +102,112 @@ var philipsHue_server = (function() {
...
+    last_received_cmd = req.method + ':' + req.originalUrl;
+    next();
+  });
+
+  _server.get('/', function (req, res) {
+    res.status(200).sendFile(
+      path.join(__dirname,'/html/philips_initial.html'));  
+  });
+
+  _server.get('/api/:foxbox/', function (req, res) {
```
JohanLorenzo: Nit: `:foxboxId`, otherwise the first if is always true.

test/integration/lib/philipsHue_server.js
```diff
@@ -160,9 +102,112 @@ var philipsHue_server = (function() {
...
+    res.status(200).sendFile(
+      path.join(__dirname,'/html/philips_initial.html'));  
+  });
+
+  _server.get('/api/:foxbox/', function (req, res) {   
+    
+     if (foxboxId === undefined) {
+      if (auth_status === false) {
+        foxboxId = req.params.foxbox;
+        res.sendStatus(200);
```
JohanLorenzo, line 131: Should we send `api_resp`, here too?

test/integration/lib/philipsHue_server.js
```diff
@@ -160,9 +102,112 @@ var philipsHue_server = (function() {
...
+              }
+          }]);
+      }     
+    }
+    else {
+      res.status(200).json(api_resp);  
+    }     
+  });
+
+  _server.post('/api', bodyParser.json(),function (req, res) {
```
JohanLorenzo, line 149: What does `bodyParser.json()` do?

test/integration/lib/philipsHue_server.js
```diff
@@ -160,9 +102,112 @@ var philipsHue_server = (function() {
...
+    
+    if (foxboxId === undefined) {
+        res.status(200).send([{
```
JohanLorenzo, line 152: Should we send 200, even though the request failed?

test/integration/lib/philipsHue_server.js
nojunpark, line 149: ah, without this, express can't see the body of the received message's response.

test/integration/lib/philipsHue_server.js
```diff
@@ -160,9 +102,112 @@ var philipsHue_server = (function() {
...
+                'address': '',
+                'description': 'link button not pressed'
+            }
+        }]);
+      }
+      else {
+        if (req.body.devicetype === 'foxbox_hub') {
+          res.status(200).send([{
+            'success': {
+            'username': 'simulatedPhilipsHub'
```
JohanLorenzo: Nit: Confusing indentation

test/integration/lib/philipsHue_server.js
```diff
@@ -160,9 +102,112 @@ var philipsHue_server = (function() {
...
+    if (req.params.foxboxId === foxboxId){ 
+      res.status(200).json(light_status); 
+    }
+    else {
+      res.status(404); 
+      throw 'foxboxId mismatch: got ' + req.params.foxboxId; 
+    }
+  });
+
+  _server.get('/api/:foxboxId/lights/:light_id', function (req, res) {
```
JohanLorenzo: Nit: Mixed camel/snake case. I'd put `:lightId`

test/integration/lib/philipsHue_server.js
```diff
@@ -160,9 +102,112 @@ var philipsHue_server = (function() {
...
+        }
+      }
+  });
+
+  _server.get('/api/:foxboxId/lights', function (req, res) {
+    if (req.params.foxboxId === foxboxId){ 
+      res.status(200).json(light_status); 
+    }
+    else {
+      res.status(404);
```
JohanLorenzo, line 176: I'd centralize foxboxId check in one central place, to avoid duplication. For instance:

``` js
_server.get('/api/:foxboxId/lights', function (req, res) {
  checkFoxboxId(req, res);
  res.status(200).json(light_status); 
}

checkFoxboxId(req, res) {
  if (req.params.foxboxId !== foxboxId) {
    res.status(404);
   throw 'foxboxId mismatch: got ' + req.params.foxboxId; 
  }
} 
```

test/integration/lib/philipsHue_server.js
JohanLorenzo, line 91: Oh, I didn't see the scope of the variable. Got it! Thanks

test/integration/lib/philipsHue_server.js
nojunpark: oh no actually, the foxbox is what the Foxbox sends as its username, which is always 'unauthorized'.  this gets checks in req.params.foxbox.

test/integration/lib/philipsHue_server.js
nojunpark, line 131: actually it will request again if we don't send the api_resp here, and then we'll send the api_resp.  I'm checking to see whether the foxbox retries properly.

test/integration/lib/philipsHue_server.js
nojunpark, line 152: yes, the response is 200, but the content says there was an error, since the HTTP request didn't fail.

test/integration/lib/philipsHue_server.js
nojunpark, line 176: but for each case, the response contents are slightly different.  since there are only 2 places that does this check, it think it is more readable unless we are expanding the checks to other methods.

test/integration/test/philips_authentication_test.js
nojunpark: hmm I did it like below, but it doesn't seem to wait at all:
 function pressButton() {
    foxboxId = 'simulatedPhilipsHub';

```
return new Promise(resolve => {
  setTimeout(resolve, 5000);
});
```

  }

Could you tell me what am I doing wrong here?

test/integration/test/philips_authentication_test.js
```diff
@@ -0,0 +1,48 @@
...
+      lights = Object.keys(listResponse.body);
+      expect(lights.length).equals(0);
+      expect(listResponse).to.have.status(200);
+    });
+  });
+
+  describe('Authenticate with Philips Hue', function () {
+    before('Press the button', function(done) {
+      Prepper.philipshue_server.pressButton();
+      setTimeout(done, 5000);
```
nojunpark: so for this, could you give me a suggestion?  It seems like setTimeout gets ignored when I put it inside the pressButton method, like below, but it didn't wait at all.
function pressButton() {
    foxboxId = 'simulatedPhilipsHub';
     return new Promise(resolve => {
          setTimeout(resolve, 3000);
    });
  }

test/integration/lib/philipsHue_server.js
```diff
@@ -160,9 +104,112 @@ var philipsHue_server = (function() {
...
+
+  _server.get('/api/:foxboxId/', function (req, res) {   
+      
+     if (foxboxId === undefined) {
+      if (auth_status === false) {
+        foxboxId = req.params.foxboxId;
+        res.sendStatus(200);  
+      }
+      else if (auth_status === true) {
+        res.status(200).send([{
```
JohanLorenzo, line 135: Follow up from https://github.com/fxbox/foxbox/pull/508#discussion_r65592362

Even though the HTTP request was valid, I believe you need to show there is something unexpected in the request. In this case, the user is forbidden the access, which corresponds to [an error 403](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_Error). 

If the HTTP request failed for an unlisted reason, then it would be a generic 400, IMO.

test/integration/lib/philipsHue_server.js
JohanLorenzo, line 176: Ok. Let's postpone that change until the server becomes bigger

test/integration/test/philips_authentication_test.js
JohanLorenzo: I don't see anything wrong in this snippet. Maybe the before part contains an error? It should be close to

``` js
before('Press the button', function() {
  return Prepper.philipshue_server.pressButton();
})
```

(notice the `done` disappearing and `return` fading in.)

test/integration/test/philips_authentication_test.js
JohanLorenzo: See https://github.com/fxbox/foxbox/pull/508#discussion_r65676418

test/integration/lib/philipsHue_server.js
nojunpark, line 135: You're correct, but strangely, this is how Philips hue hub sends out the response - it gives 200 OK, yet returns an error msg.  I thought this was odd too.

JohanLorenzo, line 135: Okay, in this case, just add a comment saying we mimic philips hue, but we know it's weird

test/integration/test/philips_authentication_test.js
nojunpark, line 42: added

test/integration/test/philips_authentication_test.js
nojunpark: that worked! thanks for reminding me to bubble up the promise object

test/integration/test/ip_camera_test.js
nojunpark, line 77: fixed

JohanLorenzo:
Looks good to me, r+!

Issue #519, opened 2016-06-08
Thinkerbell rules should be delivered as JSON, not strings


Yoric:

Yoric:
I've started working on this. The main difficulty is that we need to write a serializer for scripts, which is a bit time-consuming. Other than that, this shouldn't be problematic.

PR #520, opened 2016-06-08
Decentralizing: Removing big centralized Value

Yoric:
This PR removes the big centralized enum `Value`, effectively ending all centralization in Taxonomy. The first two beneficiaries are Thinkerbell (which now defines its own value type `RuleSource`) and WebPush (which now defines its own value type `WebPushNotify`).
- [x] r? @fabricedesre for Foxbox.
- [x] r? @mcav for Thinkerbell.
- [x] r? @julienw for ZWave.
- [x] r? @cr for Philps Hue.
- [x] r? @dhylands for IP Camera.
- [x] r? @aosmond for WebPush.
- [x] r? @fabricedesre for TTS.
- [x] r? @npark-mozilla for integration tests.
- [x] f? @azasypkin for synchronization with front-end.

I still have a few PRs planned after this lands, but at this stage, Taxonomy will be Decentralized \o/

### DIFF
--- a/components/openzwave-adapter/src/lib.rs
+++ b/components/openzwave-adapter/src/lib.rs
@@ -148,7 +148,7 @@ fn ozw_vid_as_taxo_value(vid: &ValueID) -> Option<Value> {
                 if ref_eq(kind, &DOOR_IS_OPEN) {
                     Some(Value::new(if value {OpenClosed::Open} else {OpenClosed::Closed}))
                 } else if ref_eq(kind, &DOOR_IS_LOCKED) {
-                    Some(Value::new(if value {DoorLocked::Locked} else {DoorLocked::Unlocked}))
+                    Some(Value::new(if value {IsLocked::Locked} else {IsLocked::Unlocked}))
                 } else {
                     None
                 }
@@ -170,8 +170,8 @@ fn set_ozw_vid_from_taxo_value(vid: &ValueID, value: Value) -> Result<(), TaxoEr
         ValueType::ValueType_Bool => {
             if let Some(open_closed) = value.downcast::<OpenClosed>() {
                 vid.set_bool(*open_closed == OpenClosed::Open)
-            } else if let Some(locked_unlocked) = value.downcast::<DoorLocked>() {
-                vid.set_bool(*locked_unlocked == DoorLocked::Locked)
+            } else if let Some(locked_unlocked) = value.downcast::<IsLocked>() {
+                vid.set_bool(*locked_unlocked == IsLocked::Locked)
             } else {
                 return Err(TaxoError::InvalidValue) // TODO InvalidType would be better but we'll need to fix specific types for specific TaxoIds
             }
--- a/components/openzwave-adapter/src/lib.rs
+++ b/components/openzwave-adapter/src/lib.rs
@@ -107,18 +107,18 @@ trait RangeChecker {
     fn should_send(&self, &Value, EventType) -> bool;
 }
 
-impl RangeChecker for Option<Box<Range>> {
+impl RangeChecker for Option<Value> {
     fn should_send(&self, value: &Value, event_type: EventType) -> bool {
         match *self {
             None => event_type == EventType::Enter, // no range means we send only Enter events
-            Some(ref range) => range.contains(value)
+            Some(ref range) => range == value // FIXME: This won't scale up to interesting ranges.
         }
     }
 }
 
-impl RangeChecker for Range {
+impl RangeChecker for Value {
     fn should_send(&self, value: &Value, _: EventType) -> bool {
-        self.contains(value)
+        self == value // FIXME: This won't scale up to interesting ranges.
     }
 }
 
@@ -146,9 +146,9 @@ fn ozw_vid_as_taxo_value(vid: &ValueID) -> Option<Value> {
                     return None;
                 };
                 if ref_eq(kind, &DOOR_IS_OPEN) {
-                    Some(Value::OpenClosed(if value {OpenClosed::Open} else {OpenClosed::Closed}))
+                    Some(Value::new(if value {OpenClosed::Open} else {OpenClosed::Closed}))
                 } else if ref_eq(kind, &DOOR_IS_LOCKED) {
-                    Some(Value::DoorLocked(if value {DoorLocked::Locked} else {DoorLocked::Unlocked}))
+                    Some(Value::new(if value {DoorLocked::Locked} else {DoorLocked::Unlocked}))
                 } else {
                     None
                 }
@@ -168,11 +168,12 @@ fn set_ozw_vid_from_taxo_value(vid: &ValueID, value: Value) -> Result<(), TaxoEr
 
     let result = match vid.get_type() {
         ValueType::ValueType_Bool => {
-            match value {
-                //Value::OnOff(onOff) => { vid.set_bool(onOff == OnOff::On) } // TODO support switches
-                Value::OpenClosed(open_closed) => { vid.set_bool(open_closed == OpenClosed::Open) }
-                Value::DoorLocked(locked_unlocked) => { vid.set_bool(locked_unlocked == DoorLocked::Locked) }
-                _ => { return Err(TaxoError::InvalidValue) } // TODO InvalidType would be better but we'll need to fix specific types for specific TaxoIds
+            if let Some(open_closed) = value.downcast::<OpenClosed>() {
+                vid.set_bool(*open_closed == OpenClosed::Open)
+            } else if let Some(locked_unlocked) = value.downcast::<DoorLocked>() {
+                vid.set_bool(*locked_unlocked == DoorLocked::Locked)
+            } else {
+                return Err(TaxoError::InvalidValue) // TODO InvalidType would be better but we'll need to fix specific types for specific TaxoIds
             }
         }
         _ => { return Err(TaxoError::InternalError(InternalError::GenericError(format!("Unsupported OZW type: {:?}", vid.get_type())))) }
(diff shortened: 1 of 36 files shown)


aosmond:
r=me for WebPush.

mcav:
thinkerbell lgtm

components/taxonomy/src/lib.rs
```diff
@@ -4,6 +4,8 @@
 #![plugin(clippy)]
 // To prevent clippy being noisy with derive(...)
 #![allow(used_underscore_binding)]
+#![allow(let_unit_value)] // For some reason, clippy decides to display this warning, without
```
fabricedesre: nit: finish that sentence and remove extra blank line.

components/taxonomy/src/values.rs
```diff
@@ -51,6 +55,228 @@ impl error::Error for TypeError {
...
+    fn serialize(source: &Self, binary: &BinaryTarget) -> Result<JSON, Error> {
+        
+        match *source {
+            None => Ok(JSON::Null),
+            Some(ref val) => T::serialize(val, binary)
+        }
+    }
+}
+
+*/
```
fabricedesre: Do we want to keep this commented code here?

fabricedesre:
r=me thanks!

src/adapters/ip_camera/upnp_listener.rs
```diff
@@ -54,13 +54,12 @@ impl UpnpListener for IpCameraUpnpListener {
...
-        let mut udn = try_get!(service.description, "/root/device/UDN").clone();
+        let udn = try_get!(service.description, "/root/device/UDN")
         // The UDN is typically of the for uuid:SOME-UID-HERE, but some devices
         // response with just a UUID. We strip off the uuid: prefix, if it exists
         // and use the resulting UUID as the service id.
-        if udn.starts_with("uuid:") {
-            udn = String::from(&udn[5..]);
-        }
+            .trim_left_matches("uuid:")
+            .to_owned();
```
dhylands, line 62: nit: Thanks for the rewrite. I'd probably put the comment in front of the let. I think the comment in the middle of a statement looks funny.

dhylands:
r=me for camera stuff.

src/adapters/ip_camera/upnp_listener.rs
Yoric, line 62: Note that I rewrote that only because Clippy kept pestering me about the style of this statement.

dhylands, line 62: I filed a bug with clippy, and it's been fixed in the latest version (not sure when it will be released).
See: https://github.com/Manishearth/rust-clippy/issues/975

Looks like it should be fixed in release 0.0.72

components/taxonomy/src/values.rs
Yoric: Oops, good catch.

Yoric:
Adding:
- r? @fabricedesre for TTS (which I had forgotten in the previous PR);
- Cc @azasypkin because of upcoming breakages.

Yoric:
Adding r? @npark-mozilla for integration tests.

Also, @azasypkin , you can now look at the integration tests to see the API changes.

cr:
r+ for the Hue changes and the changes to the Color type.

components/openzwave-adapter/src/lib.rs
```diff
@@ -107,18 +107,18 @@ trait RangeChecker {
...
+            Some(ref range) => range == value // FIXME: This won't scale up to interesting ranges.
         }
     }
 }
 
-impl RangeChecker for Range {
+impl RangeChecker for Value {
     fn should_send(&self, value: &Value, _: EventType) -> bool {
-        self.contains(value)
+        self == value // FIXME: This won't scale up to interesting ranges.
```
julienw, line 121: what's the path forward with this FIXME? Maybe you could file an issue and put the issue # here and above ?

Yoric, line 121: Well, this `FIXME` is actually a reminder that `should_send` only works for simple cases. For more complicated cases, well, a different algorithm will be needed. As I have no clue about which cases we are going to handle in the future, I don't really see how I can be much more specific.

azasypkin:
f+ from the client side (app related PR is https://github.com/fxbox/app/pull/172)!

Just a few notes (maybe they are the same on master though, didn't check):

1) When incorrect rule is saved (e.g. without `execute` part), there is very vague error is returned `{"thinkerbell-add-rule":{"Error":"Internal Error"}}`; 

2) It was there before, but since you're updating related pieces: it's a bit hard to distinguish successful response with value and failed one (especially for batch requests where response code should be 200 in any case):
- `{"OpenZWave-d32394d4-01000000024c0000":"Closed"}`;
- `{"OpenZWave-d32394d4-01000000024c0000":{"Error":{"OperationNotSupported": ...}}`

Currently we rely on `Error` property of returned object, but that's dumb :) Theoretically valid value can be object with `Error` property. Something like this `{"OpenZWave-d32394d4-01000000024c0000":{ value: "Closed", error: {"OperationNotSupported": ...}}` would probably be less confusing.

components/taxonomy/src/parse.rs
```diff
@@ -168,7 +178,15 @@ impl ParseError {
 }
 
 #[derive(Debug)]
-pub struct JSONError(error::Error);
+pub struct JSONError(pub error::Error);
```
julienw, line 181: just wondering why you need the "pub" for the argument ?

components/taxonomy/src/parse.rs
```diff
@@ -168,7 +178,15 @@ impl ParseError {
...
+impl PartialEq for JSONError {
+    fn eq(&self, other: &JSONError) -> bool {
+        format!("{}", self.0) == format!("{}", other.0)
+    }
+}
+
+impl Eq for JSONError {}
 
 impl Clone for JSONError {
     fn clone(&self) -> JSONError {
```
julienw, line 192: such a useful implementation ;) (not for this patch, as this is not changed here, but I was quite surprised by this so wanted to mention it)

components/taxonomy/src/parse.rs
```diff
@@ -1,5 +1,8 @@
 //! Utilities for defining a JSON parser.
 
+#![allow(identity_op)] // Keep clippy happy with [De]serialize
```
julienw, line 3: do you know why clippy is unhappy without this ?

Yoric:
> 1) When incorrect rule is saved (e.g. without execute part), there is very vague error is returned {"thinkerbell-add-rule":{"Error":"Internal Error"}};

Almost sure that's not new. Looks like something that should be solved as part of https://github.com/fxbox/foxbox/issues/519.

> 2) It was there before, but since you're updating related pieces: it's a bit hard to distinguish successful response with value and failed one (especially for batch requests where response code should be 200 in any case):
> [...]
> Currently we rely on Error property of returned object, but that's dumb :) Theoretically valid value can be object with Error property. Something like this {"OpenZWave-d32394d4-01000000024c0000":{ value: "Closed", error: {"OperationNotSupported": ...}} would probably be less confusing.

Good idea, shouldn't be too hard. Can you file an issue?

components/taxonomy/src/values.rs
```diff
@@ -51,6 +55,189 @@ impl error::Error for TypeError {
...
+                Some(Self::parse(path, v, binary))
+            } else {
+                None
+            }
+        } else {
+            Some(Err(Error::ParseError(ParseError::type_error(field_name, &path, "object"))))
+        }
+    }
+}
+mopafy!(Data);
```
julienw, line 195: so that I understand properly, this is how we'll let adapters implement their own data types ?
Mopa is like an "open" enum ?

components/taxonomy/src/values.rs
```diff
@@ -51,6 +55,189 @@ impl error::Error for TypeError {
...
+    fn eq(&self, other: &Value) -> bool {
+        (self.content.eq)(&*self.content.data, &*other.content.data)
+    }
+}
+
+pub trait Data: Debug + Send + Sync + mopa::Any {
+    /// A human-readable description of the _type_ of the value.
+    ///
+    /// Used mainly in `TypeError` error messages.
+    fn description() -> String where Self: Sized;
```
julienw, line 142: wondering if this could be a `&'static str` like for Errors ?

components/taxonomy/src/parse.rs
Yoric, line 192: I needed `PartialEq` somewhere above, so I had to find a way to make it fit :)

components/taxonomy/src/parse.rs
Yoric, line 3: I haven't checked the generated code.

components/taxonomy/src/values.rs
julienw, line 142: actually I don't really care so just forget this comment :)

components/taxonomy/src/values.rs
Yoric, line 195: Basically, yes. Mopa is a more flexible + type-safe `Any` that can be used much more comfortably to implement an open enum.

components/taxonomy/src/values.rs
```diff
@@ -1556,19 +927,26 @@ impl TimeStamp {
...
 }
-impl Parser<TimeStamp> for TimeStamp {
+
+impl Data for TimeStamp {
     fn description() -> String {
-        "TimeStamp".to_owned()
+        "TimeStamp (RFC 3339)".to_owned()
     }
-    fn parse(path: Path, source: &JSON) -> Result<Self, ParseError> {
+    fn parse(path: Path, source: &JSON, _binary: &BinarySource) -> Result<Self, Error> {
```
julienw, line 935: wondering why _binary is here at all, I don't see where it's used ? but maybe I missed something.

components/taxonomy/tests/test_manager.rs
```diff
@@ -26,12 +26,12 @@ fn target_map<K, T>(mut source: Vec<(Vec<K>, T)>) -> TargetMap<K, T> where K: Cl
...
+    fn as_value(&self) -> Value;
 }
 
-impl Transform<Value> for (Payload, Arc<Format>) {
-    fn transform(&self) -> Value {
+impl AsValue for (Payload, Arc<Format>) {
+    fn as_value(&self) -> Value {
         self.0.to_value(&self.1).unwrap()
     }
 }
```
julienw, line 37: wondering if you could not simply use the rustier `impl From<(Payload, Arc<Format>)> for Value` ?

julienw, line 37: although it's not totally the same as this would take the ownership of self. OK forget this comment please :)

julienw, line 37: or... `impl From<&(Payload, Arc<Format>)> for Value` ? maybe it's not worth it :)

julienw:
r=me, I left some nits and questions but not really important.

We noticed that a DoorLock now uses Open/Closed instead of Locked/Unlocked, and I found that's because `DOOR_IS_LOCKED` uses Open/Closed. Is it expected ?

azasypkin:
> Good idea, shouldn't be too hard. Can you file an issue?

Done https://github.com/fxbox/foxbox/issues/522!

nojunpark:
r+ for me. glad to see the syntax got simpler.  It would be awesome if the wiki (https://wiki.mozilla.org/Connected_Devices/Projects/Project_Link/Taxonomy) is updated as well :)

Yoric:
> We noticed that a DoorLock now uses Open/Closed instead of Locked/Unlocked, and I found that's because DOOR_IS_LOCKED uses Open/Closed. Is it expected ?

I'll fix this.

components/taxonomy/src/values.rs
Yoric, line 935: Placeholder for a future patch :)

components/taxonomy/src/values.rs
Yoric, line 142: It makes things a bit more complicated for `Range<T>`, unfortunately.

components/taxonomy/src/parse.rs
Yoric, line 181: Er... I don't remember, but I remember that I do :)

fabricedesre:
r=me for tts

PR #523, opened 2016-06-09
Add a feature to turn off zwave support at build time (Closes #521)

### LINKED ISSUE #521: Add a feature to turn on/off zwave support at build time.
For platforms that don't have any zwave support, we should not even bring in any of this code.

@julienw could you take a look?

julienw:


julienw:
r? @fabrice
Is this what you had in mind ?

julienw:
@JohanLorenzo If you can also have a look at the test script change ?

tools/travis-linux.sh
```diff
@@ -5,11 +5,21 @@ set -ev
...
+}
 
-build() {
+build_default() {
+    echo '> Building foxbox with default features.'
     cargo build
 }
 
+build() {
+  build_without_features
```
JohanLorenzo: Nit: The other internal functions are prefixed by `_`. We could either rename `build_*` or change the implicit rule. I don't have a strong opinion on this.

JohanLorenzo:
LGTM for the travis script

tools/travis-linux.sh
julienw: fixed :)

src/adapters/mod.rs
```diff
@@ -65,11 +79,8 @@ impl<T: Controller> AdapterManager<T> {
         ip_camera::IPCameraAdapter::init(manager, self.controller.clone()).unwrap();
         let scripts_path = &self.controller.get_profile().path_for("thinkerbell_scripts.sqlite");
         ThinkerbellAdapter::init(manager, scripts_path).unwrap(); // FIXME: no unwrap!
-        let profile_openzwave = &self.controller.get_profile().path_for("openzwave");
-
-        let openzwave_devices = self.controller.clone().get_config().get("openzwave", "devices");
-        OpenzwaveAdapter::init(manager, profile_openzwave, openzwave_devices).unwrap();
```
fabricedesre, line 71: you could do:
#[cfg(feature = "zwave")]
{
  let profile_openzwave = ...
}

I'm always unhappy with the empty methods when the feature is disabled.

julienw, line 71: yeah I tried it first and I got an error, I'd need to enable an experimental rust feature. I thought it was better to do it the non-experimental way. What do you think ?

fabricedesre, line 71: We already use other experimental features, so I don't mind adding one. But no big deal if you prefer to keep it the way it is in this PR.

julienw, line 71: I think it makes sense to have a separate method because we have 3 logical lines in the "non-empty" case :)

julienw:
so @fabricedesre is it r+ ?

fabricedesre:
yep, r=me

PR #526, opened 2016-06-10
User activation logic and UI. r=fabrice

ferjm:


ferjm:
r? @fabricedesre

We need to merge https://github.com/fxbox/users/pull/87/files and change the dependency url before merging this one.

fabricedesre:
lgtm. I wonder if we could give users some hint that they need to be on the same network when accessing the validation link.

ferjm:
Absolutely. We need some UX input about what to send in the invitation email. We can send a different email body depending on the tunnel status. I'll file an issue for that. Thanks!

PR #529, opened 2016-06-16
Thinkerbell Adapter now expects rules as an object instead of a serialized string.

Yoric:
This should make things a bit more readable and a bit more robust.

In particular, this should fix 1) of https://github.com/fxbox/foxbox/pull/520#issuecomment-224869077 .

### DIFF
--- a/components/taxonomy/src/io.rs
+++ b/components/taxonomy/src/io.rs
@@ -17,9 +17,11 @@ pub struct BinarySource;
 /// Placeholder.
 pub struct BinaryTarget;
 
-/// Placeholder.
+
 #[derive(Clone, Debug, PartialEq, Eq, Serialize)]
-pub struct SerializeError;
+pub enum SerializeError {
+    JSON(String)
+}
 impl fmt::Display for SerializeError {
     fn fmt(&self, formatter: &mut fmt::Formatter) -> Result<(), fmt::Error> {
         (self as &fmt::Debug).fmt(formatter)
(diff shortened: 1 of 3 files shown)


Yoric:
Also, f? @azasypkin for the API changes. See changes in the tests for details.

mcav:
lgtm

azasypkin:
f+, looks good to me, just one question: is it expected that source getter still returns string instead of plain object? 

`https://local.609aebe9f46a699f80d6e0284c3334ddf879e125.box.knilxof.org:3000/api/v1/channels/get' -X PUT -H ..... --data-binary '[{"id":"thinkerbell/Everyday in the morning, camera takes a picture./source"}]'
`

Response:
`{"thinkerbell/Everyday in the morning, camera takes a picture./source":"{\"name\":\"Everyday in the morning, camera takes a picture.\",\"rules\":[{\"conditions\":[{\"feature\":\"clock/time-of-day-seconds\",\"source\":[{\"id\":\"getter:timeofday.clock@link.mozilla.org\"}],\"when\":{\"Geq\":{\"Duration\":28800}}}],\"execute\":[{\"destination\":[{\"id\":\"channel:snapshot.18db9b78-f188-11e5-9ce9-3ca9f47e346c@link.mozilla.org\"}],\"feature\":\"camera/store-snapshot\",\"value\":null}]}]}"}
`

Yoric:
Good catch. I'll make both match.

Yoric:
Ok, fixed.

@mcav, if you have a chance to take a look at the revised version, that would be great.

mcav:
Okay, still looks good to me

PR #567, opened 2016-12-08
Update README

samgiles:
@fabricedesre r?  

Not sure if the package names are correct for the Linux distros for `libev`.


samgiles:
Not sure why this is failing - this was just a README change.. :|

fabricedesre (approved):
The build failure on arm is from a previous PR, though I can't repro locally...

